PR #1, opened 2017-12-01
Prompt fallback if connection to Sentry fails

HazAT:


HazAT:
😬 did already @mitsuhiko 
Added it too late to `.gitignore` when it was already there, now it's gone.

PR #5, opened 2018-03-19
Fix three bugs with electron symbol uploader

ibash:
1. If on a beta version of electron the electronVersion would be
   incorrect since it would remove the word "beta".
2. Symbols were now downloaded / uploaded correctly since `symbols: true` wasn't set.
3. Symbols are breakpad, not dsym


jan-auer:
Thanks 🥇

PR #9, opened 2018-04-03
ref(electron): Update Electron integration code

jan-auer:
Updated integration code example due to https://github.com/getsentry/sentry-electron/pull/53:

- [x] Update code example with new syntax
- [x] Use public DSN instead of private one
- [x] Do not require `electron-download`
- [x] Update link to docs (moved to `/clients/electron`)

screen shot 2018-04-09 at 09 17 22

### DIFF
--- a/lib/Steps/Integrations/Electron.ts
+++ b/lib/Steps/Integrations/Electron.ts
@@ -4,16 +4,36 @@ import * as _ from 'lodash';
 import * as path from 'path';
 import { Args } from '../../Constants';
 import { exists } from '../../Helper/File';
-import { debug, green, l, nl, red } from '../../Helper/Logging';
+import { debug, dim, green, l, nl, red } from '../../Helper/Logging';
 import { SentryCli } from '../../Helper/SentryCli';
 import { BaseIntegration } from './BaseIntegration';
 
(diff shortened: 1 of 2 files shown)


jan-auer:
**HOLD**, this needs more fixes...

lib/Steps/Integrations/Electron.ts
```diff
@@ -4,17 +4,37 @@ import * as _ from 'lodash';
...
-// tslint:disable-next-line:radix
 const MIN_ELECTRON_VERSION = parseInt(
   MIN_ELECTRON_VERSION_STRING.replace(/\D+/g, ''),
+  10,
 );
+
+const CODE_EXAMPLE = `const Sentry = require('@sentry/electron');
+
+Sentry.init({
+  dsn: '___DSN___',
```
kamilogorek: Should we want use `___PUBLIC_DSN___` here as well?

kamilogorek:
Just this one comment. Other than this, its ready to go 👍

lib/Steps/Integrations/Electron.ts
jan-auer: It's not replaced by the same mechanism, and TIL `___DSN___` will be changed to emit the public DSN soon, so I'm gonna leave it for brevity.

PR #12, opened 2018-04-24
SentryCli is not defined

### LINKED ISSUE #11: SentryCli is not defined
When executing `node sentry-symbols.js` i receive the following error:
```
const sentryCli = new SentryCli("./sentry.properties")
                  ^

ReferenceError: SentryCli is not defined
```

Which makes sense as const are block-scoped you should wrapp all the code in to the try/catch

https://github.com/getsentry/sentry-wizard/blob/18bfdf69eff67d7f9a53bbe2adbb9f4a9e45519d/scripts/Electron/symbols.js#L3-L6

oliverwebr:
Fixes #11

### DIFF
--- a/scripts/Electron/symbols.js
+++ b/scripts/Electron/symbols.js
@@ -1,96 +1,101 @@
 #!/usr/bin/env node
 
+let SentryCli
+let download
+
 try {
-  const SentryCli = require('@sentry/cli');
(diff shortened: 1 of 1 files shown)


oliverwebr:
Sure thing 👍

oliverwebr:
Nice, keep up the good work. 👍

Issue #39, opened 2019-10-10
Allow skipping XCode build phases via envvar


brownoxford:
It would be great if it were possible to skip the xcode build phases for stripping unused architectures and uploading dsym data via an environment variable. This would be useful during tasks like snapshot generation where multiple builds are executed and these phases are not necessarily useful.

brownoxford:
@HazAT is this something that you would consider merging? I'm happy to modify the PR if needed

PR #40, opened 2019-10-11
feat: allow short-circuiting build phases with environment variables

brownoxford:
Skip 'Sentry strip unused archs from Framework' with SENTRY_SKIP_FRAMEWORK_PATCH
Skip 'Upload Debug Symbols to Sentry' with SENTRY_SKIP_DSYM_UPLOAD

Refs #39

### DIFF
--- a/lib/Steps/Integrations/Cordova.ts
+++ b/lib/Steps/Integrations/Cordova.ts
@@ -187,6 +187,11 @@ export class Cordova extends BaseIntegration {
       {
         shellPath: '/bin/sh',
         shellScript:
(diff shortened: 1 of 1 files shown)


HazAT:
Thx!

galfert:
Could you please tell me where I have to set these environment variables to skip the steps?

brownoxford:
> Could you please tell me where I have to set these environment variables to skip the steps?

I use this functionality to avoid these default Sentry actions when I am generating screenshots of my mobile app automatically, so I set the environment variables in my Fastlane configuration. The specific place you would set them depends on what you are doing, but in general you will want them to be set in the environment of whatever program is causing the actions to happen (where you don't actually want them happening).

If you have more information about your environment and what it is that you are doing, I may be able to give you more details.

PR #43, opened 2019-10-15
Move snyk to dev dependencies

sandfox:
As per https://github.com/getsentry/sentry-react-native/pull/621 snyk has added itself to the main/production deps which means end users have to install it and all of it's deps.

This change should stop library consumers having to install snyk as transitive
dependency.


HazAT:
Thx!

PR #66, opened 2020-05-15
fix: Add react-native sourcemap output path

HazAT:
This adds `EXTRA_PACKAGER_ARGS` to the iOS build script to build the source map in derived data location to make sure it's not bundled with the app.

### DIFF
--- a/lib/Steps/Integrations/ReactNative.ts
+++ b/lib/Steps/Integrations/ReactNative.ts
@@ -235,6 +235,7 @@ export class ReactNative extends MobileProject {
       let code = JSON.parse(script.shellScript);
       code =
         'export SENTRY_PROPERTIES=sentry.properties\n' +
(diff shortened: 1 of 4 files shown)


lib/Steps/Integrations/ReactNative.ts
```diff
@@ -265,14 +266,6 @@ export class ReactNative extends MobileProject {
...
   }
 
-  private addZLibToXcode(proj: any): void {
-    proj.addPbxGroup([], 'Frameworks', 'Application');
-    proj.addFramework('libz.tbd', {
-      link: true,
-      target: proj.getFirstTarget().uuid,
-    });
-  }
-
```
HazAT, line 275: React Native is based on cocoapods since a long time and we get the dep for zlib there when installing the sdk

HazAT:
Tests are failing because of too old node version

PR #68, opened 2020-07-01
Only upload sentry symbols for prod builds

steelbrain:
Current code makes it so Sentry symbols are uploaded on each build, even debug ones.
Sentry symbols uploading only makes sense in prod build so in addition to fixing this in my app, here's a PR so other people can enjoy faster debug builds

### DIFF
--- a/lib/Steps/Integrations/ReactNative.ts
+++ b/lib/Steps/Integrations/ReactNative.ts
@@ -260,8 +260,10 @@ export class ReactNative extends MobileProject {
       {
         shellPath: '/bin/sh',
         shellScript:
-          'export SENTRY_PROPERTIES=sentry.properties\\n' +
-          '../node_modules/@sentry/cli/bin/sentry-cli upload-dsym',
+          'if [ "$CONFIGURATION" != "Debug" ]; then\\n' +
+          '  export SENTRY_PROPERTIES=sentry.properties\\n' +
+          '  ../node_modules/@sentry/cli/bin/sentry-cli upload-dsym\\n' +
+          'fi\\n',
       },
     );
   }


lib/Steps/Integrations/ReactNative.ts
```diff
@@ -260,8 +260,10 @@ export class ReactNative extends MobileProject {
       {
         shellPath: '/bin/sh',
         shellScript:
-          'export SENTRY_PROPERTIES=sentry.properties\\n' +
-          '../node_modules/@sentry/cli/bin/sentry-cli upload-dsym',
+          'if [ "$CONFIGURATION" != "Debug" ]; then\\n' +
```
bruno-garcia, line 263: This is a good idea but I believe opting out completely isn't the way to go.
There are legitimate use cases to have your debug symbols uploaded in Debug builds.

Can we guard this check with a configuration property instead? 
Default value should keep the current behavior

steelbrain: Hi @bruno-garcia 

Thanks for the review! When you say configuration property check, do you mean have settings in `sentry.properties` that control whether or not the symbols are uploaded?

If we do so, we'll be continuing the inconsistency that exists today, currently Android does not upload symbols in dev, and iOS does. This PR aimed to make them both consistent, if we make the new change of API config, it'll break existing Android behavior

bruno-garcia: I'd like to keep the current behavior (unless we're bumping a major here).

A configuration option would allow your and anyone else to opt-out of uploading debug builds but would not break current setups. I was leaning towards continuing with that behavior.

Worse case it should be possible to opt-in to the old behavior (uploading debug) for those who we break once we ship this new (major then) release.

That said, I still prefer keeping things as they are and allowing the opt-out to avoid friction and the major bump but maybe @HazAT can chime in here.

steelbrain:
Closing for disagreement in the direction this is taking

PR #94, opened 2021-04-13
feat(nextjs): Add Next.js config files

iker-barriocanal:
The wizard was already adding the `next.config.js` to a Next.js project in the configuration. This PR adds the feature to the wizard to add `sentry.client.config.js` and `sentry.server.config.js` config files, where the Next.js SDK initialization will take place. This is a different approach to the previous one, which was based on setting the initialization parameters on `next.config.js`, but it was changed in https://github.com/getsentry/sentry-javascript/pull/3301/commits/15e4baa8f1337ec85b6fd8c3bb481007eb380937.


rhcarvalho (changes requested):
Right now `sentry.client.config.js` and `sentry.server.config.js` are identical.
Could this be something that confuses users?

I suggest adding a comment to each file with a brief explanation of when you would use each file, and a link to the relevant documentation.

client => configuration that affects code that runs in the browser whenever a page is visited

server => configuration that affects the node backend serving dynamic content

rhcarvalho:
Other than the above, everything looks reasonable.

(For the record, we also talked on a call about how we're ignoring version compatibility while we're on beta, in particular that the new config files are only useful in a currently unreleased version of the Next.js SDK/plugin)

rhcarvalho (approved):

PR #95, opened 2021-04-14
fix(nextjs): Clean and add docs to `next.config.js`

iker-barriocanal:
Unnecessary config elements from the `next.config.js` have been removed, and some docs have been added to explain what the file is about.


rhcarvalho (approved):
As the automation suggested, we can also get rid of `printExample`

rhcarvalho:
> As the automation suggested, we can also get rid of `printExample`

Never mind, I see it was done in #96

PR #98, opened 2021-04-17
chore(nextjs): Clarify messages, ignore `coverage` when linting

lobsterkatie:
A handful of small changes:

- Clarify that it's the user's `package.json` version of `next` which can't be set to `latest`.
- Clarify why we have to change the version number on our plugin
- Add a link to other sentry-webpack-plugin options
- Ignore `coverage` when linting

### DIFF
--- a/scripts/NextJs/configs/next.config.js
+++ b/scripts/NextJs/configs/next.config.js
@@ -21,7 +21,8 @@ const COMMIT_SHA =
 const SentryWebpackPlugin = require('@sentry/webpack-plugin');
 const fs = require('fs');
 
-// We require this to fake that our plugin matches the next version
+// Next.js requires a plugin's version to match the next.js version, so we fake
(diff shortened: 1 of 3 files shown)


HazAT (approved):

scripts/NextJs/configs/next.config.js
```diff
@@ -21,7 +21,8 @@ const COMMIT_SHA =
 const SentryWebpackPlugin = require('@sentry/webpack-plugin');
 const fs = require('fs');
 
-// We require this to fake that our plugin matches the next version
+// Next.js requires a plugin's version to match the next.js version, so we fake
+// it here by rewriting our plugin's package.json
```
rhcarvalho, line 25: This will be copied and live in the root of a users' nextjs project. The meaning of "ours" is not clear there, we better be more explicit.

PR #108, opened 2021-04-30
ref(nextjs): Don't hardcode package name

iker-barriocanal:
The package name was hardcoded, instead of using the local variable with the appropriate package name. This PR fixes this, using that variable instead.

### DIFF
--- a/lib/Steps/Integrations/NextJs.ts
+++ b/lib/Steps/Integrations/NextJs.ts
@@ -167,7 +167,7 @@ export class NextJs extends BaseIntegration {
       isNaN(devDepVersion)
     ) {
       red(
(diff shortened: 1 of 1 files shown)


lobsterkatie:
Given that this function is only used locally, in the `next` integration file, there's no reason for it to be a variable at all. So my vote would be to go the other way, and hardcode the spots where we already use a variable, because it's simpler and easier to read...

... except that there's a reason those other spots have a variable, and it's because this `_checkDep` is more or less identical to [this `_checkDep`](https://github.com/getsentry/sentry-wizard/blob/3c1fcc065f0e74c253f73df659f193d11f887f82/lib/Steps/Integrations/Electron.ts#L131). No requirement, but you might consider pulling this out into a helper function which could be used by both.

PR #113, opened 2021-05-21
ref(nextjs): Make webpack silent by default

iker-barriocanal:
The [Sentry Webpack plugin](https://github.com/getsentry/sentry-webpack-plugin) outputs logs that may be annoying for some users, especially when running in `dryRun` mode. Adding `silent` suppresses all logs, decreasing noise to users. Docs have already been updated, in https://github.com/getsentry/sentry-docs/pull/3612.

### DIFF
--- a/scripts/NextJs/configs/next.config.js
+++ b/scripts/NextJs/configs/next.config.js
@@ -15,6 +15,8 @@ const SentryWebpackPluginOptions = {
   // recommended:
   //   release, url, org, project, authToken, configFile, stripPrefix,
   //   urlPrefix, include, ignore
+
+  silent: true, // Suppresses all logs
   // For all available options, see:
   // https://github.com/getsentry/sentry-webpack-plugin#options.
 };


scripts/NextJs/configs/next.config.js
```diff
@@ -15,6 +15,8 @@ const SentryWebpackPluginOptions = {
   // recommended:
   //   release, url, org, project, authToken, configFile, stripPrefix,
   //   urlPrefix, include, ignore
+
+  silent: true, // Suppresses all logs
```
lobsterkatie, line 19: This is a matter of opinion, but I think defaulting to giving people information (with an easy opt-out) is preferable to hiding that information by default.

```suggestion
  // Uncomment this to suppress all logs
  // silent: true, 
  
```

kamilogorek: Our webpack plugin debug output is very noisy and it's not giving you much info tbh (as the output is way too detailed).

AbhiPrasad (approved):

PR #122, opened 2021-07-13
fix(nextjs): Parse version as a string, instead of int

iker-barriocanal:
Currently, the minimum version check only converts the version to an integer, so version such as `10.3` are considered lower than `10.3.0`. This PR fixes that, by parsing the version as strings; and only then, comparing the major, minor and patch numbers. `latest` is also accepted as a matching version.


lobsterkatie (changes requested):
I think we should use a package for this instead of rolling our own checker - something like https://github.com/npm/node-semver - as it can handle more edge cases (`10.x`, for example, all the things with `~` and `^`, anything specified with a compartor like  `<` or `>=`, etc, etc).

lib/Steps/Integrations/NextJs.ts
```diff
@@ -144,20 +145,17 @@ export class NextJs extends BaseIntegration {
...
     );
-    const devDepVersion = parseInt(
-      _.get(appPackage, ['devDependencies', packageName], '0').replace(
-        /\D+/g,
-        '',
-      ),
-      10,
+    const devDepVersion = _.get(
+      appPackage,
+      ['devDependencies', packageName],
```
AbhiPrasad, line 155: Why do we check `devDeps`? Next should always be a dependency, if it's not something is wrong.

AbhiPrasad (approved):

lib/Steps/Integrations/NextJs.ts
```diff
@@ -176,25 +181,26 @@ export class NextJs extends BaseIntegration {
...
-    ] = proposed.split('.');
-    const [minMajor, minMinor, minPatch] = MIN_NEXTJS_VERSION.split('.');
-    if (
-      parseInt(minMajor) > parseInt(proposedMajor) ||
-      parseInt(minMinor) > parseInt(proposedMinor) ||
-      parseInt(minPatch) > parseInt(proposedPatch)
-    ) {
+
+    const cleanedVersion = clean(version);
+    if (cleanedVersion) {
```
lobsterkatie, line 192: When would `cleanedVersion` not be defined?

lib/Steps/Integrations/NextJs.ts
```diff
@@ -176,25 +181,26 @@ export class NextJs extends BaseIntegration {
...
+      // gte(x, y) : true if x >= y
+      return gte(cleanedVersion, MIN_NEXTJS_VERSION);
+    }
+
+    const minVersionRange = `>=${MIN_NEXTJS_VERSION}`;
+    const userVersionRange = validRange(version);
```
lobsterkatie, line 198: Does `validRange` return a boolean about whether or not the range is valid, or is it a normalization function (or both)? It's not clear to me whether or not we know by this point that we're dealing with a range.

lobsterkatie (commented):
Does one of your code paths handle the `10.x` possibility?

lib/Steps/Integrations/NextJs.ts
iker-barriocanal: Yes, it should. But having it as a dev dependency runs the project successfully, and users might have it there. I rather check it, than not doing it and having the chance of getting issues because of it.

lib/Steps/Integrations/NextJs.ts
iker-barriocanal: Thought about adding these comments, but I think it's better to read the docs of the function/module. If the module gets updated, docs may get outdated and not represent what the code is doing. IMO, having comments stating the code to be doing a thing it actually isn't (even worse if there're examples) is worse than having to look at the documentation to know what it's doing. The comment I added for `gte` contradicts what I'm saying though. Wdyt?

lib/Steps/Integrations/NextJs.ts
iker-barriocanal: Same answer as before. If it was a boolean, the variable would probably be named `isValidRange` (note the `is` at the beginning). I can add the comments, but I think it has more cons than pros.

iker-barriocanal:
@lobsterkatie  [`validRange`](https://github.com/npm/node-semver#ranges-1) expects a range, and [X-Ranges](https://github.com/npm/node-semver#x-ranges-12x-1x-12-) are valid ranges. So yes, `10.x` is handled in that path.

lib/Steps/Integrations/NextJs.ts
lobsterkatie: I am 100% Team Comment Your Code. We pin the version of the dependency, so things shouldn't change out from under us without us knowing, and as a reader of the code I def do not want to have to go read the docs for some other library, just to understand what our code is doing.

I missed your replies and I see that you just merged it, so NBD, but in the future, I think it is definitely better to answer whatever questions you anticipate the reader might have than hope they'll go read docs elsewhere.

PR #128, opened 2021-09-03
fix(nextjs): Don't expose auth token in `sentry.properties`

iker-barriocanal:
The Wizard is currently adding every property required by Sentry CLI in `sentry.properties`, including the auth token (that shouldn't be exposed). This PR fixes this behavior by adding the auth token to `.sentryclirc` (the Sentry CLI config file), which is added to the `.gitignore`. `sentry.properties` can now be safely committed.

### DIFF
--- a/lib/Steps/Integrations/NextJs.ts
+++ b/lib/Steps/Integrations/NextJs.ts
@@ -6,11 +6,13 @@ import { clean, gte, minVersion, satisfies, validRange } from 'semver';
 
 import { Args } from '../../Constants';
 import { debug, green, l, nl, red } from '../../Helper/Logging';
-import { SentryCli } from '../../Helper/SentryCli';
+import { SentryCli, SentryCliProps } from '../../Helper/SentryCli';
 import { BaseIntegration } from './BaseIntegration';
 
 const MIN_NEXTJS_VERSION = '10.0.8'; // Must be a fixed version: `X.Y.Z`
 const PROPERTIES_FILENAME = 'sentry.properties';
+const SENTRYCLIRC_FILENAME = '.sentryclirc';
+const GITIGNORE_FILENAME = '.gitignore';
 const CONFIG_DIR = 'configs/';
 const MERGEABLE_CONFIG_PREFIX = '_';
 
@@ -35,12 +37,7 @@ export class NextJs extends BaseIntegration {
     nl();
 
     const sentryCliProps = this._sentryCli.convertAnswersToProperties(answers);
-    fs.writeFileSync(
-      `./${PROPERTIES_FILENAME}`,
-      this._sentryCli.dumpProperties(sentryCliProps),
-    );
-    green(`Successfully created sentry.properties`);
-    nl();
+    this._createSentryCliConfig(sentryCliProps);
 
     const configDirectory = path.join(
       __dirname,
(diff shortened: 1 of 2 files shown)


lib/Steps/Integrations/NextJs.ts
```diff
@@ -97,6 +94,59 @@ export class NextJs extends BaseIntegration {
...
+
+    fs.writeFileSync(
+      `./${PROPERTIES_FILENAME}`,
+      this._sentryCli.dumpProperties(cliProps),
+    );
+    green(`Successfully created sentry.properties`);
+    nl();
+  }
+
+  private _gitIgnoreFile(filepath: string, errorMsg: string): void {
```
AbhiPrasad, line 128:
```suggestion
  private _addToGitignoreFile(filepath: string, errorMsg: string): void {
```

lib/Steps/Integrations/NextJs.ts
```diff
@@ -97,6 +94,59 @@ export class NextJs extends BaseIntegration {
...
+    const authToken = cliProps['auth/token'];
+
+    /**
+     * To not commit the auth token to the VCS, instead of adding it to the
+     * properties file (like the rest of props), it's added to the Sentry CLI
+     * config, which is added to the gitignore. This way makes the properties
+     * file safe to commit without exposing any auth tokens.
+     */
+    if (authToken) {
+      const authTokenProp = { 'auth/token': authToken };
```
AbhiPrasad, line 107: nit: No need for a var, it’s only used once.

lib/Steps/Integrations/NextJs.ts
```diff
@@ -97,6 +94,59 @@ export class NextJs extends BaseIntegration {
...
+      fs.appendFileSync(
```
AbhiPrasad, line 108: Can we use async await and fs promises?

lib/Steps/Integrations/NextJs.ts
```diff
@@ -97,6 +94,59 @@ export class NextJs extends BaseIntegration {
...
+        SENTRYCLIRC_FILENAME,
+        this._sentryCli.dumpProperties(authTokenProp),
+      );
+      this._gitIgnoreFile(
+        SENTRYCLIRC_FILENAME,
+        `⚠ Could not add ${SENTRYCLIRC_FILENAME} to ${GITIGNORE_FILENAME}, ` +
+          'please add it to not commit your auth key.',
+      );
+      delete cliProps['auth/token'];
```
AbhiPrasad, line 117: Not a fan of mutating the function argument. Can we do this in an immutable way?

lib/Steps/Integrations/NextJs.ts
```diff
@@ -94,8 +94,11 @@ export class NextJs extends BaseIntegration {
     return this.shouldConfigure;
   }
 
-  private _createSentryCliConfig(cliProps: SentryCliProps): void {
-    const authToken = cliProps['auth/token'];
+  private async _createSentryCliConfig(
+    cliProps: SentryCliProps,
+  ): Promise<void> {
+    const cliPropsToWrite = { ...cliProps };
```
AbhiPrasad, line 100: Let’s instead do
```suggestion
    const { “auth/token”: authToken, ...cliPropsToWrite } = cliProps:
```

that way we don’t gotta use the delete statement.

lib/Steps/Integrations/NextJs.ts
```diff
@@ -97,6 +94,83 @@ export class NextJs extends BaseIntegration {
...
+    const authToken = cliPropsToWrite['auth/token'];
+
+    /**
+     * To not commit the auth token to the VCS, instead of adding it to the
+     * properties file (like the rest of props), it's added to the Sentry CLI
+     * config, which is added to the gitignore. This way makes the properties
+     * file safe to commit without exposing any auth tokens.
+     */
+    if (authToken) {
```
AbhiPrasad, line 109: Can we log an error if auth token is undefined (else condition on this if statement)?

AbhiPrasad (approved):

PR #129, opened 2021-09-10
fix(nextjs): Correctly set token in .sentryclirc

AbhiPrasad:
Should resolve https://github.com/getsentry/sentry-javascript/issues/3968

We need to fix docs as well: https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#use-configuration-files, can do that afterwards


AbhiPrasad:
Closing in favour of https://github.com/getsentry/sentry-wizard/pull/131

PR #135, opened 2021-09-23
feat: Jira GHA

HazAT:


vladanpaunovic (commented):
We can examine later how can we assign a `NEEDS TRIAGE` status

PR #146, opened 2021-12-14
fix(nextjs): Do not capture 404s as exceptions in `_error`

koenpunt:
By checking the status code of the response it's possible to filter out errors for not found pages. 

I've also updated it to pass the full page context to `getInitialProps`, because otherwise this example would fail to build when converted to typescript.

### DIFF
--- a/scripts/NextJs/configs/_error.js
+++ b/scripts/NextJs/configs/_error.js
@@ -23,7 +23,7 @@ MyError.getInitialProps = async (context) => {
   // getInitialProps has run
   errorInitialProps.hasGetInitialPropsRun = true;
 
-  // Returning early beacause we don't want to log 404 errors to Sentry.
+  // Returning early because we don't want to log 404 errors to Sentry.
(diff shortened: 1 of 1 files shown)


lobsterkatie:
Hey, @koenpunt, thanks for the contribution!

This looks good, but can you please move your 404 check to be right after line 24? I think bailing before we get to any kind of error handling code is cleaner.

```
  // we don't want to log these in Sentry, so at this point we're done
  if (res.statusCode === 404) {
    return errorInitialProps;
  }
  ```

koenpunt:
I think it's possible to cause a 404 and have an error, and in that case I would still expect that error to be logged.

I'll verify this later today.

koenpunt:
> I think it's possible to cause a 404 and have an error

Okay, never mind, this isn't possible. I've updated the code.

lobsterkatie (approved):
LGTM! Thanks, @koenpunt!

PR #161, opened 2022-04-09
Fix Electron version detection to use electron/package.json

andersk:
The top-level `package.json` might specify a caret range that doesn’t tell us exactly which version is installed. Look at Electron’s `package.json` instead.

### DIFF
--- a/scripts/Electron/symbols.js
+++ b/scripts/Electron/symbols.js
@@ -12,15 +12,13 @@ try {
   process.exit(1);
 }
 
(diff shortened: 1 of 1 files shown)


andersk:
Oh boy—another place I get to post meaningless periodic comments to prevent my contributions from being closed by a bot! And—three weeks—really? There’s been a single upstream commit during these last three weeks, updating a link in a comment. That’s not going to have invalidated any issues and PRs here. Please consider whether this stale bot configuration is really a good use of either your time or mine.

lobsterkatie (approved):
Sorry for the delay. This LGTM. Thanks for the contribution!

PR #173, opened 2022-07-07
Patch React Native xcode RN 0.69

marandaneto:
Fix https://github.com/getsentry/sentry-react-native/issues/2267
No need of doing https://docs.sentry.io/platforms/react-native/troubleshooting/#react-native-069-and-higher manually
We should bump a minor 1.3.0 since it's not compatible with RN < 0.69

Next: uninstall steps are broken already even for the current version of RN.
Wondering if we should get rid of the uninstall patches or not.
https://github.com/getsentry/sentry-wizard/pull/66/files Adds `EXTRA_PACKAGER_ARGS` but does not remove it during uninstall steps.
It does not unpatch the App.js that calls `Sentry.init`.
It does not delete the sentry.properties files, but just comment the fields such as `#defaults.org=`.

### DIFF
--- a/lib/Steps/Integrations/ReactNative.ts
+++ b/lib/Steps/Integrations/ReactNative.ts
@@ -231,9 +231,9 @@ export class ReactNative extends MobileProject {
     for (const script of buildScripts) {
       if (
         !script.shellScript.match(
-          /(packager|scripts)\/react-native-xcode\.sh\b/,
+          /\/scripts\/react-native-xcode\.sh/i,
         ) ||
-        script.shellScript.match(/sentry-cli\s+react-native[\s-]xcode/)
+        script.shellScript.match(/sentry-cli\s+react-native\s+xcode/i)
       ) {
         continue;
       }
@@ -243,9 +243,10 @@ export class ReactNative extends MobileProject {
         'export SENTRY_PROPERTIES=sentry.properties\n' +
         'export EXTRA_PACKAGER_ARGS="--sourcemap-output $DERIVED_FILE_DIR/main.jsbundle.map"\n' +
         code.replace(
-          /^.*?\/(packager|scripts)\/react-native-xcode\.sh\s*/m,
-          (match: any) =>
-            `../node_modules/@sentry/cli/bin/sentry-cli react-native xcode ${match}`,
+          '$REACT_NATIVE_XCODE',
+          () =>
+            // eslint-disable-next-line no-useless-escape
+            '\\\"../node_modules/@sentry/cli/bin/sentry-cli react-native xcode $REACT_NATIVE_XCODE\\\"',
         );
       script.shellScript = JSON.stringify(code);
     }
--- a/lib/Steps/Integrations/ReactNative.ts
+++ b/lib/Steps/Integrations/ReactNative.ts
@@ -340,38 +340,20 @@ export class ReactNative extends MobileProject {
       }
 
       // ignore scripts that do not invoke the react-native-xcode command.
-      if (!script.shellScript.match(/sentry-cli\s+react-native[\s-]xcode\b/)) {
+      if (!script.shellScript.match(/sentry-cli\s+react-native\s+xcode/i)) {
         continue;
       }
 
       script.shellScript = JSON.stringify(
         JSON.parse(script.shellScript)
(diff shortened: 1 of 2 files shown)


lib/Steps/Integrations/ReactNative.ts
```diff
@@ -231,9 +231,9 @@ export class ReactNative extends MobileProject {
     for (const script of buildScripts) {
       if (
         !script.shellScript.match(
-          /(packager|scripts)\/react-native-xcode\.sh\b/,
+          /\/scripts\/react-native-xcode\.sh/i,
```
kamilogorek, line 234: Why do we need to be case-insensitive in all checks?

kamilogorek: If not, we can just use `shellScript.includes('/scripts/react-native-xcode.sh')`

lib/Steps/Integrations/ReactNative.ts
```diff
@@ -231,9 +231,9 @@ export class ReactNative extends MobileProject {
...
         ) ||
-        script.shellScript.match(/sentry-cli\s+react-native[\s-]xcode/)
+        script.shellScript.match(/sentry-cli\s+react-native\s+xcode/i)
```
kamilogorek, line 236: Is there even a way to have more whitespace? If not, i'd also go with `includes` here.

lib/Steps/Integrations/ReactNative.ts
```diff
@@ -243,9 +243,10 @@ export class ReactNative extends MobileProject {
         'export SENTRY_PROPERTIES=sentry.properties\n' +
         'export EXTRA_PACKAGER_ARGS="--sourcemap-output $DERIVED_FILE_DIR/main.jsbundle.map"\n' +
         code.replace(
-          /^.*?\/(packager|scripts)\/react-native-xcode\.sh\s*/m,
-          (match: any) =>
-            `../node_modules/@sentry/cli/bin/sentry-cli react-native xcode ${match}`,
+          '$REACT_NATIVE_XCODE',
+          () =>
+            // eslint-disable-next-line no-useless-escape
+            '\\\"../node_modules/@sentry/cli/bin/sentry-cli react-native xcode $REACT_NATIVE_XCODE\\\"',
```
kamilogorek, line 249: You can try to use backticks instead, it should make this eslint rule redundant and reduce number of escapes needed

lib/Steps/Integrations/ReactNative.ts
```diff
@@ -339,38 +340,20 @@ export class ReactNative extends MobileProject {
       }
 
       // ignore scripts that do not invoke the react-native-xcode command.
-      if (!script.shellScript.match(/sentry-cli\s+react-native[\s-]xcode\b/)) {
+      if (!script.shellScript.match(/sentry-cli\s+react-native\s+xcode/i)) {
```
kamilogorek, line 343: Same as above

kamilogorek (changes requested):

lib/Steps/Integrations/ReactNative.ts
marandaneto: I've tried with backticks and it does not help, linter still complains, worth noting that I'm not scaping the String, I actually need 3 `\\\` -> https://docs.sentry.io/platforms/react-native/troubleshooting/#react-native-069-and-higher

lib/Steps/Integrations/ReactNative.ts
marandaneto: https://github.com/getsentry/sentry-wizard/pull/173#discussion_r916784526

kamilogorek (approved):
One nit, but feel free to merge after approval.

marandaneto:
@kamilogorek and @bruno-garcia If I got it right -> https://classic.yarnpkg.com/en/docs/dependency-versions#toc-caret-ranges
Releasing Sentry Wizard as `1.3.0` won't help React Native because older versions of the SDK use the `^1.2.17` syntax and `1.3.0` is a valid range within the `^1.2.17` definition.
We'd need to release Sentry Wizard as `2.0.0` so older versions of the React Native SDK would look up for `1.2.17`, and newer versions would pin to `2.0.0`, keeping retro compatibility, is that right?

PR #178, opened 2022-07-21
Revert "Revert "Patch React Native xcode RN 0.69""

marandaneto:
Reverts getsentry/sentry-wizard#177

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,14 @@
 # Changelog
 
-## 1.3.1
+## 2.0.0
+
+- fix(react-native) Xcode linking for RN 0.69 and higher (#173)
 
 ## 1.3.0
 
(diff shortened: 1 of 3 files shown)


kamilogorek (approved):
`# Unreleased` would also do, but we know its `2.0.0` so we can leave it as is.

Also is there nothing else that we could ship with `2.0`? It's a good timing to do so.

vladanpaunovic (approved):

marandaneto:
> `# Unreleased` would also do, but we know its `2.0.0` so we can leave it as is.
> 
> Also is there nothing else that we could ship with `2.0`? It's a good timing to do so.

https://github.com/getsentry/sentry-wizard/runs/7445675957?check_suite_focus=true failed because of `No changelog entry found for version "1.3.1"` so I thought that it was necessary for some reason.
The 2.0.0 is a P0 for RN, so we have to do it asap.

PR #179, opened 2022-07-25
Add page to send onboarding sample error for Next.js

Neo-Zhixing:
image
image

This PR added a new page that we insert to the user's project directory during installation. This page shows a button that, once clicked, triggers an error intentionally.

This allows us to remove instructions related to sending sample errors during onboarding.

### DIFF
--- a/lib/Steps/Integrations/NextJs.ts
+++ b/lib/Steps/Integrations/NextJs.ts
@@ -4,6 +4,7 @@ import { Answers, prompt } from 'inquirer';
 import * as _ from 'lodash';
 import * as path from 'path';
 import { satisfies, subset, valid, validRange } from 'semver';
+import Chalk from 'chalk';
 
 import { Args } from '../../Constants';
 import { debug, green, l, nl, red } from '../../Helper/Logging';
@@ -50,12 +51,15 @@ export class NextJs extends BaseIntegration {
     const sentryCliProps = this._sentryCli.convertAnswersToProperties(answers);
     await this._createSentryCliConfig(sentryCliProps);
 
-    const configDirectory = path.join(
+    const templateDirectory = path.join(
       __dirname,
       '..',
       '..',
       '..',
       'NextJs',
+    );
+    const configDirectory = path.join(
+      templateDirectory,
       CONFIG_DIR,
     );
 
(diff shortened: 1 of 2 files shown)


scefali (approved):
LGTM but other folks hsould review as well

lib/Steps/Integrations/NextJs.ts
```diff
@@ -68,6 +72,30 @@ export class NextJs extends BaseIntegration {
       nl();
     }
 
+    const selectedProjectSlug: string | null = answers.config?.project?.slug;
+    if (selectedProjectSlug) {
+      const hasFirstEvent = answers.wizard?.projects?.find?.((p: { slug: string; }) => p.slug === selectedProjectSlug)?.firstEvent;
+      if (!hasFirstEvent) {
```
iker-barriocanal, line 78: I don't remember what the answers look like, but the only reason you get the project slug is to check if there's already an event. Assuming the project is selected after fetching them from Sentry, can you get that `firstEvent` directly? Something like:
```js
const hasFirstEvent = answers.config?.project?.firstEvent;
```

lib/Steps/Integrations/NextJs.ts
```diff
@@ -68,6 +72,30 @@ export class NextJs extends BaseIntegration {
...
+          templateDirectory,
+          'sentry_sample_error.js',
+          ['pages', 'src/pages'],
+          dsn,
+        );
+        l(
+          Chalk.bgYellowBright(`
+|------------------------------------------------------------------------|
+|                          Installation Complete                         |
+| To verify your installation and finish onboarding, launch your Nest.js |
```
iker-barriocanal, line 89:
```suggestion
| To verify your installation and finish onboarding, launch your Next.js |
```

lib/Steps/Integrations/NextJs.ts
iker-barriocanal: Heads up if you also want to target big orgs: there's a limit of max 50 projects fetched per org in the API, [see](https://github.com/getsentry/sentry/blob/cdfe392ddbfa78c1560f51ce85f0d39c3b42ab00/src/sentry/web/frontend/setup_wizard.py#L47). Projects are sorted by date added desc, so it shouldn't be a problem though. Feel free to land this as-is, and open a follow-up PR if you want to add further changes.

iker-barriocanal (approved):

lib/Steps/Integrations/NextJs.ts
Neo-Zhixing: When I tested this, answers.config.project only has the slug property and nothing else. That's why I had to do this lookup.

PR #183, opened 2022-08-05
fix(electron): Remove Electron `symbols.js` script

timfish:
Uploading symbols is no longer required:
https://github.com/getsentry/sentry-electron/issues/524

This also bumps `MIN_ELECTRON_VERSION_STRING` to `2.0.0` since this has been the minimum version since v3 of the SDK.


AbhiPrasad (commented):
We'll need to update the docs first https://docs.sentry.io/platforms/javascript/guides/electron/#uploading-debug-information

Also reminder - this needs a changelog entry!

AbhiPrasad (approved):

Issue #184, opened 2022-08-11
Does not patch root file (App.js) in react-native project


esthor:
### Environment

React Native

How do you use Sentry?
Sentry SaaS

Which SDK and version?
4.2.3 (latest stable)
5.0.0-alpha.1 (latest alpha, for react-native 0.69+ support)

### Steps to Reproduce

1. Run `npx @sentry/wizard -i reactNative -p ios android`
2. You get a `✅ Patched App.js file.` message in the terminal
3. You're App.js isn't patched 😭 

### Expected Result
App.js to be patched.

### Actual Result

No App.js or js files were updated.

Looking at the source in `lib/Steps/Integrations/ReactNative.ts` I can see that this message will just be printed regardless of whether a successful patch occurred.

Now...here's two spots where it can get real juicy...

**Tree traversal**: It's common to have the `App.js` "root" component in the root directory of a project or in a `src` folder...it can really be anywhere.
1. one naive approach might be to traverse the whole projects file structure until you find an `App.js` and then assume that's the one to patch. You'd then need to consider priority and ignore `node_modules` of course, and it could get messy for other reasons..._[foreshadowing 👻 ]_
2. Another naive approach might be to look at the `index.js` file and see what component it passes to the `AppRegistry.registerComponent()` function call and then look where that is imported from...because that's going to be your "root" component that you want to patch. Problem here is folks might not have a `index.js` because they may be defining another entry file or heck, they might have it as a `index.jsx` or `index.ts` or `index.tsx` even if they keep named "index"...but that isn't even the spookiest... _[foreshadowing intensifies 👻 👻 ]_
3. Ok, well, actually, I've pretty much said the biggest issues with a naive approach of trying to be smart/automated about pattern matching. Even freshly init'd react native projects aren't going to follow a standard pattern due to the freedom the `react-native-template-*` feature of `react-native` cli gives template authors. You could have nested root components of any name (`main` and `root` are other common ones). So, what could you do? ...
4. So, you could make it an interactive CLI. Just ask me where my App.js or equivalent file is and what it's called. 🤷 
5. At least, don't tell me it patched a file that doesn't exist 🙏  especially after I've had too much coffee and spare time to write a novel of a GitHub issue.

krystofwoldrich:
Hi, thank you for the info,
I've just tried it with a clean project and the `App.js` was updated correctly.

Steps I've done...
```bash
npx react-native init MyTestProjectName

yarn add @sentry/react-native   
[...]

PR #185, opened 2022-08-18
feat(rn): Add support for RN TypeScript and other templates

krystofwoldrich:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [X] Refactoring

## :scroll: Description
- Typescript and other JS file formats support
- Searches for file in the root of the project and `src` dir.
Those seems to be the two most obvious places thus supporting novice devs. If the project has different structure we can assume that the dev will be able to complete the init manually.

### The new look
Screenshot 2022-08-18 at 16 36 02
Screenshot 2022-08-18 at 16 38 08

## :bulb: Motivation and Context
Solves https://github.com/getsentry/sentry-wizard/issues/184

## :green_heart: How did you test it?
So far only tested manually with `react-native init`

## :pencil: Checklist

- [X] I reviewed submitted code
- [ ] I added tests to verify changes
- [X] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps
- [ ] Unit or integration tests for the glob pattern


lib/Helper/File.ts
```diff
@@ -54,7 +60,7 @@ export function matchesContent(
     return false;
   }
   return matches.reduce((prev: boolean, match: string) => {
-    return (
+    return !!(
```
marandaneto, line 63: Should we rename the function if now we are using the double not?

package.json
```diff
@@ -37,6 +37,7 @@
   },
   "devDependencies": {
     "@sentry-internal/eslint-config-sdk": "^6.2.1",
+    "@types/glob": "^7.2.0",
```
marandaneto, line 40: Is this a `devDependencies`? because we are using it directly in the code.

marandaneto:
Adding the Web team for review since it touches common files and not RN only.

@krystofwoldrich would be possible to add a test for the RN integration that covers multiple files? I guess there's none right now but good timing to start doing it.

lib/Helper/File.ts
krystofwoldrich: The function was and is still returning boolean. The reason for the `!!` is ts typings, without the `ts` the boolean return value is not recognize. I would not rename is since it's still perform exactly the same function.

package.json
krystofwoldrich: Yes, it is a devDep, it's only typing for the Glob dep. And typings of our deps are used only for development. The actual glob lib is regular dep and install when wizard is used.

krystofwoldrich:
> @krystofwoldrich would be possible to add a test for the RN integration that covers multiple files? I guess there's none right now but good timing to start doing it.

Yes, for sure, I was definitely thinking at least about test for the Glob pattern (mentioned in the desc), but I was also surprised there are very little tests.

marandaneto:
> > @krystofwoldrich would be possible to add a test for the RN integration that covers multiple files? I guess there's none right now but good timing to start doing it.
> 
> Yes, for sure, I was definitely thinking at least about test for the Glob pattern (mentioned in the desc), but I was also surprised there are very little tests.

Yeah time to change that, for the File class, See https://github.com/getsentry/sentry-wizard/blob/master/lib/Helper/__tests__/File.ts
For the Logging class, I'd say to not bother about it.
And the RN integration class with the changes, we need a new class that tests the basic stuff, such as your own change, I'd say let's focus on this now and later we expand the other tests.

krystofwoldrich:
I tried to implement this with minimum impact on the current behaviour, but it comes with some compromises. I was wandering what do you think @marandaneto 

1. Do we want to support the index.{platform} files? We don't support version of RN where that was common. 
2. When the js file to patch is not found the warning shows twice for both platforms. I was thinking about moving it outside of the platforms patches, so it happens only once. (That only makes sense if 1.)
3. Is it okay to still say successfully patched? I was think about changing it to finished with warning or something like that, but it would mean unnecessary large change for probably not that impactful change.

marandaneto:
> 1. Do we want to support the index.{platform} files? We don't support version of RN where that is common.

1. not needed, they can still follow the docs for such cases https://docs.sentry.io/platforms/react-native/#initialize
2. it's fine, maybe add in the logger which file was not found, so each logging has the file name eg index.js, index.ts, etc...
3. it depends, the wizard patch multiple files, if one of them got patched, it still makes sense, so the user knows that the index was not found (because of logging), but the other files were patched.

krystofwoldrich:
> > 1. Do we want to support the index.{platform} files? We don't support version of RN where that is common.
> 
> 1. not needed, they can still follow the docs for such cases https://docs.sentry.io/platforms/react-native/#initialize
> 2. it's fine, maybe add in the logger which file was not found, so each logging has the file name eg index.js, index.ts, etc...
> 3. it depends, the wizard patch multiple files, if one of them got patched, it still makes sense, so the user knows that the index was not found (because of logging), but the other files were patched.

1. Okay, that would be nice, but is it worth a breaking change, or is that not a concern here? 
2. I was thinking about that too, because the Glob pattern is a bit complicated. But I don't think there is a way to expand it to an array of paths it matches because that can be possible unlimited. Our glob has final number of paths is matches but it would still require a writing parser for it. (Or if such a parser exist then it's easy but I don't know about it.) Or stop using glob and use just array or paths, which is inferior technical solution, but might be better UX.

lib/Steps/Integrations/ReactNative.ts
```diff
@@ -132,6 +119,32 @@ export class ReactNative extends MobileProject {
...
+      await patchMatchingFile(
+        jsFileGlob,
+        this._patchJs.bind(this),
+        answers,
+        platform,
+      );
+      dim(`✅ Patched ${jsFileToPatch.join(', ')} file(s).`);
+    } else {
+      dim(`🚨 Could not find ${platformGlob} nor ${universalGlob} files.`);
+      yellow('❓ Please, visit https://docs.sentry.io/platforms/react-native/#initialize');
```
marandaneto, line 144:
```suggestion
      dim(`🚨 Could not find ${platformGlob} nor ${universalGlob} files.`);
      yellow('❓ Please, visit https://docs.sentry.io/platforms/react-native');
```

Just in case we change the headline, the docs will be broken.

marandaneto:
@krystofwoldrich 

> Okay, that would be nice, but is it worth a breaking change, or is that not a concern here?

Since we already support it, I'd rather keep it, I thought that we didn't support it at all.

marandaneto:
> I was thinking about that too, because the Glob pattern is a bit complicated. But I don't think there is a way to expand it to an array of paths it matches because that can be possible unlimited. Our glob has final number of paths is matches but it would still require a writing parser for it. (Or if such a parser exist then it's easy but I don't know about it.) Or stop using glob and use just array or paths, which is inferior technical solution, but might be better UX.

I'd rather keep it as it is then, the goal of this PR is to support the TS templates, so let's not deviate from it too much.

krystofwoldrich:
@getsentry/team-web-sdk-frontend Is it okay to bump up the node typings, so I can use `rmSync` in the test?
Also I fixed some tiny bugs that showed when writing the test.

lforst:
> @getsentry/team-web-sdk-frontend Is it okay to bump up the node typings and node version for test, so I can use `rmSync` in the test?

I'm not too sure about bumping the typings. I lack knowledge of the history of the wizard but I would guess the types are kept at Node 10 so we "enforce" support for older Node versions. However, In the JS SDK, the oldest version is Node 8 so I'm not too sure about all that. Maybe someone with a little bit more context can weigh in 😬 (I think the right people are already tagged in this PR)

krystofwoldrich:
> 

I've removed the node version change. So now it should be ready to merge. @lforst @marandaneto

lforst (approved):

PR #188, opened 2022-08-30
ref(nextjs): Default to hiding source maps in nextjs config

lobsterkatie:
This adds a default value for `hideSourceMaps` to the `next.config.js` the wizard provides, so that new users will use `hidden-source-map` as their Webpack `devtool` by default in client-side builds. (This prevents browser devtools from seeing sourcemaps and therefore prevents original source code from showing up automatically in the `Sources` tab. (More information on `devtool` values can be found [here](https://webpack.js.org/configuration/devtool/).)

In conjunction with https://github.com/getsentry/sentry-javascript/pull/5649, https://github.com/getsentry/sentry-docs/pull/5464, and https://github.com/vercel/next.js/pull/40079, this is the first step in addressing the concerns raised in https://github.com/getsentry/sentry-javascript/issues/4489. See [here](https://github.com/getsentry/sentry-javascript/issues/4489#issuecomment-1231137629) for more details.

### DIFF
--- a/scripts/NextJs/configs/next.config.js
+++ b/scripts/NextJs/configs/next.config.js
@@ -7,6 +7,16 @@ const { withSentryConfig } = require('@sentry/nextjs');
 
 const moduleExports = {
   // Your existing module.exports
+
+  sentry: {
+    // Use `hidden-source-map` rather than `source-map` as the Webpack `devtool`
+    // for client-side builds. (This will be the default starting in
+    // `@sentry/nextjs` version 8.0.0.) See
+    // https://webpack.js.org/configuration/devtool/ and
+    // https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#use-hidden-source-map
+    // for more information.
+    hideSourceMaps: true,
+  },
 };
 
 const sentryWebpackPluginOptions = {


scripts/NextJs/configs/next.config.js
```diff
@@ -7,18 +7,34 @@ const { withSentryConfig } = require('@sentry/nextjs');
 
 const moduleExports = {
   // Your existing module.exports
+
+  sentry: {
+    // Use `hidden-source-map` rather than `source-map` as the Webpack `devtool`
+    // for client-side builds. (This will be the default starting in
+    // `@sentry/nextjs` version 8.0.0.) See
```
lforst, line 14: > (This will be the default starting in `@sentry/nextjs` version 8.0.0.)
 
Wondering if this is just too much information for people. Telling people this should not change how they're configuring or using stuff. IMO it's fine to just tell them in v8 to remove this option (like you wrote in the comment in https://github.com/getsentry/sentry-javascript/pull/5649). No hard opinion though.

scripts/NextJs/configs/next.config.js
```diff
@@ -7,18 +7,34 @@ const { withSentryConfig } = require('@sentry/nextjs');
...
+  // Keep in mind that the following options are set automatically, and
+  // overriding them is not recommended unless you know what you're doing:
+  //   authToken, configFile, dryRun, include, ignore, org, project, release,
+  //   stripPrefix, url, urlPrefix,
+
+  // (If you do want to override which files are included and ignored, don't use
+  // `ignore` or `ignoreFile` directly. Instead, pass an array of objects of the
+  // form { paths: ['path/to/include'], ignore: ['path/to/ignore'] } to the
+  // `include` option. See
+  // https://github.com/getsentry/sentry-webpack-plugin#include.)
```
lforst, line 35: > If you do want to override which files are included and ignored

We should clarify for what exactly the files are included/ignored.

lobsterkatie: Wait, now I'm confused. I didn't add that part in this PR, at least not intentionally... I would have said it was already there... 🤔

scripts/NextJs/configs/next.config.js
lobsterkatie: I also don't feel strongly about it, but it's what I've seen other tools do, and I figure it lets people make their own `TODO` if they want to. I'm inclined to leave it in.

lforst: Sounds good!

lforst (approved):

Issue #181, opened 2022-08-01
Replace deprecated `upload-dsym` for RN-iOS and add support for the new one


marandaneto:
RN bits.
https://github.com/getsentry/sentry-wizard/blob/cd7e22a1b5351ed47823934ccf504a87c3d3e741/lib/Steps/Integrations/ReactNative.ts#L270-L271

Sentry CLI's PR.
https://github.com/getsentry/sentry-cli/releases/tag/2.0.2

First we have to bump Sentry-CLI as well, https://github.com/getsentry/sentry-react-native/pull/2385

marandaneto:
We likely need to have a plugin option to either pass or not `--include-sources` which is the subcommand responsible for uploading the source context bundle.

PR #199, opened 2022-10-17
Kw/update rn sentry cli commands

krystofwoldrich:
Closes: https://github.com/getsentry/sentry-wizard/issues/181

Relates to CLI update in RN SDK https://github.com/getsentry/sentry-react-native/issues/2246 and https://github.com/getsentry/sentry-react-native/pull/2457

### DIFF
--- a/lib/Steps/Integrations/ReactNative.ts
+++ b/lib/Steps/Integrations/ReactNative.ts
@@ -267,7 +267,7 @@ export class ReactNative extends MobileProject {
 
   private _addNewXcodeBuildPhaseForSymbols(buildScripts: any, proj: any): void {
     for (const script of buildScripts) {
-      if (script.shellScript.match(/sentry-cli\s+upload-dsym/)) {
+      if (script.shellScript.match(/sentry-cli\s+(upload-dsym|debug-files upload)/)) {
         return;
       }
     }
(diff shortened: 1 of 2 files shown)


krystofwoldrich:
The `sentry.properties` file in `ios` dir would like like this:

```
debug_files.upload.include_sources=true
```

lib/Steps/Integrations/ReactNative.ts
```diff
@@ -279,9 +279,13 @@ export class ReactNative extends MobileProject {
       null,
       {
         shellPath: '/bin/sh',
-        shellScript:
-          'export SENTRY_PROPERTIES=sentry.properties\\n' +
-          '../node_modules/@sentry/cli/bin/sentry-cli upload-dsym',
+        shellScript:`
+export SENTRY_PROPERTIES=sentry.properties
+INCLUDE_SOURCES_KEY=debug_files.upload.include_sources
```
marandaneto, line 284: what if we implement that directly on Sentry CLI? does it make sense? Sentry CLI already loads the sentry.properties and uses its configurations, maybe we just need to add support for it in there and it'd work for Android as well?

krystofwoldrich: The values from the `properties` file are in the cli, so implementation would be simple if statement.

But I'm not sure if a non-global attribute should be passed thru the `properties`. 

Here it makes sense to me because I'm looking at the properties as settings of the Xcode plugin. 

The android plugin uses the `properties` directly. Also, it uses a different command (react-native gradle), so I'm not sure if it will even need this flag.

krystofwoldrich:
I've simplified this to env `SENTRY_INCLUDE_NATIVE_SOURCES`. This way it should be simple enough to opt-in for the feature and it won't affect the cli properties file.

lib/Steps/Integrations/ReactNative.ts
```diff
@@ -279,9 +279,11 @@ export class ReactNative extends MobileProject {
       null,
       {
         shellPath: '/bin/sh',
-        shellScript:
-          'export SENTRY_PROPERTIES=sentry.properties\\n' +
-          '../node_modules/@sentry/cli/bin/sentry-cli upload-dsym',
+        shellScript:`
+export SENTRY_PROPERTIES=sentry.properties
+[[ $SENTRY_INCLUDE_NATIVE_SOURCES == "true" ]] && INCLUDE_SOURCES_FLAG="--include-sources" || INCLUDE_SOURCES_FLAG=""
+../node_modules/@sentry/cli/bin/sentry-cli debug-files upload "$INCLUDE_SOURCES_FLAG"
```
marandaneto, line 285: Double-check that sentry-wizard writes the file properly and it executes correctly, sometimes because of the char skipping, it breaks.

marandaneto:
We can specify this on the docs as well https://docs.sentry.io/platforms/react-native/upload-debug/
How to export the var or add it manually via the UI on Xcode.

marandaneto (approved):

Issue #208, opened 2022-11-07
[@sentry/cli] Path resolution to `@sentry/cli` is unreliable


lobsterkatie:
As reported by Cramer, following our stock nextjs SDK instructions to use the wizard through `npx` led to `@sentry/cli` being resolved to `../../.npm/_npx/a8388072043b4cbc/node_modules/@sentry/cli/bin/sentry-cli` rather than the copy installed in the project's `node_modules` folder. (This then gets inserted in the `sentry.properties` file the wizard generates.)

- [ ] Figure out why this is happening
- [ ] Perhaps first decide if we even need to include the path to the executable in `sentry.properties` when dealing with `@sentry/cli`. It should download the binary to a standard place, so we might not need this at all.

smeubank:
@kamilogorek this something you might have an idea about off the top of your head?

lobsterkatie:
Related: https://github.com/getsentry/sentry-cli/issues/915

lobsterkatie:
Per @kamilogorek  in slack:

> fwiw `npx` doesnt install its traasnsitive dependencies locally, so as long as someone has `NODE_PATH` set to follow their `volta/yarn/npm/whatever` node_modules, it will never work, due to `@sentry/cli` being installed inside `.npx` directory

`@sentry/cli` should be installed as a dependency of `@sentry/nextjs`. We’re working on making it not a dependency of the wizard: https://github.com/getsentry/sentry-wizard/issues/201.

vladanpaunovic:
Fixed in release https://github.com/getsentry/sentry-wizard/releases/tag/v2.3.1

Thanks @timfish!

PR #209, opened 2022-11-09
fix: Always check for both, nextjs and sdk version

kamilogorek:
Prior to that change, `@sentry/nextjs` check was never called if `next` was there, as `true || fn()` will not call `fn`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


timfish:
I don't think this PR fixes the linked issue since #208 seems to be caused by `require.resolve` when the wizard is called by `nxp`. 

The CLI path is wrong [in Remix projects](https://github.com/getsentry/vanguard/blob/main/sentry.properties) too.

kamilogorek:
Right, had few tabs opened and linked it here by accident.

AbhiPrasad (approved):

PR #210, opened 2022-11-09
feat(react-native): Xcode plugin includes collect modules script

krystofwoldrich:
context:
- https://github.com/getsentry/sentry-react-native/pull/2606

Maybe this should go to v1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- feat(react-native): Xcode plugin includes collect modules script
+- feat(react-native): Xcode plugin includes collect modules script ([#210](https://github.com/getsentry/sentry-wizard/pull/210))
 
 ## 2.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - Raise nextjs version limit to include 13 (#206)
 - feat(react-native): Xcode plugin includes collect modules script ([#210](https://github.com/getsentry/sentry-wizard/pull/210))
 
+## 1.4.0
+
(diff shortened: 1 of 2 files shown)


marandaneto:
If I got this right, this is based on the current wizard v2, which patches RN >= 0.69 which is the v5 alpha.
https://github.com/getsentry/sentry-react-native/pull/2606 is based on main, which is RN <= 0.68.
So this change has to be either on both wizards v1 and v2, or https://github.com/getsentry/sentry-react-native/pull/2606 has to make it into v5 alpha as well before merging this, and releasing both versions synced with RN releases.

krystofwoldrich:
Added v1 patch https://github.com/getsentry/sentry-wizard/pull/213

marandaneto:
Be sure to test the install and uninstall process directly invoking wizard rather than copy+paste the bash script.
What I'd do is: create an empty project using RN template, run wizard, test it by compiling and guaranteeing that the whole setup works (source map + packages upload on events), uninstall, compile again, etc, if this is tested, LGTM.

marandaneto (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 - Raise nextjs version limit to include 13 (#206)
 - feat(react-native): Xcode plugin includes collect modules script ([#210](https://github.com/getsentry/sentry-wizard/pull/210))
 
+## 1.4.0
+
+- feat(react-native): Xcode plugin includes collect modules script ([#213](https://github.com/getsentry/sentry-wizard/pull/213))
```
marandaneto, line 10: I'd only merge after being released tho.

krystofwoldrich: Yes, yes.

PR #211, opened 2022-11-09
fix: `cli.executable` should be resolved from cwd

### LINKED ISSUE #208: [@sentry/cli] Path resolution to `@sentry/cli` is unreliable

timfish:
Closes #208 

To test this PR I:
- Built and packed the wizard `yarn build && npm pack`
- Installed via `npx sentry-wizard-2.0.1.tgz`
- Created a new Next.js project and followed the docs instructions
- When it came to running the wizard I ran `~/.npm/_npx/f5b3fa0ab77970d2/node_modules/@sentry/wizard/dist/bin.js -i nextjs`

`sentry.properties` ends up with:
```ini
cli.executable=node_modules/@sentry/cli/bin/sentry-cli
```

This property isn't actually used for many of the wizard targets so I could have excluded it but it seemed like it would be better to fix this to work with `npx`.


lobsterkatie (approved):
Nice. Glad it was such an easy fix!

AbhiPrasad (approved):

PR #213, opened 2022-11-28
feat(react-native): Xcode plugin includes collect modules script

krystofwoldrich:
The same as https://github.com/getsentry/sentry-wizard/pull/210

For RN < 0.69

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,10 +13,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master, 1.x ]
+    branches: [ master ]
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [ master, 1.x ]
+    branches: [ master ]
   schedule:
     - cron: '31 0 * * 5'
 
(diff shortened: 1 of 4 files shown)


marandaneto:
I'd rather call `patch/v1.4` instead of `1.x` branch as we talked about before.
You'd need to add the changelog later back into `master` as well since you can't merge the branch into `master`.

marandaneto:
Similar comment https://github.com/getsentry/sentry-wizard/pull/210#issuecomment-1329097376
If this is tested, LGTM.

.github/workflows/codeql-analysis.yml
```diff
@@ -13,10 +13,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, 1.x ]
```
marandaneto, line 16: If you make a draft PR from `1.x` to `master` (after merging), all the workflows would run and this is not necessary.
So you can either keep this and later remove the 1.x from workflows, or merge without adding the 1.x workflow but open a draft PR to check that everything is green.

marandaneto (approved):

krystofwoldrich:
It's tested both the v1 and v2.

jakobsen9:
Hi,

The added collect modules script fails my build on RN 0.64.4. I get an error saying "@sentry/react-native/scripts/collect-modules.sh: No such file or directory". Any ideas why it fails? It seems above that you tested properly?

PR #214, opened 2022-11-28
patch/v1.4

krystofwoldrich:


AbhiPrasad:
Note due to the way craft works atm, when we backport changes we'll have to release the lower version first, and then the higher version, otherwise npm's `package-name@next` will refer to the lower version.

krystofwoldrich:
@AbhiPrasad Thanks for the warning. So first I release 1.4.0 and then 2.4.0.

PR #219, opened 2022-12-09
feat(nextjs): Add logic to add `@sentry/nextjs` if it's missing

scefali:
This PR adds logic to add the `@sentry/nextjs` package if it's not installed and we either have a `yarn.lock` or `package-lock.json` file. The purpose of this change is to enable a more seamless installation process where the user doesn't require the additional step of installing `@sentry/nextjs`.

Note that this only works if there is a lock file because we need that to know which package manager to install. IMO this is probably OK. I would imagine most folks would already have that file if they are thinking about adding Sentry. I think it's unlikely that people at Sentry before even installing packages and running whatever they are using locally.

Also note that this means the `nextjs` installation works slightly differently than the other platforms. In the future, we can back propagate these changes if we find them to be helpful.

- [ ] Adding tests
- [x] Update changelog

### DIFF
--- a/lib/Steps/Integrations/NextJs.ts
+++ b/lib/Steps/Integrations/NextJs.ts
@@ -13,6 +13,8 @@ import { debug, green, l, nl, red } from '../../Helper/Logging';
 import { SentryCli, SentryCliProps } from '../../Helper/SentryCli';
 import { BaseIntegration } from './BaseIntegration';
 
+type PackageManager = 'yarn' | 'npm' | 'pnpm';
+
 const COMPATIBLE_NEXTJS_VERSIONS = '>=10.0.8 <14.0.0';
 const COMPATIBLE_SDK_VERSIONS = '>=7.3.0';
 const PROPERTIES_FILENAME = 'sentry.properties';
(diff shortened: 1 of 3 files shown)


lib/Steps/Integrations/NextJs.ts
```diff
@@ -303,6 +319,32 @@ export class NextJs extends BaseIntegration {
...
+
+  private _getPackageMangerChoice(): 'yarn' | 'npm' | null {
+    if (fs.existsSync(path.join(process.cwd(), 'yarn.lock'))) {
+      return 'yarn';
+    }
+    if (fs.existsSync(path.join(process.cwd(), 'package-lock.json'))) {
+      return 'npm';
+    }
+    return null;
+  }
```
lforst, line 336: I recommend we add a case for [pnpm](https://pnpm.io/) here too since it seems very popular in Next.js community.

scefali:
@lforst Added pnpm support. I was planning on writing some tests but I realize we don't have any tests at all for nextjs 😬

lib/Steps/Integrations/NextJs.ts
```diff
@@ -303,6 +321,48 @@ export class NextJs extends BaseIntegration {
...
+        throw new Error(`Unknown package manager: ${packageManager}`);
+    }
+  }
+
+  private async _installPackage(
+    packageName: string,
+    packageManager: PackageManager,
+  ): Promise<void> {
+    const command = this._getInstallCommand(packageManager);
+    await promisify(exec)(`${command} ${packageName}`);
```
lobsterkatie, line 361: L: Any reason not to just use `execSync` here?

lobsterkatie (approved):
Good change!

lib/Steps/Integrations/NextJs.ts
scefali: @lobsterkatie In case you want to parallelize it and have it run faster

PR #221, opened 2022-12-13
feat(onboarding): Add project_platform as query param if -s and -i are set

scefali:
This PR adds the `project_platform` query parameter if we set the `signup` and `integration` option when running the wizard. The purpose of this is so we can automatically create that project during signup as implemented [here](https://github.com/getsentry/getsentry/pull/9122).

Note we shouldn't merge this until all the dependent PRs are merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - adds logic to add @sentry/nextjs if it's missing when running the wizard ([#219](https://github.com/getsentry/sentry-wizard/pull/219))
+- adds project_platform as query param if -s and -i are set ([#221](https://github.com/getsentry/sentry-wizard/pull/221))
 
 
 ## 2.4.0
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Options:
                                              [array] [choices: "ios", "android"]
   -u, --url          The url to your Sentry installation
                      env: SENTRY_WIZARD_URL      [default: "https://sentry.io/"]
+  -s, --signup       Redirect to signup page if not logged in          [boolean]
 ```
 
 ## Resources
--- a/README.md
+++ b/README.md
@@ -73,7 +73,8 @@ Options:
                                              [array] [choices: "ios", "android"]
   -u, --url          The url to your Sentry installation
                      env: SENTRY_WIZARD_URL      [default: "https://sentry.io/"]
-  -s, --signup       Redirect to signup page if not logged in          [boolean]
+  -s, --signup       Redirect to signup page if not logged in
+                     Use if don't have a Sentry account                [boolean]
 ```
 
 ## Resources
--- a/lib/Steps/OpenSentry.ts
+++ b/lib/Steps/OpenSentry.ts
@@ -1,4 +1,5 @@
 import { Answers } from 'inquirer';
+import { URL } from 'url';
 
 import { mapIntegrationToPlatform } from '../Constants';
 import { BottomBar } from '../Helper/BottomBar';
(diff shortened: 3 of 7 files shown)


lib/Steps/OpenSentry.ts
```diff
@@ -29,7 +30,16 @@ export class OpenSentry extends BaseStep {
...
+      let urlToOpen = `${baseUrl}account/settings/wizard/${data.hash}/`;
+      if (this._argv.signup) {
+        urlToOpen += '?signup=1';
+        // integration maps to platform in the wizard
+        if (this._argv.integration) {
+          urlToOpen += `&project_platform=${mapIntegrationToPlatform(
+            this._argv.integration,
+          )}`;
+        }
+      }
```
lforst, line 42: This is fine for now I think but should this grow in complexity let's use [`new URL` + `URLSearchParams`](https://nodejs.org/api/url.html#class-urlsearchparams)

lib/Steps/SentryProjectSelector.ts
```diff
@@ -22,6 +26,9 @@ export class SentryProjectSelector extends BaseStep {
     let selectedProject = null;
     if (answers.wizard.projects.length === 1) {
       selectedProject = { selectedProject: answers.wizard.projects[0] };
+      // the wizard CLI closes too quickly when we skip the prompt
+      // as it will cause the UI to be stuck saying Waiting for wizard to connect
+      await sleep(1000);
```
lforst, line 31: This feels a bit flakey to me. How do we know that 1s is enough? I am not too familiar with the execution flow of the wizard but is there a way to wait more reliably here?

README.md
```diff
@@ -73,6 +73,7 @@ Options:
                                              [array] [choices: "ios", "android"]
   -u, --url          The url to your Sentry installation
                      env: SENTRY_WIZARD_URL      [default: "https://sentry.io/"]
+  -s, --signup       Redirect to signup page if not logged in          [boolean]
```
lforst, line 76: Can we explain a bit better what this means exactly? Redirect when? Why?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - feat: Add logic to add @sentry/nextjs if it's missing when running the wizard (#219)
 - fix: Print localhost with `http` instead of `https` (#212)
+- feat: Add project_platform as query param if -s and -i are set (#221)
```
lforst, line 7: A more general question. Why is redirecting to the login not the default behavior?

lib/Steps/SentryProjectSelector.ts
scefali: @lforst I think the issue is basically waiting for the web page to load. As far as I know, there's no way we can know when that actually opens.

lforst: The wizard could host a webserver locally that we try to ping from the website via localhost. Just an idea tho. Obviously there would need to be some kind fallback if that's not possible. For example if the wizard runs on some kind of remote machine.

scefali: @lforst Yea it's an idea. If this simple sleep doesn't always work we can evaluate more complicated and more robust solutions.

CHANGELOG.md
lforst: Ah I confused singup for login. Gotcha.

lforst (approved):

lib/Steps/OpenSentry.ts
scefali: @lforst is that even available with node 16?

PR #222, opened 2023-01-04
automatic merging of next.config.js files

roggenkemper:
This pr adds in automatic merging of the next.config.js files. If there is no existing next.config.js file, then we will add the existing default next.config.js file in. If we are able to merge Sentry's with the existing, then the file will be updated (with the original saved as next.config.original.js). If we are unable to merge, we will create a template file and ask them to merge (the status quo approach) .

### DIFF
--- a/lib/Helper/MergeConfig.ts
+++ b/lib/Helper/MergeConfig.ts
@@ -1,0 +1,165 @@
+import * as fs from 'fs';
+import * as ts from 'ts';
+
+// checks if statement uses module.exports format
+function isModuleExport(statement: ts.Statement) {
+  if (statement.kind !== ts.SyntaxKind.ExpressionStatement) {
+    return false;
+  }
+  if (!statement.expression) {
+    return false;
+  }
+  const { left } = statement.expression;
+  // check if this is module.exports
+  return (
+    left.expression?.escapedText === 'module' &&
+    left.name?.escapedText === 'exports'
+  );
+}
+
+// checks if statement uses default export format
+function isDefaultExport(statement: ts.Statement) {
+  if (statement.kind !== ts.SyntaxKind.ExportAssignment) {
+    return false;
+  }
+  if (!statement.expression) {
+    return false;
+  }
+  return true;
+}
+
+// finds the the source variable name from the statement
+function getSourceVariableName(statement: ts.Statement, sourceText: string) {
+  let sourceVariableName, declarationText;
+  if (isModuleExport(statement)) {
+    const { right } = statement.expression;
+
+    // modules.exports = () => ({...})
+    if (
+      right?.kind === ts.SyntaxKind.ArrowFunction ||
+      right?.kind === ts.SyntaxKind.FunctionExpression
+    ) {
+      ({ sourceVariableName, declarationText } = parseFunction(
+        statement,
+        sourceText,
+      ));
+    }
+    // module.exports = {...}
+    else {
+      sourceVariableName = right?.escapedText;
+    }
+  }
+  // default export {...}
+  else if (isDefaultExport(statement)) {
+    sourceVariableName = statement.expression.escapedText;
+  }
+  return { variableName: sourceVariableName, text: declarationText };
+}
+
+// merges the config files
+function mergeOutput({
+  topText,
+  declarationText,
+  sourcePath,
+  templatePath,
+}: {
+  topText: string;
+  declarationText: string;
+  sourcePath: string;
+  templatePath: string;
+}) {
+  const baseFile = fs.readFileSync(templatePath, 'utf8');
+  const newText = baseFile
+    .replace('// INSERT TOP TEXT', topText)
+    .replace('// INSERT CONFIG TEXT', declarationText);
+  fs.writeFileSync(sourcePath, newText);
+}
+
+// parses the function statement for source variable name and declaration text
+function parseFunction(funcStatement: ts.Statement, sourceText: string) {
+  const { right } = funcStatement.expression;
+
+  let sourceVariableName, declarationText;
+
+  const functionStatement = right?.body.statements;
+  if (functionStatement) {
+    sourceVariableName =
+      functionStatement[functionStatement.length - 1].expression.escapedText;
+
+    if (functionStatement[0].declarationList) {
+      declarationText = getDeclaration(
+        functionStatement[0].declarationList,
+        sourceText,
+      );
+    }
+  }
+  return { sourceVariableName, declarationText };
+}
+
+// gets the declaration text from non-function statements
+function getDeclaration(
+  declarationList: ts.VariableDeclarationList,
+  sourceText: string,
+) {
+  const declaration = declarationList.declarations[0];
+  const { initializer } = declaration;
+  const text = sourceText.substring(initializer.pos, initializer.end).trim();
+  if (text[0] === '{' && text[text.length - 1] === '}') {
+    return text.substring(1, text.length - 1);
+  }
+}
+
+// merges the config files
+export function mergeConfigFile(sourcePath: string, templatePath: string) {
+  const node = ts.createSourceFile(
+    'sourceText.ts', // fileName
+    fs.readFileSync(sourcePath, 'utf8'), // sourceText
+    ts.ScriptTarget.Latest, // langugeVersion
+  );
+
+  const sourceText = node.text;
+
+  let topText = '',
+    declarationText,
+    sourceVariableName;
+
+  for (const statement of node.statements) {
+    const { variableName, text } = getSourceVariableName(statement, sourceText);
+    if (text) {
+      declarationText = text;
+    }
+    if (variableName) {
+      sourceVariableName = variableName;
+    }
+
+    if (statement.declarationList) {
+      declarationText = getDeclaration(statement.declarationList, sourceText);
+    }
+
+    if (statement.kind === ts.SyntaxKind.ImportDeclaration) {
+      let defaultImport = '',
+        namedImports = '',
+        separator = '';
+
+      if (statement.importClause.name) {
+        defaultImport = statement.importClause.name.escapedText;
+      }
+      if (statement.importClause.namedBindings) {
+        namedImports = `{${statement.importClause.namedBindings.elements
+          .map((el: ts.ImportSpecifier) => el.name.escapedText)
+          .join(', ')}}`;
+      }
+      if (defaultImport && namedImports) {
+        separator = ', ';
+      }
+      topText = topText.concat(
+        `import ${defaultImport}${separator}${namedImports} from "${statement.moduleSpecifier.text}";\n`,
+      );
+    }
+  }
+  if (declarationText) {
+    mergeOutput({ topText, declarationText, sourcePath, templatePath });
+    return true;
+  }
+  return false;
+}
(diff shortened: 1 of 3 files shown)


lib/Helper/MergeConfig.ts
```diff
@@ -0,0 +1,185 @@
...
+
+// merges the config files
+export function mergeConfigFile(
+  sourcePath: string,
+  templatePath: string,
+): boolean {
+  const node = ts.createSourceFile(
+    'sourceText.ts', // fileName
+    fs.readFileSync(sourcePath, 'utf8'), // sourceText
+    ts.ScriptTarget.Latest, // langugeVersion
```
scefali, line 133:
```suggestion
    ts.ScriptTarget.Latest, // languageVersion
```

lib/Helper/MergeConfig.ts
```diff
@@ -0,0 +1,185 @@
...
+        statement,
+        sourceText,
+      ));
+    }
+    // module.exports = {...}
+    else {
+      sourceVariableName = right?.escapedText;
+    }
+  }
+  // default export {...}
```
scefali, line 55: @roggenkemper these kind of comments about which syntax you're looking for are great! Having them in the other methods would make this much easier to understand

scefali (commented):
@roggenkemper I pulled a random example online and it doesn't look like we can handle it, is it expected for this config?
```
module.exports = {
  images: {
    domains: [],
  },
}
```

lforst:
Just jumping in to give some early feedback. From our experience, mucking with the AST is super super fragile. If I were to tackle this I would probably look into some more robust (but uglier) solutions first.

We could even change the signature and functionality of `withSentryConfig` (in a backwards compatible way) to make this easier.

scefali:
@lforst What kinds of robust solutions are you thinking of? We are open to ideas. That being said, I think this approach could do the job. Just need to write about ~10 unit tests to cover different formats that we commonly see and I think this will work 50-90% of the time.  Even if we can't make this work 100% of the time...anything better than 0% is an improvement over today.

roggenkemper:
made some updates based off the conversation with Luca - the flow now is to not deal with AST's, but instead essentially copy over their original config, but add in the call with withSentryConfig which will be updated (https://github.com/getsentry/sentry-javascript/pull/6721) to allow us to add in the Sentry options that we need to add into their config.

lib/Helper/MergeConfig.ts
```diff
@@ -0,0 +1,17 @@
...
+// merges the config files
+export function mergeConfigFile(
+  sourcePath: string,
+  templatePath: string,
+): boolean {
+  try {
+    const templateFile = fs.readFileSync(templatePath, 'utf8');
+    const sourceFile = fs.readFileSync(sourcePath, 'utf8');
+    const newText = templateFile.replace('// ORIGINAL CONFIG', sourceFile);
+    fs.writeFileSync(sourcePath, newText);
```
scefali, line 12: @roggenkemper can we also try to parse the file and make sure it's valid JS?

lib/Steps/Integrations/NextJs.ts
```diff
@@ -286,23 +289,73 @@ export class NextJs extends BaseIntegration {
...
-      if (!fs.existsSync(destinationPath)) {
-        this._fillAndCopyTemplate(templatePath, destinationPath, dsn);
-      } else if (!fs.existsSync(mergeableFilePath)) {
-        this._fillAndCopyTemplate(templatePath, mergeableFilePath, dsn);
-        red(
-          `File \`${templateFile}\` already exists, so created \`${mergeableFilePath}\`.\n` +
-            'Please merge those files.',
-        );
-        nl();
+      if (templateFile === 'next.config.js') {
```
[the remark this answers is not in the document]
roggenkemper: is this an edge case that needs to be handled? it seems rare and currently we would create a next.config.js file that they would be able to merge.

lib/Helper/Configs/next.config.1.js
```diff
@@ -0,0 +1,6 @@
+/** @type {import('next').NextConfig} */
```
scefali, line 1: @roggenkemper Are these only used in tests? If so, can we put configs in  it in __tests__?

scefali (approved):
Let's get someone from SDK's to approve but looks good from my end

lib/Helper/MergeConfig.ts
```diff
@@ -0,0 +1,18 @@
...
+// merges the config files
+export function mergeConfigFile(
+  sourcePath: string,
+  templatePath: string,
+): boolean {
+  try {
+    const templateFile = fs.readFileSync(templatePath, 'utf8');
+    const sourceFile = fs.readFileSync(sourcePath, 'utf8');
+    const newText = templateFile.replace('// ORIGINAL CONFIG', sourceFile);
+    new Function(newText); // check if the file is valid javascript
```
lforst, line 12: I don't think this will throw if you pass in invalid JS. I would just remove this line.

lib/Steps/Integrations/NextJs.ts
```diff
@@ -437,4 +457,58 @@ export class NextJs extends BaseIntegration {
...
+        destinationDir,
+        this._spliceInPlace(templateFile.split('.'), -1, 0, 'original').join(
+          '.',
+        ),
+      );
+      // makes copy of original next.config.js
+      fs.writeFileSync(originalFilePath, fs.readFileSync(destinationPath));
+      await this._addToGitignore(
+        originalFilePath,
+        'Unable to next.config.original.js add to gitignore',
```
lforst, line 485: I think this message has a wording mistake.

lib/Steps/Integrations/NextJs.ts
```diff
@@ -437,4 +457,58 @@ export class NextJs extends BaseIntegration {
...
+      fs.copyFileSync(templatePath, destinationPath);
+      green('Created File `next.config.js`');
+      nl();
+    } else {
+      const originalFilePath = path.join(
+        destinationDir,
+        this._spliceInPlace(templateFile.split('.'), -1, 0, 'original').join(
+          '.',
+        ),
+      );
```
lforst, line 480: Can we add a comment what's happening here. This reads a bit messy.

lib/Steps/Integrations/NextJs.ts
```diff
@@ -437,4 +457,58 @@ export class NextJs extends BaseIntegration {
...
+          `Updated \`next.config.js\` with Sentry. The original next.config.js was saved as \`next.config.original.js\`.\n` +
+            "Information about Sentry's changes can be found at https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/",
+        );
+        nl();
+      } else {
+        // if merge fails, we'll create a copy of the `next.config.js` template and ask them to merge
+        fs.copyFileSync(templatePath, mergeableFilePath);
+        await this._addToGitignore(
+          mergeableFilePath,
+          'Unable to next.config.js template add to gitignore',
```
lforst, line 504: I think there is also a wording mistake here.

lib/Steps/Integrations/NextJs.ts
```diff
@@ -437,4 +457,58 @@ export class NextJs extends BaseIntegration {
...
+    destinationPath: string,
+    templatePath: string,
+    destinationDir: string,
+    templateFile: string,
+    configDirectory: string,
+    mergeableFilePath: string,
+  ): Promise<void> {
+    // if no next.config.js exists, we'll create one
+    if (!fs.existsSync(destinationPath)) {
+      fs.copyFileSync(templatePath, destinationPath);
```
lforst, line 471: The written file is gonna be a bit ugly because of the `// ORIGINAL CONFIG` comment. Can we either replace it with something a little more instructive or remove it before writing it to disk?

We should probably replace it even with

```
module.exports = {
  // Your Next.js configuration here
}
```

or smth like that.

lib/Steps/Integrations/NextJs.ts
```diff
@@ -437,4 +457,58 @@ export class NextJs extends BaseIntegration {
...
+      );
+
+      const mergedTemplatePath = path.join(
+        configDirectory,
+        'next.config.template.js',
+      );
+      // attempts to merge with existing next.config.js, if true -> success
+      if (mergeConfigFile(destinationPath, mergedTemplatePath)) {
+        green(
+          `Updated \`next.config.js\` with Sentry. The original next.config.js was saved as \`next.config.original.js\`.\n` +
```
lforst, line 495: This logmessage should probably use the `templateFile` variable if we're going so far as passing it to the function.

lib/Steps/Integrations/NextJs.ts
```diff
@@ -437,4 +457,58 @@ export class NextJs extends BaseIntegration {
...
+            "Information about Sentry's changes can be found at https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/",
```
lforst, line 496:
```suggestion
Information on the changes made to the Next.js configuration file can be found at https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/
```

Don't know if that reads better but to me "Information about Sentry's changes" sounds a bit ambiguous.

lib/Steps/Integrations/NextJs.ts
```diff
@@ -437,4 +457,58 @@ export class NextJs extends BaseIntegration {
...
+        nl();
+      } else {
+        // if merge fails, we'll create a copy of the `next.config.js` template and ask them to merge
+        fs.copyFileSync(templatePath, mergeableFilePath);
+        await this._addToGitignore(
+          mergeableFilePath,
+          'Unable to next.config.js template add to gitignore',
+        );
+        red(
+          `Unable to merge  \`next.config.js\`, so created \`${mergeableFilePath}\`.\n` +
```
lforst, line 507: This message should probably also use the `templateFile` variable

lib/Steps/Integrations/NextJs.ts
```diff
@@ -437,4 +457,58 @@ export class NextJs extends BaseIntegration {
...
+            'Please merge those files.',
```
lforst, line 508: We could maybe be a little bit more instructive here. How about "Please integrate next.config.wizardcopy.js into your next.config.js or next.config.ts file".

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- feat: Automatic merging of next.config.js files (#222)
```
lforst, line 5:
```suggestion
- feat: Merge next.config.js files automatically (#222)
```

lib/Helper/Configs/next.config.1-merged.js
```diff
@@ -0,0 +1,18 @@
+// This file sets a custom webpack configuration to use your Next.js app
```
lforst, line 1: Can we rename the folder these test templates are located to something like `test-fixtures`?

lforst (commented):
Nothing major but definitely some stuff to adress before we release this.

lib/Steps/Integrations/NextJs.ts
roggenkemper: If no next.config.js exists, then it will use the existing `next.config.js` template instead of the new one with the `// ORIGINAL CONFIG` line.

lib/Helper/MergeConfig.ts
lforst: Ah you're right. I messed up how I verified this. Sorry about that!

lib/Steps/Integrations/NextJs.ts
lforst: Ah you're right. I hope this isn't just me but to me the entire logic of this file is super hard to get behind. We should def clean this up some time (not your code but the one that existed before - it probably just grew over time).

lforst (approved):

PR #223, opened 2023-01-06
feat(promotions): promo code option

scefali:
This PR adds an option called `--promo-code` that will pre-fill in the promo code field during signup as shown below. Note this argument only means anything if the `-s` signup argument is passed and the user is currently not logged in. We intend to use this option for specific campaigns to give folks discounts when signing up from this flow.

Screen Shot 2023-01-06 at 1 10 03 PM


lforst (approved):
What is the end goal here? Is it that we will have some copy paste snippet with the promo code somewhere?

scefali:
@lforst Yea basically. We'll use it for partnerships.

PR #237, opened 2023-02-22
Fix the typo in the sentry option array

nlopin:
According to TS types and documentation, the spelling of the option is `hideSourceMaps`


nlopin:
Haven't noticed there is a PR with the same change

Issue #244, opened 2023-04-17
Support SvelteKit SDK Setup


Lms24:
The Sentry Wizard should support setting up the SvelteKit SDK, similarly to setting up the NextJS SDK.

```[tasklist]
### Tasks

(this is subject to change as I have very little context on the wizard yet)

- [ ] https://github.com/getsentry/sentry-wizard/pull/245
- [ ] https://github.com/getsentry/sentry-wizard/pull/246
- [ ] #249
- [ ] https://github.com/getsentry/sentry-wizard/issues/247
- [ ] #257
- [ ] https://github.com/getsentry/sentry-wizard/pull/259
- [ ] https://github.com/getsentry/sentry-wizard/pull/260
- [ ] https://github.com/getsentry/sentry-wizard/pull/266
- [ ] https://github.com/getsentry/sentry-wizard/pull/267
- [ ] https://github.com/getsentry/sentry-wizard/pull/268
```

ref https://github.com/getsentry/sentry-javascript/issues/6692

Lms24:
Removed the last task of wrapping `load` functions in favour of doing this at build time. We're not going to add a wizard-based solution at this time, as we're still waiting on a `handleLoad` hook for the final solution. The interim solution will be a build-time based auto-instrumentation similar to what we're doing in the NextJs SDK.

PR #245, opened 2023-04-17
feat(sveltekit): Add SvelteKit menu option and SDK installation

Lms24:
This PR marks the first step of adding Wizard support for the SvelteKit SDK:

* Add `SvelteKit` integration class
* SvelteKit can be selected in the integration selection prompt
* Support for `-i sveltekit` option
* Check if compatible SvelteKit version or SvelteKit SDK version is installed
* Install `@sentry/sveltekit` package if not yet installed

I'm not sure about the release cadence of the wizard. Happy to merge the steps into a branch for now and release in one go later on if we don't want the halfway-finished SvelteKit menu option to show up before it is finished. 

For now:
#skip-changelog

ref #244

### DIFF
--- a/lib/Helper/Package.ts
+++ b/lib/Helper/Package.ts
@@ -59,3 +59,20 @@ function fulfillsVersionRange(
       : satisfies(cleanedUserVersion, acceptableVersions))
   );
 }
+
+/**
+ * Determines if the passed `package.json` object has the passed package installed.
+ *
+ * @param appPackage The `package.json` object
+ * @param packageName The name of the package to check for
+ *
+ * @returns `true` if the package is installed, `false` otherwise
+ */
+export function hasPackageInstalled(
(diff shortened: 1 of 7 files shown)


lib/Helper/Package.ts
```diff
@@ -59,3 +59,20 @@ function fulfillsVersionRange(
...
+
+/**
+ * Determines if the passed `package.json` object has the passed package installed.
+ *
+ * @param appPackage The `package.json` object
+ * @param packageName The name of the package to check for
+ *
+ * @returns `true` if the package is installed, `false` otherwise
+ */
+export function hasPackageInstalled(
```
Lms24, line 71: Extracted this from the NextJS integration class. Will probably extract more methods going forward as the initial setup is very similar

AbhiPrasad (approved):

PR #256, opened 2023-04-20
ref: Rewrite Next.js wizard

lforst:
This PR rewrites the Next.js wizard to

- look a bit prettier
- have a clearer flow
- consider a few more cases
- generate more useful example code

Try it out using `npx @sentry/wizard@next -i -nextjs`.

May god have mercy on @abhiprasad who is reluctantly gonna review this PR.

PR #258, opened 2023-04-20
ref(nextjs): Clean up minor things

lforst:
S/o to @anonrig for pointing out a few things in his review of #256.

### DIFF
--- a/src/clack-utils.ts
+++ b/src/clack-utils.ts
@@ -35,9 +35,7 @@ export function abortIfCancelled<T>(
 ): asserts input is Exclude<T, symbol> {
   if (clack.isCancel(input)) {
     clack.cancel('Wizard setup cancelled.');
-    return process.exit(0);
-  } else {
-    return;
+    process.exit(0);
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/clack-utils.ts
```diff
@@ -81,7 +79,7 @@ export async function confirmContinueEvenThoughNoGitRepo(): Promise<void> {
     childProcess.execSync('git rev-parse --is-inside-work-tree', {
       stdio: 'ignore',
     });
-  } catch (e) {
+  } catch {
```
AbhiPrasad, line 82: Haha I now realized that optional catch binding was added in Node 10, so we can go through our SDK and update this also.

I guess the habit has stuck from working on the JS SDK codebase :)

AbhiPrasad (approved):

src/clack-utils.ts
lforst: Yeah caught me too. Either ts or eslint in the SDK codebase disallow this though iirc.

AbhiPrasad: eslint in the codebase disallows empty block statements, but using optional catch is fine.

you just need to do

```js
catch {
  // plz eslint don't scream at me
}
```

It's es2019 though so no way we can do in the browser packages 😭

PR #259, opened 2023-04-21
feat(sveltekit): Add Sentry handlers to `handle` and `handleError` hooks 

Lms24:
This PR adds our handlers to the `handle` and `handleError` hooks in the SveltKit hooks files. I tested this with various cases of already existing, user-defined handlers. 

One limtiation:
This currently only works for arrow function exports ( `export const handleError = () => {}`). The Kit docs examples always show that handlers should be exported with arrow functions but it is possible to add normal function exports (`export function handleError(){}`). In this case we just bail out and ask users to apply our handler manually. I believe this is a rather edge-casey scenario, so IMO it's fine to just bail at the moment.

#skip-changelog

ref #244

### DIFF
--- a/lib/Steps/Integrations/SvelteKit.ts
+++ b/lib/Steps/Integrations/SvelteKit.ts
@@ -81,6 +81,8 @@ export class SvelteKit extends BaseIntegration {
       return this._shouldConfigure;
     }
 
+    await this._createOrMergeSvelteKitFiles('somedes');
+
     nl();
 
     let userAnswers: Answers = { continue: true };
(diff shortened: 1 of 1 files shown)


lib/Steps/Integrations/SvelteKit.ts
```diff
@@ -408,6 +341,112 @@ export class SvelteKit extends BaseIntegration {
...
+      if (!declaration) {
+        return;
+      }
+      if (declaration.type === 'FunctionDeclaration') {
+        if (!declaration.id || declaration.id.name !== 'handleError') {
+          return;
+        }
+        foundHandleError = true;
+        yellow(
+          'Cannot safely wrap an `export function handleError` declaration. Please wrap it manually with `handleErrorWithSentry`',
```
AbhiPrasad, line 363: Why can't we wrap an export function handleError declaration?

Could we make the transform something like this?

```js
export const handleError = Sentry.handleErrorWithSentry(function _handleError() {
  // ...
});
```

Little more complicated for `Sentry.sentryHandle` because of sequence, so I'm fine to log out a warning there.

Lms24: We can, but I'm not sure if there are more problems involved if we convert a function export to an arrow function export. For instance the differences with `this` in the arrow vs normal functions?

Happy to take another look at this. Overall, I just want to avoid doing even more AST modification than we're currently doing. But your proposed transformation looks more reasonable to me than what I came up with before for this case😅 

As for handle, I'm just always adding `sequence`, regardless of if we're the only handler or if there are others. So I think the complexity shouldn't be much higher than for handleError

AbhiPrasad: I think if we keep the function declaration (with `function _handleError`) we should be fine - if this takes too much time though (more the 2 hours), it's probably not worth it.

AbhiPrasad (approved):
Let's merge this in - you can add the changes to wrapping function exports in another PR if you decide its worth the time - sorry for not LOGAFing.

Lms24:
>Let's merge this in - you can add the changes to wrapping function exports in another PR if you decide its worth the time - sorry for not LOGAFing.

All good! It's a valid concern! I'll take a look at it

PR #262, opened 2023-04-24
ref: Replace old Next.js wizard

lforst:
Replaces the old Next.js wizard so that the Next.js wizard still works when the command is invoked without -i flag.


AbhiPrasad (approved):
rip

PR #268, opened 2023-05-02
feat(sveltekit): Add example page creation

Lms24:
This PR adds the creation of a sample SvelteKit route with a page and an API route to send example frontend and backend errors (+ a transaction) to Sentry. 

Example of page in the default Sverdle app:

image

yes, the button might feature a nice animation ;)

#skip-changelog

ref #244


lforst (approved):
Nice!

PR #272, opened 2023-05-11
fix: Don't crash in environments without browser

11joselu:
Hello :)

We are using docker without browsers and there are an error when the cli tries to open the url automatically.

https://github.com/getsentry/sentry-wizard/assets/8685132/7eafc84f-833f-466c-b66d-cd89ad9eb615


lforst:
Thanks! :)

lforst (approved):

PR #309, opened 2023-06-28
feat: Add telemetry helper

lforst:
Adds a telemetry helper `withTelemetry` that can be used to send anonymous crash behaviour to Sentry.

We had to adjust the `abort` and `abortIfCanceled` functions and their usages to accomodate async flushing of Sentry events.

### DIFF
--- a/src/nextjs/nextjs-wizard.ts
+++ b/src/nextjs/nextjs-wizard.ts
@@ -95,13 +95,15 @@ export async function runNextjsWizard(
         existingConfigs.push(tsConfig);
       }
 
-      const overwriteExistingConfigs = await clack.confirm({
+      let overwriteExistingConfigs = await clack.confirm({
         message: `Found existing Sentry ${configVariant} config (${existingConfigs.join(
           ', ',
         )}). Overwrite ${existingConfigs.length > 1 ? 'them' : 'it'}?`,
       });
 
-      abortIfCancelled(overwriteExistingConfigs);
+      overwriteExistingConfigs = await abortIfCancelled(
+        overwriteExistingConfigs,
+      );
 
       shouldWriteFile = overwriteExistingConfigs;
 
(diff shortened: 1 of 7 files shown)


src/telemetry.ts
```diff
@@ -0,0 +1,77 @@
...
+function createSentryInstance(enabled: boolean, integration: string) {
+  const client = new NodeClient({
+    dsn: 'https://8871d3ff64814ed8960c96d1fcc98a27@o1.ingest.sentry.io/4505425820712960',
+    enabled: enabled,
+
+    tracesSampleRate: 1,
+    sampleRate: 1,
+
+    release: packageJson.version,
+    integrations: [],
```
Lms24, line 53: Since we're initializing the client here, we need to at least add some integrations back, right? GlobalHandlers, Http?

Lms24 (approved):
LGTM, see my comments

src/nextjs/nextjs-wizard.ts
```diff
@@ -95,13 +95,15 @@ export async function runNextjsWizard(
...
+      let overwriteExistingConfigs = await clack.confirm({
         message: `Found existing Sentry ${configVariant} config (${existingConfigs.join(
           ', ',
         )}). Overwrite ${existingConfigs.length > 1 ? 'them' : 'it'}?`,
       });
 
-      abortIfCancelled(overwriteExistingConfigs);
+      overwriteExistingConfigs = await abortIfCancelled(
+        overwriteExistingConfigs,
+      );
```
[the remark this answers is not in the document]
lforst: Ah much cleaner

src/telemetry.ts
lforst: Http: yes. GlobalHandlers I wouldn't add because we try catch in the wrapper anyhow.

PR #314, opened 2023-06-28
feat(sourcemaps): Add setup flow for sentry-cli

lforst:
Resolves https://github.com/getsentry/sentry-wizard/issues/312

Adds catch-all sourcemap flow that:
- Will install `@sentry/cli`
- Asks the user to provide a path to their build artifacts
- Asks the user to verify that their build tool is generating source maps
- Will create a `package.json` script `sentry:ci` that will inject debug IDs into build artifacts and upload the artifacts
- Create a `.sentryclirc` file with an auth token so that the user can try the whole flow locally

### DIFF
--- a/src/sourcemaps/tools/sentry-cli.ts
+++ b/src/sourcemaps/tools/sentry-cli.ts
@@ -28,7 +28,7 @@ export const configureSentryCLI: SourceMapUploadToolConfigurationFunction =
       const rawArtifactPath = await abortIfCancelled(
         clack.text({
           message:
-            'Provide a path to the folder that contains your build artifacts:',
+            'Where are your build artifacts located?',
           placeholder: `path${path.sep}to${path.sep}your${path.sep}build${path.sep}output`,
         }),
       );
--- a/src/sourcemaps/tools/sentry-cli.ts
+++ b/src/sourcemaps/tools/sentry-cli.ts
@@ -75,9 +75,15 @@ export const configureSentryCLI: SourceMapUploadToolConfigurationFunction =
     );
 
     packageDotJson.scripts = packageDotJson.scripts || {};
-    packageDotJson.scripts[
-      'sentry:ci'
-    ] = `sentry-cli sourcemaps inject --org ${options.orgSlug} --project ${options.projectSlug} ${relativePosixArtifactPath} && sentry-cli sourcemaps upload --org ${options.orgSlug} --project ${options.projectSlug} ${relativePosixArtifactPath}`;
+    packageDotJson.scripts['sentry:ci'] = `sentry-cli sourcemaps inject --org ${
+      options.orgSlug
+    } --project ${
+      options.projectSlug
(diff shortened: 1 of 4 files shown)


src/sourcemaps/tools/sentry-cli.ts
```diff
@@ -0,0 +1,120 @@
...
+      alreadyInstalled: hasPackageInstalled('@sentry/cli', packageDotJson),
+    });
+
+    let validPath = false;
+    let relativeArtifactPath;
+    do {
+      const rawArtifactPath = await abortIfCancelled(
+        clack.text({
+          message:
+            'Provide a path to the folder that contains your build artifacts:',
```
Lms24, line 31: m: Should we rephrase this as a question? 

For example
```suggestion
            'Where are your build artifacts located?',
```

src/sourcemaps/tools/sentry-cli.ts
```diff
@@ -0,0 +1,120 @@
...
+        )}`,
+        options: [{ label: 'I checked. Continue!', value: true }],
+        initialValue: true,
+      }),
+    );
+
+    packageDotJson.scripts = packageDotJson.scripts || {};
+    packageDotJson.scripts[
+      'sentry:ci'
+    ] = `sentry-cli sourcemaps inject --org ${options.orgSlug} --project ${options.projectSlug} ${relativePosixArtifactPath} && sentry-cli sourcemaps upload --org ${options.orgSlug} --project ${options.projectSlug} ${relativePosixArtifactPath}`;
```
Lms24, line 81: m: We need to pass the `url` here, too, for self-hosted, right?

src/sourcemaps/tools/sentry-cli.ts
```diff
@@ -0,0 +1,120 @@
...
+            ),
+          },
+        ],
+        initialValue: true,
+      }),
+    );
+
+    if (!addedToCI) {
+      clack.log.info("Don't forget! :)");
+    }
```
Lms24, line 117: Nice, we might wanna add this to the Vite flow 😅

src/sourcemaps/tools/sentry-cli.ts
```diff
@@ -0,0 +1,120 @@
...
+    });
+
+    let validPath = false;
+    let relativeArtifactPath;
+    do {
+      const rawArtifactPath = await abortIfCancelled(
+        clack.text({
+          message:
+            'Provide a path to the folder that contains your build artifacts:',
+          placeholder: `path${path.sep}to${path.sep}your${path.sep}build${path.sep}output`,
```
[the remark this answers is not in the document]
lforst: Not everybody might know what dist is, I'm going with `./out`.

src/sourcemaps/tools/sentry-cli.ts
lforst: ah right

Lms24 (approved):

Issue #305, opened 2023-06-26
[sourcemaps] Provide copy/paste instructions for specific bundlers (Prio: Webpack > Vite > tsc > esbuild > Rollup)


Lms24:
```[tasklist]
### High prio
- [ ] https://github.com/getsentry/sentry-wizard/pull/308
- [ ] https://github.com/getsentry/sentry-wizard/pull/317
```

```[tasklist]
### Low Prio
- [ ] https://github.com/getsentry/sentry-wizard/pull/324
- [ ] https://github.com/getsentry/sentry-wizard/pull/327
- [ ] https://github.com/getsentry/sentry-wizard/pull/325
```

Issue #290, opened 2023-06-23
[Epic] Sentry Wizard for Uploading Source Maps


Lms24:
WIP

smeubank:
### Learnings after initial release:
- NextJS users should be directed to correct or set the NextJS webpack config
- react-native devs should be directed
- print to terminal after completion how to test locally
- Remix users should have a unique flow until we can introduce breaking changes
- Svelte kit users already covered by vite?
- `Did you add a step to your CI pipeline that runs the sentry:ci script right after building your application?`
  - `Do you not know what we are talking about?`
  - we can link to a new page in docs with some example
  - And we can instruct to test locally regardless
  - even this can be a last point to update CI

![image](https://github.com/getsentry/sentry-wizard/assets/47563310/c61c76c4-7cca-40ee-86be-b8120493c642)

Point 1 could be a bit more explicit, which scripts are the standard to run. Could we add a final optional step that the wizard runs scripts for them as a final step to test?

Lms24:
Action points from "Learnings after inital release meeting":
 
* Treat NextJS as a 1st class citizen - fast forward to NextJS wizards source map setup
* ReactNative - determine if there's value in adding a sourcemaps to the wizard?
* Detect Remix -> check docs; tell them about the yarn command
* No CI -> add option "that's fine, just do it before deploying"
* Adjust outro/validation message: detect build toll, add generic example (e.g. yarn build)

Next Steps:
* Injection/AST-modification - start w/ Webpack and Vite; lower prio esbuild/rollup 
  * It doesn't have to be perfect. Cover the basic config forms first
  * fall back to copy/paste when it's not possible to inject

lforst:
> ReactNative - determine if there's value in adding a sourcemaps flow to the wizard?

@krystofwoldrich is working on debug IDs for RN.

It would imo make sense to have a redirection like in sveltekit and Next.js because people seem to get lost.

PR #324, opened 2023-06-30
feat(sourcemaps): Add `tsc` flow

lforst:
Adds a separate tsc wizard to the sourcemaps flow that mostly reuses the sentry cli flow but has a separate prompt to instruct users to correctly configure their tsconfig.json.

Ref:
https://github.com/getsentry/sentry-wizard/issues/305
https://github.com/getsentry/sentry-wizard/issues/290

### DIFF
--- a/src/sourcemaps/sourcemaps-wizard.ts
+++ b/src/sourcemaps/sourcemaps-wizard.ts
@@ -16,14 +16,19 @@ import { SourceMapUploadToolConfigurationOptions } from './tools/types';
 import { configureVitePlugin } from './tools/vite';
 import { configureSentryCLI } from './tools/sentry-cli';
 import { configureWebPackPlugin } from './tools/webpack';
+import { configureTscSourcemapGenerationFlow } from './tools/tsc';
 
 interface SourceMapsWizardOptions {
   promoCode?: string;
 }
 
-type SupportedBundlers = 'webpack' | 'vite' | 'rollup' | 'esbuild';
-type SupportedTools = 'sentry-cli';
-type SupportedBundlersTools = SupportedBundlers | SupportedTools;
+type SupportedTools =
+  | 'webpack'
+  | 'vite'
+  | 'rollup'
+  | 'esbuild'
+  | 'tsc'
+  | 'sentry-cli';
 
 export async function runSourcemapsWizard(
   options: SourceMapsWizardOptions,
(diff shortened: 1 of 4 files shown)


src/sourcemaps/sourcemaps-wizard.ts
```diff
@@ -139,6 +144,11 @@ async function askForUsedBundlerTool(): Promise<SupportedBundlersTools> {
           value: 'vite',
           hint: 'Configure source maps upload using Vite',
         },
+        {
+          label: 'tsc',
```
Lms24, line 148: wdyt about

```suggestion
          label: 'Typescript (tsc)',
```

?

src/sourcemaps/sourcemaps-wizard.ts
```diff
@@ -16,14 +16,19 @@ import { SourceMapUploadToolConfigurationOptions } from './tools/types';
...
+import { configureTscSourcemapGenerationFlow } from './tools/tsc';
 
 interface SourceMapsWizardOptions {
   promoCode?: string;
 }
 
-type SupportedBundlers = 'webpack' | 'vite' | 'rollup' | 'esbuild';
-type SupportedTools = 'sentry-cli';
-type SupportedBundlersTools = SupportedBundlers | SupportedTools;
+type SupportedTools =
```
Lms24, line 25: good cleanup, i initially separated these because I was toying with the idea of creating one flow for the 4 bundlers which is why i split it up. But now that we have separate flows (which is probably better anyway) it was just unnecessary 😅

src/sourcemaps/tools/tsc.ts
```diff
@@ -0,0 +1,39 @@
...
+    )}`,
+  );
+
+  // Intentially logging directly to console here so that the code can be copied/pasted directly
+  // eslint-disable-next-line no-console
+  console.log(codeSnippet);
+
+  await abortIfCancelled(
+    select({
+      message: 'Did you update your config like in the snippet above?',
```
Lms24, line 21: l:
```suggestion
      message: 'Did you update your config as shown in the snippet above?',
```

Lms24 (approved):
Nice, LGTM!
(great reuse pattern with injecting the source map generation flow!)

src/sourcemaps/sourcemaps-wizard.ts
lforst: I intentionally chose tsc to avoid ambiguity with people using e.g. typescript + webpack

PR #325, opened 2023-06-30
feat(sourcemaps): Add Rollup plugin flow

Lms24:
This PR adds the source maps setup flow for rollup (with our rollup plugin). It's basically identical to the Vite/Webpack flows, just with adjusted code snippets and package installs.

(Tested w/ a node app that's bundled with rollup - works as expected)

ref #305

### DIFF
--- a/src/sourcemaps/tools/rollup.ts
+++ b/src/sourcemaps/tools/rollup.ts
@@ -16,27 +16,27 @@ import {
 
 const getCodeSnippet = (options: SourceMapUploadToolConfigurationOptions) =>
   chalk.gray(`
-  ${chalk.greenBright(
-    'import { sentryRollupPlugin } from "@sentry/rollup-plugin";',
-  )}
+${chalk.greenBright(
+  'import { sentryRollupPlugin } from "@sentry/rollup-plugin";',
+)}
 
-  export default {
-    output: {
-      ${chalk.greenBright(
-        'sourcemap: true, // Source map generation must be turned on',
-      )}
-    },
-    plugins: [
-      // Put the Sentry rollup plugin after all other plugins
-      ${chalk.greenBright(`sentryRollupPlugin({
-        authToken: process.env.SENTRY_AUTH_TOKEN,
-        org: "${options.orgSlug}",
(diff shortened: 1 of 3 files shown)


src/sourcemaps/tools/rollup.ts
```diff
@@ -0,0 +1,69 @@
...
+      // Put the Sentry rollup plugin after all other plugins
+      ${chalk.greenBright(`sentryRollupPlugin({
+        authToken: process.env.SENTRY_AUTH_TOKEN,
+        org: "${options.orgSlug}",
+        project: "${options.projectSlug}",${
+        options.selfHosted ? `\n      url: "${options.url}",` : ''
+      }
+      }),`)}
+    ],
+  };
```
lforst, line 39: We should probably remove the base level indentation of this snippet.

src/sourcemaps/tools/rollup.ts
```diff
@@ -0,0 +1,69 @@
...
+    await installPackage({
+      packageName: '@sentry/rollup-plugin',
+      alreadyInstalled: hasPackageInstalled(
+        '@sentry/rollup-plugin',
+        await getPackageDotJson(),
+      ),
+    });
+
+    clack.log.step(
+      `Add the following code to your ${chalk.bold('rollup.config.js')} file:`,
```
lforst, line 53: Hm, maybe just the following just so we don't make any assumptions about the users setup? But I honestly don't know.

```suggestion
      `Add the following code to your rollup config:`,
```

lforst (approved):

lforst (approved):

src/sourcemaps/tools/rollup.ts
Lms24: Hmm yeah let's go with this. I guess since rollup can be used directly in JS, we can't even be sure they have a file... (I guess the same thing holds for Vite but I haven't seen this yet outside framework-internal builds).

src/sourcemaps/tools/rollup.ts
Lms24: Good catch, copy 🍝

PR #329, opened 2023-07-04
Use `with-environment.sh` in Xcode Build Phases

liamjones:
I'd like to suggest making use of React Native's `with-environment.sh` script in the "Upload Debug Symbols to Sentry" build phase added by Sentry. If this is acceptable I can raise separate PRs for updating the manual installation docs.

My reasoning;

I was following the [docs for adding source context to the iOS native side of an RN project](https://docs.sentry.io/platforms/react-native/upload-debug/#uploading-source-context-with-xcode). It states:

> To enable [Source Context](https://docs.sentry.io/platforms/apple/guides/ios/data-management/debug-files/source-context/) set up a [Run Script build phase for uploading debug symbols to Sentry](https://docs.sentry.io/platforms/react-native/manual-setup/manual-setup/#upload-debug-symbols-to-sentry) and set `SENTRY_INCLUDE_NATIVE_SOURCES=true` before the build.

`ios/.xcode.env` seems like a good place to set this, [the comment at the top of it](https://github.com/facebook/react-native/blob/v0.72.1/packages/react-native/template/ios/_xcode.env) states:

```shell
# This `.xcode.env` file is versioned and is used to source the environment
# used when running script phases inside Xcode.
```

[It's already sourced in the "Bundle React Native code and images" build phase](https://github.com/facebook/react-native/blob/v0.72.1/packages/react-native/template/ios/HelloWorld.xcodeproj/project.pbxproj#L267), what do you think about adding it to the "Upload Debug Symbols to Sentry" build phase too?


lib/Steps/Integrations/ReactNative.ts
```diff
@@ -420,6 +420,7 @@ The snippet will create a button that, when tapped, sends a test event to Sentry
       {
         shellPath: '/bin/sh',
         shellScript: `
+source "../node_modules/react-native/scripts/xcode/with-environment.sh"
```
krystofwoldrich, line 423: Because we support RN from 0.65.0 and newer we need to check if the file exists first.

```suggestion
WITH_ENVIRONMENT="../node_modules/react-native/scripts/xcode/with-environment.sh"
if [ -f "$WITH_ENVIRONMENT" ]; then
    source "$WITH_ENVIRONMENT"
fi 
```

krystofwoldrich (changes requested):
I like the idea, it's very convenient to set env in the file than in the Xcode project scripts.

We just have to make sure not to break projects with older RN versions.

liamjones:
I've got a couple of extra edits to make to this PR before merging. ShellCheck highlighted that source isn't supported in `/bin/sh`

liamjones:
I think I'm done with the changes on this one.

Two follow on questions;

1. Shall I go raise a separate PR against the Sentry docs to mention setting the environment variable in `.xcode.env` now? 2. In a similar vein - is a PR needed to add a new "migrations" section to the docs to tell users to update their "Upload Debug Symbols to Sentry" script to match what sentry-wizard will produce after this PR is merged?

krystofwoldrich:
Hi, thank you for the other PR, but let's add all the `with-environment.sh` changes in one PR.

lib/Steps/Integrations/ReactNative.ts
```diff
@@ -507,7 +511,7 @@ export SENTRY_PROPERTIES=sentry.properties
           // remove sentry properties export
           .replace(/^export SENTRY_PROPERTIES=sentry.properties\r?\n/m, '')
           .replace(
-            /^\/bin\/sh ..\/node_modules\/@sentry\/react-native\/scripts\/collect-modules.sh\r?\n/m,
+            /^\/bin\/sh -c.*?..\/node_modules\/@sentry\/react-native\/scripts\/collect-modules.sh"?\r?\n/m,
```
krystofwoldrich, line 514: I've missed this at first, but also the `-c` should go, so we match the prev version of the command.
```suggestion
            /^\/bin\/sh .*?..\/node_modules\/@sentry\/react-native\/scripts\/collect-modules.sh"?\r?\n/m,
```

krystofwoldrich (approved):
Looks good. This is a very nice and convenient feature for React Native Apple devs. It makes using envs super simple and consistent. 

After the merge, we also need to update our docs.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - fix(sveltekit): Bump `magicast` to handle vite configs declared as variables (#380)
 - ref(sveltekit): Add vite plugin insertion fallback mechanism (#379)
 - ref(sveltekit): Insert project config into vite config instead of `sentry.properties` (#378)
+- Use `with-environment.sh` in React Native Xcode Build Phases (#329)
```
krystofwoldrich, line 10:
```suggestion
- feat(reactnative): Use `with-environment.sh` in Xcode Build Phases (#329)
```

PR #333, opened 2023-07-05
ref(build): Use rollup to build the wizard

Lms24:
This PR changes the build process of the wizard by replacing transpilation with `tsc` to transpiling and bundling with Rollup and a bunch of Rollup plugins:

* Babel for TS->JS transpilation
* CommonJS conversion 
* Type declaration generation with @rollup/plugin-typescript and tsc

The package entry points remain the same so technically this is not a breaking change. (Even if, cutting a major wouldn't be a big deal for the wizard anyway...).

Draft because this still needs some cleanup and more thorough testing, preferrably also for the mobile/RN wizard flows. I'll come back to this after finishing some more important tasks.

Other things that are left to do or follow up after merging this:
- upgrade some dependencies that are now ESM-only (`open` 👀 )
- adjust `tsconfig` and remove `@ts-ignores` around `clack` imports, now that it can be converted to cjs


Lms24:
Closing for now. I think this PR is a PoC and we can revisit it if it comes up again that we need a better bundling process. For now, I don't have the time to work on this further.

PR #334, opened 2023-07-05
feat(apple): Add support for iOS

brustolin:
Added the flow to add Sentry to an iOS project.


brustolin:
This PR is big already, I will open for review and add support for cocoapod and fastlane in another one.

Lms24 (commented):
Looks good to me from a JS dev perspective :) Can't say much about the code snippets/SDK configuration specifics but I think the general wizard flow looks good. I had some comments/suggestions.

src/apple/apple-wizard.ts
```diff
@@ -0,0 +1,110 @@
...
+    promoCode: options.promoCode,
+  });
+
+  let hasCLI = bash.hasSentryCLI();
+  if (!hasCLI) {
+    if (!(await askToInstallSentryCLI())) {
+      clack.log.warn('Without Sentry-cli, you wont be able to upload debug symbols to Sentry. You can install it later by following the instructions at https://docs.sentry.io/cli/');
+    } else {
+      await bash.installSentryCLI();
+      hasCLI = true;
```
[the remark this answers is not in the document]
brustolin: Good point. At first I was thinking to not add build script without CLI, but I changed my mind. I can totally remove this variable.

package-lock.json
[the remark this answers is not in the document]
brustolin: Yeap. Im using npm because its easier to debug in the vscode. I will remove this file.

src/apple/apple-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const projectDir = process.cwd();
+  const xcodeProjFile = findFilesWithExtension(projectDir, ".xcodeproj")[0];
+
+  if (!xcodeProjFile) {
+    clack.log.error('No xcode project found. Please run this command from the root of your project.');
+    return;
+  }
+
+  const pbxproj = path.join(projectDir, xcodeProjFile, "project.pbxproj");
```
Lms24, line 54: Just to sanity-check (for someone who has no idea about xcode/ios apps): Is this a valid path? 

```
wizard/wizard-apple/test.xcodeproj/project.pbxproj
```

Background: I briefly tried the wizard locally but couldn't get beyond the pbxproj step:

image

src/apple/apple-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+      await bash.installSentryCLI();
+    }
+  }
+
+  const projectDir = process.cwd();
+  const xcodeProjFile = findFilesWithExtension(projectDir, ".xcodeproj")[0];
+
+  if (!xcodeProjFile) {
+    clack.log.error('No xcode project found. Please run this command from the root of your project.');
+    return;
```
Lms24, line 51: One thing I'm also realizing here is that we shouldn't return from the wizard flow without first aborting. Calling `await abort` from `clack-utils.ts` will make sure, we exit with an error exit code and (if telemetry data is collected) we flush out pending Sentry events before exiting. 

I also did this incorrectly in the SvelteKit wizard. Will open an issue to fix there.

Lms24 (commented):
Thanks for adding my suggestions! 

One more thing:

image

For folks who start the wizard without the -i arg, I think the grouping here isn't ideal. Wdyt about moving iOs to the top or below RN? I think we should sort the selection here roughly by platform (mobile, browser/JS + utilities like source maps).

At some point, depending on how many other wizards we add, we might want to somehow make these groups more apparent but for now I'd say moving iOS up should be good.

src/apple/apple-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+): Promise<void> {
+  printWelcome({
+    wizardName: 'Sentry Apple Wizard',
+    promoCode: options.promoCode,
+  });
+
+
+  if (!bash.hasSentryCLI()) {
+    if (!(await askToInstallSentryCLI())) {
+      clack.log.warn("Without Sentry-cli, you won't be able to upload debug symbols to Sentry. You can install it later by following the instructions at https://docs.sentry.io/cli/");
```
philipphofmann, line 41:
```suggestion
      clack.log.warn("Without sentry-cli, you won't be able to upload debug symbols to Sentry. You can install it later by following the instructions at https://docs.sentry.io/cli/");
```

src/apple/apple-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+    } else {
+      await bash.installSentryCLI();
+    }
+  }
+
+  const projectDir = process.cwd();
+  const xcodeProjFile = findFilesWithExtension(projectDir, ".xcodeproj")[0];
+
+  if (!xcodeProjFile) {
+    clack.log.error('No xcode project found. Please run this command from the root of your project.');
```
philipphofmann, line 51:
```suggestion
    clack.log.error('No Xcode project found. Please run this command from the root of your project.');
```

src/apple/apple-wizard.ts
```diff
@@ -0,0 +1,91 @@
...
+    return;
+  }
+
+  const { project, apiKey } = await getSentryProjectAndApiKey(options.promoCode, options.url);
+
+  xcManager.updateXcodeProject(pbxproj, project, apiKey, true, true);
+
+  const codeAdded = codeTools.addCodeSnippetToProject(projectDir, project.keys[0].dsn.public);
+  if (!codeAdded) {
+    clack.log.warn('Sentry dependency was added to your project, but could not add Sentry code snippet to it. Please add it manually by following this: https://docs.sentry.io/platforms/apple/guides/ios/#configure');
```
philipphofmann, line 69: Maybe a bit cleaner:
```suggestion
    clack.log.warn('Added the Sentry dependency to your project but could not add the Sentry code snippet. Please add the code snipped manually by following the docs: https://docs.sentry.io/platforms/apple/guides/ios/#configure');
```

src/apple/apple-wizard.ts
brustolin: Yes. The ".xcodeproj" is actually a folder that keeps project configuration files.

I think we should not show the "Successfully..." message in this case.

> I briefly tried the wizard locally but couldn't get beyond the pbxproj step:

You just tried to run it, or you really have an xcode project to test it?

src/apple/code-tools.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const swiftAppLaunchRegex = /(func\s+application\s*\(_\sapplication:[^,]+,\s*didFinishLaunchingWithOptions[^,]+:[^)]+\)\s+->\s+Bool\s+{)|(init\s*\([^)]*\)\s*{)/gim;
+const objcAppLaunchRegex = /-\s*\(BOOL\)\s*application:\s*\(UIApplication\s*\*\)\s*application\s+didFinishLaunchingWithOptions:\s*\(NSDictionary\s*\*\)\s*launchOptions\s*{/gim;
+
+
+function isAppDelegateFile(filePath: string): boolean {
+    const appLaunchRegex = filePath.toLowerCase().endsWith(".swift") ? swiftAppLaunchRegex : objcAppLaunchRegex;
+
+    const fileContent = fs.readFileSync(filePath, 'utf8');
+    return (fileContent.includes("UIApplicationDelegate") && appLaunchRegex.test(fileContent)) || /struct[^:]+:\s*App\s*{/.test(fileContent);
```
philipphofmann, line 13: `m`:  Will this also work for Objective-C? Cause the `UIApplicationDelegate` is in the header file.

src/apple/apple-wizard.ts
```diff
@@ -0,0 +1,94 @@
...
+  const projectDir = process.cwd();
+  const xcodeProjFile = findFilesWithExtension(projectDir, ".xcodeproj")[0];
+
+  if (!xcodeProjFile) {
+    clack.log.error('No xcode project found. Please run this command from the root of your project.');
+    await abort();
+    return;
+  }
+
+  const pbxproj = path.join(projectDir, xcodeProjFile, "project.pbxproj");
```
philipphofmann, line 56: `m`: Does this also work for `.xcworkspace`?

src/apple/code-tools.ts
```diff
@@ -0,0 +1,77 @@
...
+        return true;
+    }
+
+    let match = appLaunchRegex.exec(fileContent);
+    if (!match) {
+        const swiftUIMatch = /struct[^:]+:\s*App\s*{/.exec(fileContent)
+        if (!swiftUIMatch) {
+            return false;
+        }
+        //Is SwiftUI with no init
```
philipphofmann, line 64: `m`: I wonder how that can work for SwiftUI when we exit above if there is no AppDelegate?

brustolin:
> Wdyt about moving iOs to the top or below RN? 

Sound good. Can we try to find out which project exists to reorder the list? 
Lets say there is a `.xcodeproj` directory,  we know this may be an iOS or RN project, I guess we have other rules to indicate a RN project.

src/apple/templates.ts
```diff
@@ -0,0 +1,47 @@
...
+export function getSwiftSnippet(dsn: string): string {
+    return `           SentrySDK.start { options in
+            options.dsn = "${dsn}"
+            options.enableTracing = true
+            #if DEBUG
+            options.debug = true
+            options.environment = "Development"
+            #else
+            options.environment = "Release"
+            #endif
```
philipphofmann, line 22: `m`: We don't have that in our docs anywhere. Why would you like to change it?
```suggestion
            options.debug = true // Enabled debug when first installing is always helpful
```

src/apple/templates.ts
```diff
@@ -0,0 +1,47 @@
...
+    // eslint-disable-next-line no-useless-escape
+    return `# This script is responsable to upload debug symbols and source context for Sentry.\\nif which sentry-cli >/dev/null; then\\nexport SENTRY_ORG=${orgSlug}\\nexport SENTRY_PROJECT=${projectSlug}\\nexport SENTRY_AUTH_TOKEN=${apiKey}\\nERROR=$(sentry-cli debug-files upload ${uploadSource ? "--include-sources " : ""}\"$DWARF_DSYM_FOLDER_PATH\" 2>&1 >/dev/null)\\nif [ ! $? -eq 0 ]; then\\necho \"warning: sentry-cli - $ERROR\"\\nfi\\nelse\\necho \"warning: sentry-cli not installed, download from https://github.com/getsentry/sentry-cli/releases\"\\nfi\\n`;
+}
+
+export const scriptInputPath = "\"${DWARF_DSYM_FOLDER_PATH}/${DWARF_DSYM_FILE_NAME}/Contents/Resources/DWARF/${TARGET_NAME}\"";
+
+export function getSwiftSnippet(dsn: string): string {
+    return `           SentrySDK.start { options in
+            options.dsn = "${dsn}"
+            options.enableTracing = true
```
philipphofmann, line 16: `m`: I think `enableTracing` should be below setting `debug = true`.

src/apple/templates.ts
```diff
@@ -0,0 +1,47 @@
...
+            #if DEBUG
+            options.debug = true
+            options.environment = "Development"
+            #else
+            options.environment = "Release"
+            #endif
+            options.attachScreenshot = true
+            options.attachViewHierarchy = true
+            options.enableTimeToFullDisplay = true
```
philipphofmann, line 25: `m`: I'm unsure if we should enable that by default. If you don't call the manual API the transactions will look a bit weird.

src/apple/templates.ts
```diff
@@ -0,0 +1,47 @@
...
+            options.dsn = "${dsn}"
+            options.enableTracing = true
+            #if DEBUG
+            options.debug = true
+            options.environment = "Development"
+            #else
+            options.environment = "Release"
+            #endif
+            options.attachScreenshot = true
+            options.attachViewHierarchy = true
```
philipphofmann, line 24: `h`: I don't think we should enable that by default. Especially the screenshots could contain sensitive data, even in a development build.

philipphofmann (commented):
Thanks for doing this, @brustolin 💪 . I added some comments.

src/apple/apple-wizard.ts
brustolin: That depends. `.xcworkspace` just groups projects in a workspace, it alone is irrelevant to the wizard. 
But if we have two or more `.xcodeproj` in the same directory, which may be common for someone using `.xcworkspace`, than we're changing the first one the wizard finds, we should ask which project should be updated.

The easy answer is that `.xcworkspace` does not matter for the wizard, but we should check whether there is more than one project in the same directory.

src/apple/apple-wizard.ts
Lms24: > Yes. The ".xcodeproj" is actually a folder that keeps project configuration files.

 Okay, all good then! I guess I was just confused by the `xcodeProjectFile` variable then 😅 

> You just tried to run it

Yeah, but not very methodically and without an actual xcode project. Just wanted to get over a few steps so I created some empty files 😅

Lms24:
> Can we try to find out which project exists to reorder the list?
We sort of do this in the `ChooseIntegration` step for RN and Cordova:

https://github.com/getsentry/sentry-wizard/blob/ebb501ecabc480983e7c9d17d5c036aef4f8f30a/lib/Steps/ChooseIntegration.ts#L58-L90

However, this just influences the default selection in the list, not the order. I'd argue this should be enough. 
Fwiw we didn't add detection for the JS frameworks (Next, SvelteKit) yet but it should be fairly straight-forward.

src/apple/apple-wizard.ts
philipphofmann: Yes, please. Let's do that. Thanks for the clarification, @brustolin.

src/apple/code-tools.ts
brustolin: Yes, we will, and thats true.
Thanks!!

src/apple/code-tools.ts
brustolin: `findAppDidFinishLaunchingWithOptions` also pass for the SwiftUI app file.

src/apple/apple-wizard.ts
smeubank: @brustolin do you know a repo which would be a example app to test this stuff with ourselves?

src/apple/templates.ts
smeubank: what else would be required to give the best product experience? Does it require manual instrumentation which would be way too complicated for the wizard to manage at this stage?

smeubank:
> This PR is big already, I will open for review and add support for cocoapod and fastlane in another one.

@brustolin sounds good 🚀  if you could link to the parent issue for those PRs when you get there just let us know. 

only thing for me, before we make it customer facing is to ensure we have some telemetry with no PII, so we can track that it is working well. We identified a lot of simple edge cases with -sorucemaps wizard this way and are able to ship a lot of little improvements

src/apple/templates.ts
romtsn: if I may, we could just add a comment to the snippet explaining that it can have sensitive data and turn it off in case you don't want it. Or we could add another step "Enable experimental features (could send PII to Sentry)"? and only after accepting it add the snippet

src/apple/templates.ts
brustolin: I copied the options from our sample, and kept was I believe is a good start, but having environment with a fixed value didn't seem right.
I will remove it and lets hope the users know what to do with it.

src/apple/templates.ts
brustolin: I think we're good with the automatic instrumentations

src/apple/templates.ts
brustolin: I like the idea to show this options in the wizard and explain whats happening, but in another PR.

src/apple/apple-wizard.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+//Prompt for Sentry project and API key
+async function getSentryProjectAndApiKey(promoCode?: string, url?: string): Promise<{ project: SentryProjectData, apiKey: { token: string } }> {
+  const { url: sentryUrl } = await askForSelfHosted(url);
+
+  const { projects, apiKeys } = await askForWizardLogin({
+    promoCode: promoCode,
+    url: sentryUrl,
+    platform: 'javascript-nextjs',
```
romtsn, line 91: nit: this should be `apple` I guess?

brustolin: Definitely!

src/apple/apple-wizard.ts
brustolin: I do 😁 https://github.com/brustolin/XcodeProjects

src/apple/templates.ts
philipphofmann: I think adding a comment won't be sufficient. With the wizard, it can happen that some people don't check the code that much. They just hit compile and run after executing the wizard. We could add them //commented out or an extra step would also be OK.

src/apple/templates.ts
```diff
@@ -0,0 +1,39 @@
...
+    return `        SentrySDK.start { options in
+            options.dsn = "${dsn}"
+            options.debug = true // Enabled debug when first installing is always helpful
+            options.enableTracing = true 
+
+            //Uncomment the following lines to add more data to your events
+            //options.attachScreenshot = true //This will add a screenshot to the error events
+            //options.attachViewHierarchy = true //This will add the view hierarchy to the error events
+        }
+        //Remove the next line after confirming that your Sentry integration is working.
```
philipphofmann, line 23: `l`: If I'm not mistaken, code comments usually have a space after the `//` in Swift. Same applies for Objective-C below.
```suggestion
            // Uncomment the following lines to add more data to your events
            // options.attachScreenshot = true // This adds a screenshot to the error events
            // options.attachViewHierarchy = true // This adds the view hierarchy to the error events
        }
        // Remove the next line after confirming that your Sentry integration is working.
```

src/apple/apple-wizard.ts
```diff
@@ -0,0 +1,111 @@
...
+      await bash.installSentryCLI();
+      Sentry.setTag('CLI-Installed', true);
+    }
+  }
+
+  const projectDir = process.cwd();
+  const xcodeProjFiles = findFilesWithExtension(projectDir, ".xcodeproj");
+
+  if (!xcodeProjFiles || xcodeProjFiles.length === 0) {
+    clack.log.error('No xcode project found. Please run this command from the root of your project.');
```
philipphofmann, line 55: `l`: Shouldn't it be Xcode?
```suggestion
    clack.log.error('No Xcode project found. Please run this command from the root of your project.');
```

src/apple/apple-wizard.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  let xcodeProjFile;
+
+  if (xcodeProjFiles.length === 1) {
+    xcodeProjFile = xcodeProjFiles[0];
+    Sentry.setTag('multiple-projects', false);
+  } else {
+    Sentry.setTag('multiple-projects', true);
+    xcodeProjFile = (await traceStep("Choose Xcode project", () => askForItemSelection(xcodeProjFiles, "Which project do you want to add Sentry?"))).value;
```
philipphofmann, line 67: Which project do you want to add Sentry to?
`l`:
```suggestion
    xcodeProjFile = (await traceStep("Choose Xcode project", () => askForItemSelection(xcodeProjFiles, "Which project do you want to add Sentry to?"))).value;
```

philipphofmann (approved):
LGTM from the Cocoa SDK perspective.

src/apple/templates.ts
brustolin: We can added it. But is not a comment rule, it will compile.

PR #335, opened 2023-07-05
feat(sourcmaps): Add create-react-app option

lforst:
Adds a basic flow for CRA based on the sentry-cli flow.

### DIFF
--- a/src/sourcemaps/sourcemaps-wizard.ts
+++ b/src/sourcemaps/sourcemaps-wizard.ts
@@ -20,6 +20,7 @@ import { configureTscSourcemapGenerationFlow } from './tools/tsc';
 import { configureRollupPlugin } from './tools/rollup';
 import { configureEsbuildPlugin } from './tools/esbuild';
 import { WizardOptions } from '../utils/types';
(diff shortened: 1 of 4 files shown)


src/sourcemaps/sourcemaps-wizard.ts
```diff
@@ -190,8 +197,11 @@ async function startToolSetupFlow(
     case 'tsc':
       await configureSentryCLI(options, configureTscSourcemapGenerationFlow);
       break;
-    default:
+    case 'create-react-app':
       await configureSentryCLI(options);
       break;
+    default:
+      await configureSentryCLI(options, configureCRASourcemapGenerationFlow);
```
matejminar, line 204: I might be missing some context - shouldn't this second param go into the `create-react-app` case?

lforst: Oh you're totally right!

Lms24 (approved):
Nice!

matejminar (approved):
Looks good to me. I also like the change in wording regarding the access to your config 👏

PR #336, opened 2023-07-05
feat(sourcemaps): Check if correct SDK version is installed

Lms24:
This PR adds a mechanism to check if users installed an SDK version `>=7.47.0` to be compatible with debug-id based source maps. During this check, we distinguish between four cases:

1. Users didn't install any SDK yet
   -> We tell them to install an SDK and then continue with the wizard
2. Users installed an SDK in the range >=7.47.0
   -> All good, no need to do anything!
3. Users installed an SDK in the range >=7.0.0 <= 7.46.0
   -> We ask if they want to auto-update to the latest version
4. Users installed an SDK in the range <7.x
   -> We tell users to manually upgrade (migrate between majors) and then continue with the wizard

Telemetry-wise, we record the initial SDK version range (mostly based on the four cases above) and the action taken by the user (installed/updated manually/automatically or installed/updated later).

The version check is based on the version entered in `package.json` and `semver`. 

I opted to add this check early in the wizard flow because it basically is a requirement to successfully get a source-mapped event in Sentry. If folks think it should be at the end we can also just move it further down. 

ref #290


src/sourcemaps/utils/sdk-version.ts
```diff
@@ -0,0 +1,213 @@
...
+  );
+
+  Sentry.setTag(
+    'resolved_sdk_status',
+    didUpdate ? 'updated_manually' : 'update_later',
+  );
+}
+
+async function handleAutoUpdateSdk(packageName: string) {
+  Sentry.setTag('initial_sdk_version', '>=7.0.0 <= 7.47.0');
```
lforst, line 141:
```suggestion
  Sentry.setTag('initial-sdk-version', '>=7.0.0 <= 7.47.0');
```

giga l: All of our tags so far are kebab case

src/sourcemaps/utils/sdk-version.ts
```diff
@@ -0,0 +1,213 @@
...
+export async function ensureMinimumSdkVersionIsInstalled(): Promise<void> {
+  const packageJson = await getPackageDotJson();
+
+  const installedSdkPackage = SENTRY_SDK_PACKAGES.find((sdkPackage) =>
+    hasPackageInstalled(sdkPackage, packageJson),
+  );
+
+  // Case 1:
+  if (!installedSdkPackage) {
+    return await handleNoSdkInstalled();
```
lforst, line 60: Should we set a Tag here?

src/sourcemaps/utils/sdk-version.ts
```diff
@@ -0,0 +1,213 @@
...
+  getPackageVersion,
+  hasPackageInstalled,
+  installPackage,
+} from '../../utils/clack-utils';
+
+import * as Sentry from '@sentry/node';
+
+const MINIMUM_DEBUG_ID_SDK_VERSION = '7.47.0';
+
+const SENTRY_SDK_PACKAGES = [
```
lforst, line 17: Do we want to add a short comment here saying that this array is sorted by order of detection?

src/sourcemaps/utils/sdk-version.ts
```diff
@@ -0,0 +1,213 @@
...
+    hasPackageInstalled(sdkPackage, packageJson),
+  );
+
+  // Case 1:
+  if (!installedSdkPackage) {
+    return await handleNoSdkInstalled();
+  }
+
+  const installedSdkVersionOrRange =
+    getPackageVersion(installedSdkPackage, packageJson) || '';
```
lforst, line 64: Falling back to empty string is pretty hard to reason about when looking at this code. What is conna happen inside `satisfies` below?

Can we somehow either unify the `hasPackageInstalled` and `getPackageVersion` call so we don't have to do the fallback logic or explicitly bail when the version is undefined?

src/sourcemaps/utils/sdk-version.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  const shouldUpdate = await abortIfCancelled(
+    clack.select({
+      message:
+        'Do you want to automatically update your SDK to the latest version?',
+      options: [
+        {
+          label: 'Yes!',
+          value: true,
+          hint: chalk.greenBright('Recommended'),
```
lforst, line 151: I think the `bright` modifier will override any `dim`/`bright` modifiers that clack sets. It kinda breaks the visual coherency of the wizard. Do we still wanna go with `bright`?

lforst (approved):
Noice

src/sourcemaps/utils/sdk-version.ts
Lms24: We do (first line in `handleNoSdkInstalled`) by setting the `initial-sdk-version` to `none`.

Or do you mean an additional, explicit tag?

src/sourcemaps/utils/sdk-version.ts
Lms24: we use greenBright in a few places. But not in hints, you're right. Yeah let's go with green

src/sourcemaps/utils/sdk-version.ts
Lms24: I refactored this a little because I'll need to extract it anyway for the SvelteKit/NextJS check later. Typing still isn't perfect but right now it's at least reduced to the typing problem coming from `semver` rather than ours 😅

src/sourcemaps/utils/sdk-version.ts
lforst: Ah then this is fine!

PR #340, opened 2023-07-07
feat(sourcemaps): Add spans for indivudiual wizard steps

Lms24:
This PR adds a small telemetry utilitly function `traceStep` which can be used to create a span for a specific wizard step within a flow. For now we use the utility function in the sourcemaps wizard flow:

image

ref #290

### DIFF
--- a/src/sourcemaps/sourcemaps-wizard.ts
+++ b/src/sourcemaps/sourcemaps-wizard.ts
@@ -22,6 +22,7 @@ import { configureEsbuildPlugin } from './tools/esbuild';
 import { WizardOptions } from '../utils/types';
 import { configureCRASourcemapGenerationFlow } from './tools/create-react-app';
 import { ensureMinimumSdkVersionIsInstalled } from './utils/sdk-version';
+import { traceStep } from '../telemetry';
 
 type SupportedTools =
   | 'webpack'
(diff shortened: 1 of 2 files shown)


src/sourcemaps/sourcemaps-wizard.ts
```diff
@@ -42,95 +43,43 @@ export async function runSourcemapsWizard(
...
-   `)}
-   ${chalk.dim(
-     `If you encounter any issues, please refer to the Troubleshooting Guide:
-   https://docs.sentry.io/platforms/javascript/sourcemaps/troubleshooting_js
-
-   If the guide doesn't help or you encounter a bug, please let us know:
-   https://github.com/getsentry/sentry-javascript/issues`,
-   )}
-`);
+  traceStep('outro', printOutro);
```
lforst, line 82: Any reason we're tracing this?

lforst (approved):

src/sourcemaps/sourcemaps-wizard.ts
Lms24: Just for the sake of completeness. Can remove as it doesn't add much value at this point

lforst: I am fine with whatever :D

Lms24: Leaving it in then

PR #341, opened 2023-07-07
feat(sourcemaps): Detect SvelteKit and NextJS project and redirect to dedicated wizards

Lms24:
We observed that several users run the source maps wizard in NextJs (or potentially SvelteKit) projects which isn't ideal. While we should get to the bottom of _why_ they end up there (our Next/Kit docs don't mention the source maps wizard), we should let them know that a better option exists for them. Therefore, with this PR, the wizard

* attempts to detect if the wizard is run in a NextJs|SvelteKit project
  * precondition: Sentry SDK AND the framework's core package are installed
* lets users know about the dedicated wizard
* prints links to the framework's source maps docs and troubleshooting guide
* asks users if they want to run the wizard now
* provides an option to continue anyway
* provides an option to exit at this point

Example:
image

ref #290

Up for debate: 
Chicken and egg problem - do people run this wizard before they install an SDK? Then the condition for this check isn't satisfied since only the framework package is installed (but not the SDK package). However, with #336 we detect if we don't find an SDK and tell folks to install an SDK first.


src/utils/package.ts
```diff
@@ -0,0 +1,50 @@
+export type PackageDotJson = {
```
lforst, line 1: l: Do we wanna rename this file to something like `utils/package-json.ts`? `package.ts` sounds a bit ambiguous to me.

src/utils/package.ts
```diff
@@ -0,0 +1,50 @@
...
+export function findPackageFromList(
+  packageNamesList: string[],
+  packageJson: PackageDotJson,
+): NpmPackage | undefined {
+  const installedSdkPackages = packageNamesList
+    .map((packageName) => ({
+      name: packageName,
+      version: getPackageVersion(packageName, packageJson),
+    }))
+    .filter((sdkPackage): sdkPackage is NpmPackage => !!sdkPackage.version);
```
lforst, line 27: l: We could think about using `Array.find` here instead. Feel free to skip and ignore.

src/sourcemaps/utils/other-wizards.ts
```diff
@@ -0,0 +1,125 @@
...
+    frameworkName: 'SvelteKit',
+    frameworkSlug: 'sveltekit',
+    frameworkPackage: '@sveltejs/kit',
+    sourcemapsDocsLink:
+      'https://docs.sentry.io/platforms/javascript/guides/sveltekit/manual-setup/#configure-source-maps-upload',
+    wizard: runSvelteKitWizard,
+  },
+};
+
+export async function redirectToMoreSuitableWizard(): Promise<
```
lforst, line 43: From what I can see this function does two things:

- Check if a framework with better wizard is installed
- Redirects the flow if one was found

I would do either of two things:
- Rename the function to `checkIfMoreSuitableWizardExistsAndRedirect()`
- Extract the logic for checking whether next/sveltekit is installed out of this function.

The name of this function would lead me to believe that it redirects to another flow no matter what.

src/sourcemaps/utils/other-wizards.ts
```diff
@@ -0,0 +1,125 @@
...
+${sourcemapsDocsLink}
+
+Troubleshooting Source Maps:
+https://docs.sentry.io/platforms/javascript/guides/${frameworkSlug}/sourcemaps/troubleshooting_js/
+`,
+  );
+
+  const nextStep: 'redirect' | 'continue' | 'stop' = await abortIfCancelled(
+    clack.select({
+      message: `Do you want to run the ${frameworkName} wizard now?`,
```
lforst, line 98: no action required: Wondering if we should give the user a heads-up what the other wizard will do. I.e. set up the SDK in its entirety. Maybe they want to just have sourcemaps.

lforst (approved):

src/utils/package.ts
Lms24: Oh god, what did I do here 😂

src/sourcemaps/utils/other-wizards.ts
Lms24: Yes, I'll add a sentence.

Lms24: On second thought, we already have a sentence above that indicates this. I'll leave it as is for the moment.

Lms24:
Some things from this PR might potentially change in the near future. I'm still gonna merge this for the time being so that we reduce user confusion now.

PR #345, opened 2023-07-07
fix(telemetry): Only trace step if there is an active span

Lms24:
This PR fixes a bug where we would start a trace when Sentry was not enabled. This happened when starting the source maps wizard w/o the -i flag.

#skip-changelog

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -1,5 +1,6 @@
 import {
   defaultStackParser,
+  getCurrentHub,
   Hub,
   Integrations,
   makeMain,
@@ -76,8 +77,6 @@ function createSentryInstance(enabled: boolean, integration: string) {
     },
 
     transport: makeNodeTransport,
-
-    debug: true,
   });
 
   const hub = new Hub(client);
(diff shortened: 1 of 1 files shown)


mydea:
hmm, this is an interesting wrinkle to the `trace` function 🤔 maybe there should be flag or so like `onlyIfNotRoot` or whatever 😓 cc @AbhiPrasad

mydea (approved):

Lms24:
I believe I should add a little more information here so that everyone has the necessary context:

The bug I'm fixing here is that users get the following console warning when starting the wizard without the `-i` arg, everytime my `traceStep` helper calls `trace`:

```bash
Tracing extension 'startTransaction' has not been added. Call 'addTracingExtensions' before calling 'init':
Sentry.addTracingExtensions();
Sentry.init({...});
```

This warning is emitted because when manually selecting the source maps wizard flow (i.e. no `-i` arg), we don't yet initialize the SDK. So we call `trace` without Sentry being initialized and hence the tracing extension methods are missing. 

Given that `trace` is meant for internal use only, it's okay that this currently errors but I think we can fix this nevertheless in the SDK. I'll open a PR for it!

Lms24:
Closing, SDK bump should be enough lol

PR #348, opened 2023-07-10
ref(sourcemaps): Always collect source maps wizard telemetry

Lms24:
Previously, we only collected telemetry when the source maps wizard was directly started with the `-i sourcemaps` arg. This is probably the majority of cases anyway. However, for the sake of consistency, we should always collect telemetry. Therefore, this PR slightly refactors the location of the `withTelemetry` call so that we're now always using it. 

#skip-changelog
ref #290

### DIFF
--- a/src/sourcemaps/sourcemaps-wizard.ts
+++ b/src/sourcemaps/sourcemaps-wizard.ts
@@ -20,10 +20,10 @@ import { configureWebPackPlugin } from './tools/webpack';
 import { configureTscSourcemapGenerationFlow } from './tools/tsc';
 import { configureRollupPlugin } from './tools/rollup';
 import { configureEsbuildPlugin } from './tools/esbuild';
-import { WizardOptions } from '../utils/types';
+import { WizardOptions, WizardWithTelemetryOptions } from '../utils/types';
 import { configureCRASourcemapGenerationFlow } from './tools/create-react-app';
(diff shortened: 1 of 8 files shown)


src/sourcemaps/sourcemaps-wizard.ts
```diff
@@ -38,9 +38,23 @@ type SupportedTools =
   | 'create-react-app'
   | 'angular';
 
-export async function runSourcemapsWizard(
-  options: WizardOptions,
+export async function runSourcemapsWizardWithTelemetry(
```
mydea, line 41: l: I wonder if we need to rename this (and other places) like this...? Isn't it enough to have `runSourcemapsWizard({ telemetryEnabled: true })` to make this clear?

Lms24: My original intention was to make it clear that this wizard would run with telemetry but you're right, it's way cleaner if the withTelemetry call just happens inside the respective wizard's `run*Wizard` function. I refactored this in https://github.com/getsentry/sentry-wizard/pull/348#:~:text=simplify%20runSourcemapsWizard-,4b55d05,-Merge%20state

For the moment this also has an impact on the Apple wizard, namely that we're passing in the unused `telemetryEnabled` flag. cc @brustolin - are you find with me adjusting the apple wizard slightly? I can do it in a follow-up PR.

mydea (approved):
🚀

PR #349, opened 2023-07-10
feat(sourcemaps): Add CLI-based flow for Angular

Lms24:
This PR adds a CLI-based source maps upload configuration flow for Angular. 
The reason is that for Angular devs, it might not be apparent that the "best" option for them previously was "None of the above". Some clicked "Webpack" although Angular doesn't offer access to the webpack config by default. Hence this option, similar to CRA, simply redirects to the CLI flow and adds a custom source maps generation step.

ref getsentry/team-sdks#16

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - feat(apple): Add support for iOS (#334)
+- feat(sourcemaps): Add CLI-based flow for Angular (#349)
 - feat(sourcemaps): Detect SvelteKit and NextJS projects and redirect to dedicated wizards (#341)
 - ref(sourcemaps): Improve Outro message (#344) 
 


src/sourcemaps/sourcemaps-wizard.ts
```diff
@@ -105,8 +105,18 @@ export async function runSourcemapsWizard(
 async function askForUsedBundlerTool(): Promise<SupportedTools> {
   const selectedTool: SupportedTools | symbol = await abortIfCancelled(
     clack.select({
-      message: 'Which bundler or build tool are you using?',
+      message: 'Which framework, bundler or build tool are you using?',
       options: [
+        {
```
mydea, line 110: l: Not 100% sure how this works, but I figure these are shown in order - if so, should we order them by our perceived importance? Probably webpack first, ...?

Lms24: My train of thought was that as a user I'd probably expect to see the framework choices on top and only afterwards, if my framework is not in the list, the bundlers (which are ordered by our perceived importance).  From that perspective, it also wouldn't make sense to group the actual CLI and the "CLI-based" (CRA, Angular) flows together because as a user  I'm not aware that these are very similar. So therefore, I decided to order it as follows

1. Framework (by importance)
2. Bundlers (by importance)
3. CLI

While writing the order above though, I realized that I should swap CRA and Angular as CRA is probably more widely used.

mydea: that makes sense to me as well. AFAIK we have plans to auto-detect this down the line anyhow, right? (pre-select it, I mean) - which should "fix" this problem hopefully anyhow. so all good 👍

PR #350, opened 2023-07-11
Feat(apple): Add cocoapods support

brustolin:
Added cocoapods support for the iOS wizard.

If the user has a Podfile, the wizard will add reference to Sentry with CocoaPod instead of SPM.


src/apple/cocoapod.ts
```diff
@@ -0,0 +1,56 @@
...
+        "Running 'pod install'. This may take a few minutes...",
+    );
+
+    try {
+        await bash.execute("pod install --silent");
+        loginSpinner.stop("Sentry pod added to the project");
+    } catch (e) {
+        clack.log.error("'pod install' failed. You will need to run it manually.");
+        loginSpinner.stop();
+        Sentry.captureException(e);
```
Lms24, line 52: h: This error can contain PII. For instance file names in the stack frame. Is there a specific reason you want to capture the entire error?
 
The `withTelemetry` helper already catches ocurring errors and sends a very generic "Error during wizard execution" error to Sentry. I know this isn't ideal but we wanted to be on the pii-safe side for this. If you still want to capture this somehow different than the generic exception, I'd suggest one of the following things:

* Capture another string exception like in `withTelemetry` but with a different message
* Instead of capturing an exception, set a tag and let `withTelemetry` capture the generic exception + set transaction status, etc.

brustolin: I want to know the reasons that `pod install` is falling, maybe the string insertion is doing something wrong. but file paths are actually not relevant. 
I will change it to use `withTemeletry` helper and see how it goes.

src/apple/cocoapod.ts
```diff
@@ -4,6 +4,7 @@ import * as bash from '../utils/bash';
 import * as Sentry from '@sentry/node';
 // @ts-ignore - clack is ESM and TS complains about that. It works though
 import * as clack from '@clack/prompts';
+import { traceStep } from '../telemetry';
```
brustolin, line 7:
```suggestion
```

Lms24 (approved):
LGTM from the wizard side!

src/apple/cocoapod.ts
```diff
@@ -0,0 +1,56 @@
...
+    const podContent = fs.readFileSync(podfile, 'utf8');
+
+    if (/^\s*pod\s+(['"]Sentry['"]|['"]SentrySwiftUI['"])\s*$/im.test(podContent)) {
+        // Already have Sentry pod
+        return true;
+    }
+
+    let podMatch = /^( *)pod\s+['"](\w+)['"] *$/im.exec(podContent);
+    if (!podMatch) {
+        //No Podfile is empty, will try to add Sentry pod after "use_frameworks!"
```
philipphofmann, line 24: `l`:
```suggestion
        // No Podfile is empty, will try to add Sentry pod after "use_frameworks!"
```

src/apple/cocoapod.ts
```diff
@@ -0,0 +1,56 @@
...
+
+
+    const loginSpinner = clack.spinner();
+
+    loginSpinner.start(
+        "Running 'pod install'. This may take a few minutes...",
+    );
+
+    try {
+        await bash.execute("pod install --silent");
```
philipphofmann, line 47: `m`: Why `--silent`? Don't we show that to the user? If there is a problem, the user doesn't get the error message.

src/apple/cocoapod.ts
```diff
@@ -0,0 +1,56 @@
...
+        loginSpinner.stop("Sentry pod added to the project");
```
philipphofmann, line 48: `l`:
```suggestion
        loginSpinner.stop("Sentry pod added to the project.");
```

philipphofmann (approved):
One question about the `--silent` option. Apart from that LGTM.

PR #351, opened 2023-07-11
Suggestion - use with-environment.sh with collect-modules.sh

liamjones:
In a similar vein to https://github.com/getsentry/sentry-wizard/pull/329, I think it'd be helpful to call the `with-environment.sh` script with `collect-modules.sh` as well as the bundle + upload debug symbols processes so that overrides can be provided via `.xcode.env` / `.xcode.env.local`.

As an example, we're using this to set `MODULES_PATHS` overrides in our monorepo.


liamjones:
Sure

PR #352, opened 2023-07-11
Tweak RN Android build.gradle "apply from"

liamjones:
Using the project reference means it's less likely to need a manual change for monorepo or non-standard directory layouts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- feat(reactnative): Improve `build.gradle` patch so that it's more likely to work without changes in monorepos (#352)
+
 ## 3.9.1
 
 - ref(sourcemaps): Handle no vite config found case (#391)
--- a/lib/Steps/Integrations/ReactNative.ts
+++ b/lib/Steps/Integrations/ReactNative.ts
@@ -371,7 +371,7 @@ The snippet will create a button that, when tapped, sends a test event to Sentry
   private _unpatchBuildGradle(contents: string): Promise<string> {
     return Promise.resolve(
       contents.replace(
-        /^\s*apply from: project(["']:sentry_react-native["']).projectDir.getParent\(\) + ["']\/sentry.gradle["'];?\s*?\r?\n/m,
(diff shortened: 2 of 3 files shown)


liamjones:
Bump and reminder to self; `_unpatchBuildGradle` should also undo the old change format as well as the new one.

liamjones:
> Could you share a minimal example project with the change? So we can easily verify all works as expected.

I'll try and get that done sometime this week. It doesn't need to be inside the sentry repo itself does it? It can just be a standalone example repo?

liamjones:
@krystofwoldrich apologies for the delay, things got very busy!

I've pushed an example repo here: https://github.com/liamjones/sentry-wizard-pr-352-examples

It contains:
* a git submodule for sentry-wizard pointing at my fork's branch
* a regular, standalone RN app in `StandaloneApp`
* Two RN apps in a monorepo under the directory `monorepo`. I've only configured the Android side to work, I've ignored the iOS side since it's irrelevant to this PR.
* All apps have `sentry-wizard` as a `file:` dependency so you can `yarn sentry-wizard` to use the submodule version

If you check it out you'll want to:
* init the submodule
* `yarn install` and `yarn build` inside `sentry-wizard`
* I've done the sentry integrations in their own commits so you can checkout, for example, https://github.com/liamjones/sentry-wizard-pr-352-examples/commit/7ec0f86e994b58def57ec3a39dd08036bd506cf3 if you want to try the wizard install yourself.

Wizard install and uninstall seem to work fine bar there's still one niggle with monorepos but that's, I think, outside the scope of this PR. When using the wizard in the monorepo it doesn't add the `@sentry/react-native` dependency itself, but it _does_ tell you you need to do it manually.

krystofwoldrich:
@liamjones Thank you for the extensive example. 

The auto-install of the dependency is based on the lock files, therefore it doesn't work for monorepos at the moment. And I agree that's outside of the scope of this PR.

@liamjones What do you think about using `apply from: new File(["node", "--print", "require.resolve('@sentry/react-native')"].execute(null, rootDir).text.trim(), "sentry.gradle");`

We already use it in the `sentry.gradle`, the output it the same but it relies on the node package resolution and not the generated `sentry_react-native` project name.

PR #353, opened 2023-07-11
ref(apple): Move `withTelemetry` into `runAppleWizard`

Lms24:
This aligns the apple wizard with the refactor of the source maps wizard in #348 where we moved the `withTelemetry` call into the wizard. 
The benefit is that we have the call in one place instead of two and that we can potentially check the `telemetryEnabled` flag whenever we need to within the wizard, as it's now part of the wizard options. 

(This is not strictly required but IMO it makes sense to keep things aligned). 

#skip-changelog


brustolin (approved):
LGTM

brustolin:
Do we really need to way for https://github.com/getsentry/sentry-wizard/pull/348? 
This PR will have some conflicts with my other PRs, the sooner to merge this the better.

Lms24:
> Do we really need to way for https://github.com/getsentry/sentry-wizard/pull/348?

Just merged it and merging this one now :)

PR #354, opened 2023-07-11
feat(sourcemaps): Pre-select auto-detected build tool option

Lms24:
This PR adds a simple auto-detection of installed build tool packages to preselect a tool selection option in the source maps wizard. Detection is performed by checking the users' `package.json` for specific packages. Either the packages of the build tool/bundler or the framework/library (Angular/CRA). If nothing is found, we fall back to the catch-all "None of the above" option which starts the Sentry CLI flow. 

ref #290


mydea (approved):
very cool 🚀

PR #355, opened 2023-07-12
ref(sveltekit): Create `.ts` hooks files if typescript is detected

Lms24:
Just noticed while using the Kit wizard for a repro app that we create js files by default. We already have a simple typescript detection check in the NextJS wizard, so this PR
* extracts the typescript check from NextJS
* adds it to the SvelteKit wizard to create `hooks.(server|client).ts` files if TS is detected


package.json
```diff
@@ -1,118 +1,14 @@
...
-  "keywords": [
-    "sentry",
-    "wizard",
-    "sdk",
-    "cli",
-    "project",
-    "setup",
-    "install",
-    "configure"
+  "version": 3,
```
mydea, line 2: I think this file should not be changed...? 😅

Lms24: lol nope 😅

mydea (approved):

PR #360, opened 2023-07-13
fix: Ensure wizard exits after setup is completed

Lms24:
When the wizard was started with the `-i` flag, it would sometimes just keep on awaiting a promise after the `run[integration]Wizard` function returned. Seems like we broke the promise chain in #348 or sometime before that. This caused the application to appear hanging and users needing to cancel the wizard manually. 

Affected wizards.
* NextJs
* Sveltekit
* Sourcemaps
* Apple (not tested but I'm fairly certain)

This patch makes sure that the wizard exits by removing our duplicated invokation logic for the `clack`-based wizard. For the moment, we always go through the Shim classes. I'd like to refactor this eventually and get rid of the shims but for now we should fix this to unblock users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - feat(sourcemaps): Add path for users who don't use CI (#359)
+- fix: Ensure wizard exits after setup is completed (#360)
 - fix(sveltekit): Call correct API endpoint in Sentry example code (#358)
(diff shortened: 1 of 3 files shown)


mydea (approved):
can't say I 100% know what's going on, but if we duplicate this in the `run()` method already anyhow then all good :D

AbhiPrasad (approved):

PR #369, opened 2023-07-19
ref(sourcemaps): Redirect to ReactNative wizard if RN project is detected

Lms24:
This PR adds a redirection mechanism to the RN wizard if we detect an installed RN SDK and the `react-native` package in a user's `package.json`. Since for RN, we need to redirect to an old wizard flow, we can't just call `run*Wizard()` but we basically need to go through the entire wizard startup sequence again to correctly parse the CLI args. I therefore opted to just open a child process with the wizard instead. Not the prettiest solution but it seems to work well enough.  

ref https://github.com/getsentry/team-sdks/issues/16

### DIFF
--- a/src/sourcemaps/utils/other-wizards.ts
+++ b/src/sourcemaps/utils/other-wizards.ts
@@ -16,12 +16,14 @@ import {
 import * as Sentry from '@sentry/node';
 import { WizardOptions } from '../../utils/types';
 
+import * as childProcess from 'child_process';
+
 type WizardFunction = (options: WizardOptions) => Promise<void>;
 
 type FrameworkInfo = {
   frameworkName: string;
-  frameworkSlug: string;
   frameworkPackage: string;
+  troubleshootingDocsLink: string;
   sourcemapsDocsLink: string;
   wizard: WizardFunction;
 };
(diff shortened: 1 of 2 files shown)


src/sourcemaps/utils/other-wizards.ts
```diff
@@ -146,3 +159,24 @@ https://docs.sentry.io/platforms/javascript/guides/${frameworkSlug}/sourcemaps/t
...
+      cwd: process.cwd(),
+      stdio: 'inherit',
+    });
+  } catch {
+    clack.log.error(
+      `Could not redirect to the react-native wizard.\nPlease re-run the wizard manually with the \`${chalk.cyan(
+        '-i reactNative',
+      )}\` argument`,
+    );
+    return Promise.reject();
```
lforst, line 178: In the case the react wizard fails, will this message be printed? If so, I don't think that is fully fitting.

Lms24: Hmm it's a good point. Either the invocation fails (in which case the message would make sense IMO) or the RN wizard fails later (in which the message doesn't make sense) but we can't distinguish between the cases here. Let's remove the log then. In case the wizard fails, users should get another error anyway.

mydea (approved):

lforst (approved):

PR #370, opened 2023-07-19
feat: Autodetect more wizards

Lms24:
Improve the wizard selection menu when starting the wizard without the `-i/--integration` arg:

Add additional checks as we have already for RN and Cordova for to pre-select other wizards if we detect a corresponding framework package:
* Electron
* NextJS
* SvelteKit

This leaves us with source maps and Apple that don't have auto-detection. Both is fine but we can probably check for apple and default to source maps(?) in the future. 

cc @brustolin if you have a free minute, how can we check for apple here?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- feat: Autodetect more wizards (#370)
 - feat(apple): iOS wizard has support for cocoapods (#350)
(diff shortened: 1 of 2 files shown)


mydea (approved):

brustolin:
> @brustolin if you have a free minute, how can we check for apple here?

Basically looking for a directory name ending with `.xcodeproj`

PR #372, opened 2023-07-21
feat(sourcemaps): Add dedicated NextJS sourcemaps flow

Lms24:
This PR adds a dedicated NextJS source maps flow to the source maps wizard. This flow will:
* Ask users if they want to run the NextJS wizard. 
  * If yes, the wizard will run but users will return to the last steps (getting sentry into CI) to finish the source maps setup
* If not, the wizard will display the configuration for `next.config.js`
  * Also, it'll add a few links to nextjs source map troubleshooting and configuration

The challenge with this kind of redirection was to skip a second round of authentication/project selection when users choose to redirect to the nextjs wizard. Therefore, this PR adds some refactoring around project selection and already passing in the previously selected project data to another wizard.

ref https://github.com/getsentry/team-sdks/issues/16

### DIFF
--- a/src/sourcemaps/sourcemaps-wizard.ts
+++ b/src/sourcemaps/sourcemaps-wizard.ts
@@ -5,9 +5,6 @@ import * as Sentry from '@sentry/node';
 
 import {
   abortIfCancelled,
-  askForProjectSelection,
-  askForSelfHosted,
-  askForWizardLogin,
   confirmContinueEvenThoughNoGitRepo,
   detectPackageManager,
   SENTRY_DOT_ENV_FILE,
@@ -23,7 +20,7 @@ import { configureWebPackPlugin } from './tools/webpack';
 import { configureTscSourcemapGenerationFlow } from './tools/tsc';
 import { configureRollupPlugin } from './tools/rollup';
 import { configureEsbuildPlugin } from './tools/esbuild';
-import { SentryProjectData, WizardOptions } from '../utils/types';
+import { WizardOptions } from '../utils/types';
 import { configureCRASourcemapGenerationFlow } from './tools/create-react-app';
 import { ensureMinimumSdkVersionIsInstalled } from './utils/sdk-version';
 import { traceStep, withTelemetry } from '../telemetry';
--- a/src/sourcemaps/sourcemaps-wizard.ts
+++ b/src/sourcemaps/sourcemaps-wizard.ts
@@ -13,6 +13,7 @@ import {
   SENTRY_DOT_ENV_FILE,
   printWelcome,
   SENTRY_CLI_RC_FILE,
+  getOrAskForProjectData,
 } from '../utils/clack-utils';
 import { isUnicodeSupported } from '../utils/vendor/is-unicorn-supported';
 import { SourceMapUploadToolConfigurationOptions } from './tools/types';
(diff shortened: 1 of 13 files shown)


src/sourcemaps/sourcemaps-wizard.ts
```diff
@@ -73,37 +75,37 @@ You can turn this off by running the wizard with the '--disable-telemetry' flag.
 
   await traceStep('check-sdk-version', ensureMinimumSdkVersionIsInstalled);
 
-  const { url: sentryUrl, selfHosted } = await traceStep(
-    'ask-self-hosted',
-    () => askForSelfHosted(options.url),
-  );
-
-  const { projects, apiKeys } = await traceStep('login', () =>
-    askForWizardLogin({
```
lforst, line 82: l: We have a bunch of unused imports because of the removals in this file.

src/utils/clack-utils.ts
```diff
@@ -659,3 +650,42 @@ export function isUsingTypeScript() {
...
+export async function getOrAskForProjectData(options: WizardOptions): Promise<{
+  sentryUrl: string;
+  selfHosted: boolean;
+  selectedProject: SentryProjectData;
+  authToken: string;
+}> {
+  if (options.preSelectedProject) {
+    return {
+      selfHosted: options.preSelectedProject.selfHosted,
+      sentryUrl: options.url ?? 'https://sentry.io',
```
lforst, line 663: h: I believe the fallback needs a trailing slash.

src/sveltekit/sveltekit-wizard.ts
```diff
@@ -27,6 +27,8 @@ export async function runSvelteKitWizard(
     promoCode: options.promoCode,
   });
 
+  await createOrMergeSvelteKitFiles('123', { kit: {} });
```
lforst, line 30: I think we should remove this.

src/sveltekit/sdk-setup.ts
```diff
@@ -401,8 +401,12 @@ async function modifyViteConfig(viteConfigPath: string): Promise<void> {
     return;
   }
 
+  console.log(viteConfigContent, viteConfigPath);
+
   const viteModule = parseModule(viteConfigContent);
 
+  console.log(viteModule.$ast);
```
lforst, line 408: We should probably remove these logs.

src/sourcemaps/sourcemaps-wizard.ts
```diff
@@ -196,6 +204,9 @@ async function startToolSetupFlow(
         configureAngularSourcemapGenerationFlow,
       );
       break;
+    case 'nextjs':
+      await configureNextJsSourceMapsUpload(options, wizardOptions);
```
lforst, line 208: don't know where else to put this: I think we need a call to `addSentryCliRc` for the nextjs source maps wizard flow

src/sourcemaps/sourcemaps-wizard.ts
Lms24: oof, that's what happens when you push a PR on Friday afternoon 🙃 Fixed (along with the other leftovers)

src/sourcemaps/sourcemaps-wizard.ts
Lms24: Adressed by:
* printing the correct filename in the terminal when the "I have no CI" path is taken ([c7b87d9](https://github.com/getsentry/sentry-wizard/pull/372/commits/c7b87d9fced008bf97f3557b0f6c411673c8ac14))
* calling `addSentryCliRc` if users chose not to be redirected to the nextjs wizard ([6130b57](https://github.com/getsentry/sentry-wizard/pull/372/commits/6130b57f2c3547c159c550e5a4fe873e89f15635))

lforst (approved):

PR #374, opened 2023-07-25
feat(sourcemaps): Add option to add cli npm script to build command

### LINKED ISSUE #366: Option to modify existing build script
### Problem statement

If using a platform like vercel or netlify, it is not easy, if not possible at all, to add a new command after your build step in their UI. What I found to be easier was modifying the existing build script or create a new build script which starts with the original

The wizard today creates a new script in `package.json` and does not modify the existing specifically because user may not want their script modified and they may not want to upload source maps every time they build, since htey may build more often than they create actual new releases which require new source maps

image

image

### Solution Proposal

Provide an option in the wizard which prompts the user if they would like for us to ammend the existing build script

If the build script exists then wizard would ammend if the user selects yes, if no then proceeds as today

`Do you want to add uploading source maps to an existing build command?`
`Yes`: we add it so that the command becomes `build: whateverWasHereBefore && npm run sentry:sourcemaps`

Lms24:
Our CLI-based flows currently expect users to manually call the `sentry:sourcemaps` NPM script after building and before deploying and to adjust their CI steps to acommodate for this step. This leaves room for churn/error and sometimes it's not even possibly to easily add another step (see case in #366). 

Therefore, this PR introduces an optional step in the wizard to add the `sentry:sourcemaps` script to the users' build command:
- After the `sentry:sourcemaps` script is added to users' `package.json`, the wizard will ask if the automatically want to run the script after their prod build
- The wizard then ask if the detected script is the users' prod build script and fall back to letting them choose from a list of all of their npm scripts.
- Once selected, the wizard appends the script: `"build": "whateverWasHereBefore && [packageManager] run sentry:sourcemaps"`

image

closes #366

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,7 @@
 - fix(login): Avoid repeatedly printing loading message (#368)
 - fix(sveltekit): Abort the wizard when encountering an error (#376)
 - ref(sourcemaps): Redirect to ReactNative wizard if RN project is detected (#369)
-- fix(login): Avoid repeatedly printing loading message (#368)
-  
+
 ## 3.7.1
 
 fix(telemetry): Re-enable telemetry collection (#361)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,12 @@
 - feat(apple): iOS wizard has support for cocoapods (#350)
 - feat(apple): Add Fastlane detector for iOS wizard (#356)
 - feat(sourcemaps): Add dedicated NextJS sourcemaps flow (#372)
+- feat(sourcemaps): Add option to add cli npm script to build command (#374)
 - fix(login): Avoid repeatedly printing loading message (#368)
 - fix(sveltekit): Abort the wizard when encountering an error (#376)
 - ref(sourcemaps): Redirect to ReactNative wizard if RN project is detected (#369)
-
+- fix(login): Avoid repeatedly printing loading message (#368)
+  
 ## 3.7.1
 
 fix(telemetry): Re-enable telemetry collection (#361)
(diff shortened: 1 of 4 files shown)


src/sourcemaps/tools/sentry-cli.ts
```diff
@@ -126,19 +132,125 @@ async function createAndAddNpmScript(
...
+
+  if (buildCommand) {
+    isProdBuildCommand = await abortIfCancelled(
+      clack.confirm({
+        message: `Is ${chalk.cyan(
+          `${pacMan} run build`,
+        )} your production build command?`,
+      }),
+    );
+  }
```
Lms24, line 209: Right now, I only checked for `build` but we could establish an ordered list of other heuristics, such as `build:prod`, `prod`, etc. Not sure if we wanna go down this route though given that the possibilities are endless...

mydea: What about:

1. Check for `build` exact match
2. If no exact match, check for first entry that `script.includes('build')`

This is obv. not perfect but maybe a 90% good enough heuristic? IMHO it's also fine to only auto-select `build` and have no auto-selection otherwise, though...

mydea (approved):
nice!

CHANGELOG.md
```diff
@@ -6,9 +6,11 @@
 - feat(apple): iOS wizard has support for cocoapods (#350)
 - feat(apple): Add Fastlane detector for iOS wizard (#356)
 - feat(sourcemaps): Add dedicated NextJS sourcemaps flow (#372)
+- feat(sourcemaps): Add option to add cli npm script to build command (#374)
 - fix(login): Avoid repeatedly printing loading message (#368)
 - ref(sourcemaps): Redirect to ReactNative wizard if RN project is detected (#369)
-
+- fix(login): Avoid repeatedly printing loading message (#368)
+
```
Lms24, line 13:
```suggestion
```

src/sourcemaps/tools/sentry-cli.ts
Lms24: Yeah, let's do this - good suggestion!

Issue #241, opened 2023-03-26
Wizard doesn't work with Next.js app dir because it tries to place file in pages folder


lforst:
Ref: https://github.com/getsentry/sentry-javascript/issues/6726#issuecomment-1483286056

mayank1513:
PR with fix - https://github.com/getsentry/sentry-wizard/pull/385

PR #385, opened 2023-07-29
feat(nextjs): Add support for Next.js 13 app router

### LINKED ISSUE #307: Next.js 13 app router + @sentry/wizard breaks typing
### Is there an existing issue for this?

- [X] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [X] I have reviewed the documentation https://docs.sentry.io/
- [X] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Sentry Saas (sentry.io)

### Which SDK are you using?

@sentry/nextjs

### SDK Version

7.56.0

### Framework Version

13.4.5

### Link to Sentry event

_No response_

### SDK Setup

_No response_

### Steps to Reproduce

This bug appears in a Next 13 app router project using typescript and the App router exclusively. 

Create a project and set up a dynamic route in the app directory. e.g. app/post/[id]/page.tsx
On the new page use the useParams() hook from next/navigation to get the id 
`const { id } = useParams();`
This currently will be typed as a string. Running the build command should also be successful

Now run the sentry wizard as per the Sentry next.js setup instructions: 
`npx @sentry/wizard@latest -i nextjs`

Now building will fail due to a type error. This is due to the wizard creating a test route in the pages directory.
We now have both a pages and app directory so next/navigation uses the "compat" types. id is no longer of type string
but instead is of type: Record<string, string, string[]> | null

This caveat should probably be documented in the Next.js setup instructions

### Expected Result

An app router project to build successfully after running the sentry wizard

### Actual Result

image

mayank1513:
- fixes #307 and #241
- checks for appRouter as well
- defaults to creating a pages folder if neither page nor app router is detected


mayank1513:
Also fixes #241

mayank1513:
> Hey, thanks for this PR. We are very likely gonna merge it but I will hold off for a bit because I am planning on adding API routes support very soon and then we can drop the special casing.

Thanks.

Can I help in adding support for route handlers?

lforst:
> > Hey, thanks for this PR. We are very likely gonna merge it but I will hold off for a bit because I am planning on adding API routes support very soon and then we can drop the special casing.
> 
> Thanks.
> 
> Can I help in adding support for route handlers?

Yes of course. PRs are always welcome!

andycansdale:
Just to add to this, I spent a number of hours trying to debug why my middleware suddenly stopped working when installing Sentry using the wizard. I eventually tracked it down to the addition of a pages folder on my root. As I am using a src/ directory for my app folder, middleware, etc. this caused the middleware to not be run.

A possible solution might be add a step that asks the user if they are using a src folder, or perhaps check the location of the existing app directory. Not sure if it's relevant, but my app did not have a pages directory before it was added by Sentry since we using the app router for everything.

mayank1513:
> Just to add to this, I spent a number of hours trying to debug why my middleware suddenly stopped working when installing Sentry using the wizard. I eventually tracked it down to the addition of a pages folder on my root. As I am using a src/ directory for my app folder, middleware, etc. this caused the middleware to not be run.
> 
> A possible solution might be add a step that asks the user if they are using a src folder, or perhaps check the location of the existing app directory. Not sure if it's relevant, but my app did not have a pages directory before it was added by Sentry since we are using the app router for everything.

This is taken care of in this PR.

mayank1513:
Hi, I think we have waited enough for updates to sentry core functionality -> out of box support for app directory route handlers. I think we shouldn't block this PR for that. This would be against agile principles.

lforst:
@mayank1513 We will still go with the plan and merge this when route handlers are working.

Last week most of Sentry was in a hackathon and we didn't operate as usual so things should speed up again now.

Issue #364, opened 2023-07-17
[sourcemaps] Remix Wizard Support


smeubank:
Parent issue: https://github.com/getsentry/sentry-javascript/issues/8556

- [ ] source maps wizard setting up for a remix application
- [ ] SDK wizard including source maps set up
  - stretch: include prompts in wizard if user wants the SDK init to handle v2 features already like the remix router
  - bigger-stretch: could we auto-detect v2 feature usage and do it automatically

Wizard for source maps alone would need to do what is described [in docs](https://docs.sentry.io/platforms/javascript/guides/remix/sourcemaps/) for the user:

1. user login then org and project creation
2. A new script in `package.json` for example which calls `remix build --sourcemaps` [remix docs](https://remix.run/docs/en/1.18.1/other-api/dev#remix-build---sourcemap) & `yarn sentry-upload-sourcemaps`
3. creates `.sentryclirc` file with all needed variables
4. Can a script also handle the [deletion](https://docs.sentry.io/platforms/javascript/guides/remix/sourcemaps/)? 

sentry-wizard: https://github.com/getsentry/sentry-wizard

**Follow-ups:**
Once wizard flows are in place, they can be placed in docs for getting started and source maps respectively. As well could be used to update the in product project creation wizard

smeubank:
https://github.com/getsentry/sentry/issues/54615

PR #387, opened 2023-07-31
feat(remix): Add wizard for Remix.

onurtemizkan:
Resolves: #364 

Adds a wizard for auto-instrumenting Remix applications / generating and uploading sourcemaps.

- Creates `.sentryclirc`
- Updates `build` script in `package.json` to generate and upload sourcemaps.
- Instruments `root.tsx` / `root.jsx` for Remix versions 1 and 2.
- Instruments `entry.client` and `entry.server` for Remix versions 1 and 2.

Used `recast` for parts that `magicast` does not support yet.


src/remix/remix-wizard.ts
```diff
@@ -0,0 +1,66 @@
...
+  printWelcome({
+    wizardName: 'Sentry Remix Wizard',
+    promoCode: options.promoCode,
+  });
+
+  await confirmContinueEvenThoughNoGitRepo();
+
+  const remixConfig = await loadRemixConfig();
+  const packageJson = await getPackageDotJson();
+  await ensurePackageIsInstalled(packageJson, '@remix-run/node', 'Remix');
```
AbhiPrasad, line 34: is this still used if they are using the generic `@remix-run/serve` lib?

Should we instead check for `@remix-run/dev`?

src/remix/sdk-setup.ts
```diff
@@ -0,0 +1,499 @@
...
+    return true;
+  }
+  return false;
+}
+
+/**
+ * Copied from sveltekit wizard
+ * We want to insert the init call on top of the file but after all import statements
+ */
+function getInitCallInsertionIndex(originalHooksModAST: Program): number {
```
AbhiPrasad, line 57: If this is from the sveltekit wizard, @Lms24 do you think it's appropriate if we extract into a util?

src/remix/sdk-setup.ts
```diff
@@ -0,0 +1,499 @@
...
+}
+
+function insertClientInitCall(
+  dsn: string,
+  originalHooksMod: ProxifiedModule<any>,
+): void {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const initCall = builders.functionCall('Sentry.init', {
+    dsn,
+    tracesSampleRate: 1.0,
```
AbhiPrasad, line 77: lets also add `replaysSessionSampleRate: 0.1` and `replaysOnErrorSampleRate: 1.0,` here

src/remix/remix-wizard.ts
```diff
@@ -0,0 +1,66 @@
...
+  await addSentryCliRc(
+    apiKeys.token,
+    selectedProject.organization.slug,
+    selectedProject.name,
+  );
+
+  await updateBuildScript();
+  await instrumentRootRoute(isV2, isTS);
+  await initializeSentryOnEntryClient(dsn, isTS);
+  await initializeSentryOnEntryServer(dsn, isTS, isV2);
```
AbhiPrasad, line 65: Can we add debug logging for all of these calls?

Should we also handle error scenarios?

src/remix/sdk-setup.ts
```diff
@@ -0,0 +1,499 @@
...
+
+    const configUrl = url.pathToFileURL(configFilePath).href;
+    const remixConfigModule = (await import(configUrl)) as {
+      default: PartialRemixConfig;
+    };
+
+    return remixConfigModule?.default || {};
+  } catch (e: unknown) {
+    clack.log.error(
+      `Couldn't load ${REMIX_CONFIG_FILE}. Please make sure, you're running this wizard with Node 14 or newer`,
```
AbhiPrasad, line 351: The wizard should we Node 14+ - is the another error message we can write?

src/remix/remix-wizard.ts
onurtemizkan: Yes, good point 👍 Updated

src/remix/remix-wizard.ts
onurtemizkan: Added debug logs and fail-case logs that redirect to documentation.

src/remix/sdk-setup.ts
```diff
@@ -0,0 +1,509 @@
...
+  // We need to deep-copy here because reverse mutates in place
+  const copiedBodyNodes = [...originalHooksModAST.body];
+  const lastImportDeclaration = copiedBodyNodes
+    .reverse()
+    .find((node) => node.type === 'ImportDeclaration');
+
+  const initCallInsertionIndex = lastImportDeclaration
+    ? originalHooksModAST.body.indexOf(lastImportDeclaration) + 1
+    : 0;
+  return initCallInsertionIndex;
```
AbhiPrasad, line 67: can we just run this in a for loop? More efficient than cloning, and iterating through twice.

```js
for (let x = originalHooksModAST.body.length - 1; x >= 0; x--) {
  if (originalHooksModAST.body[x].type === 'ImportDeclaration') {
    return x + 1;
  }
}

return 0;
```

src/remix/sdk-setup.ts
```diff
@@ -0,0 +1,509 @@
...
+  });
+
+  const originalHooksModAST = originalHooksMod.$ast as Program;
+
+  const initCallInsertionIndex = getInitCallInsertionIndex(originalHooksModAST);
+
+  originalHooksModAST.body.splice(
+    initCallInsertionIndex,
+    0,
+    // @ts-ignore - string works here because the AST is proxified by magicast
```
AbhiPrasad, line 124: let's make all these `@ts-expect-error`

src/remix/sdk-setup.ts
```diff
@@ -0,0 +1,509 @@
...
+    clack.log.warn(
+      `File ${chalk.cyan(path.basename(fileName))} already contains Sentry code.
+Skipping adding Sentry functionality to ${chalk.cyan(
+        path.basename(fileName),
+      )}.`,
+    );
+
+    return true;
+  }
+  return false;
```
AbhiPrasad, line 50:
```suggestion
  const includesContent = fileContent.includes('@sentry/remix');
  
  if (includesContent) {
    clack.log.warn(
      `File ${chalk.cyan(path.basename(fileName))} already contains Sentry code.
Skipping adding Sentry functionality to ${chalk.cyan(
        path.basename(fileName),
      )}.`,
    );
  }
  
  return includesContent;
```

src/remix/sdk-setup.ts
```diff
@@ -0,0 +1,509 @@
...
+    process.cwd(),
+    'app',
+    clientEntryFilename,
+  );
+
+  const originalEntryClientMod = await loadFile(originalEntryClient);
+
+  if (hasSentryContent(originalEntryClient, originalEntryClientMod.$code)) {
+    // Bail out
+    clack.log.warn(
```
AbhiPrasad, line 428: Since hasSentryContent already logs out, do we need to log out again here?

src/remix/sdk-setup.ts
```diff
@@ -0,0 +1,509 @@
...
+      recast.types.builders.exportNamedDeclaration(implementation),
+    );
+  } else {
+    if (
+      hasSentryContent(
+        generateCode(handleErrorFunction).code,
+        originalEntryServerMod.$code,
+      )
+    ) {
+      // Bail out
```
AbhiPrasad, line 165:
```suggestion
```

AbhiPrasad (commented):
Round 2 - thanks for sticking with this!

src/remix/sdk-setup.ts
```diff
@@ -0,0 +1,509 @@
+import type { ExportNamedDeclaration, Program } from '@babel/types';
```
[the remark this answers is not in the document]
onurtemizkan: Moved under `codemods` folder 👍

AbhiPrasad (approved):
lgtm! lets wait till @Lms24 double checks though

src/remix/remix-wizard.ts
```diff
@@ -0,0 +1,122 @@
...
+  );
+}
+
+async function runRemixWizardWithTelemetry(
+  options: WizardOptions,
+): Promise<void> {
+  printWelcome({
+    wizardName: 'Sentry Remix Wizard',
+    promoCode: options.promoCode,
+  });
```
Lms24, line 44: l: Since we're collecting telemetry now, we should let users know that we do and how they can opt-out. 

Right now, the sourcemaps wizard does this with a custom welcome message. Perhaps we should instead refactor `printWelcome` to print the note when telemetry is enabled. I'll leave this up to you if we do it this way or just also print a custom message here as we can also easily refactor this later.

src/remix/codemods/root-v1.ts
```diff
@@ -0,0 +1,91 @@
...
+
+          rootRouteAst.exports.default = builders.raw(
+            `withSentry(${expressionToWrap})`,
+          );
+        } else {
+          clack.log.warn(
+            chalk.yellow(
+              `Couldn't instrument ${chalk.bold(
+                rootFileName,
+              )} automatically. wrap your default export with: ${chalk.dim(
```
Lms24, line 61:
```suggestion
              )} automatically. Wrap your default export with: ${chalk.dim(
```

src/remix/sdk-setup.ts
```diff
@@ -0,0 +1,266 @@
...
+
+    return {};
+  }
+}
+
+export async function instrumentRootRoute(
+  isV2?: boolean,
+  isTS?: boolean,
+): Promise<void> {
+  const rootFilename = `root.${isTS ? 'tsx' : 'jsx'}`;
```
Lms24, line 140: Do you think it's likely that people mix up ts and js? Just wondering if we should make a lookup regardless of `isTS`, similarly to what we do with `findScriptFile` in the SvelteKit wizard. If this is unlikely or even not possible in remix then let's leave it as is. I don't think it's a blocker either way.

src/remix/codemods/root-v2.ts
```diff
@@ -0,0 +1,84 @@
...
+    recast.visit(rootRouteAst.$ast, {
+      visitExportDefaultDeclaration(path) {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        const implementation = recast.parse(ERROR_BOUNDARY_TEMPLATE_V2).program
+          .body[0];
+
+        path.insertBefore(
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+          recast.types.builders.exportDeclaration(false, implementation),
+        );
```
Lms24, line 73: This is nice! Didn't look much into recast yet, so it's great to know that this works!

src/remix/remix-wizard.ts
```diff
@@ -0,0 +1,122 @@
...
+      debug(e);
+    }
+  });
+
+  await traceStep('Instrument root route', async () => {
+    try {
+      await instrumentRootRoute(isV2, isTS);
+    } catch (e) {
+      clack.log.warn(`Could not instrument root route.
+  Please do it manually using instructions from https://docs.sentry.io/platforms/javascript/guides/remix/`);
```
Lms24, line 98: I think for now this is fine. Once we put the wizard instructions into docs, I think it's likely that we'll refactor the remix docs so that the manual setup will be in its own page (similarly to Next and Svelte). We just have to adjust the links then. I added a task in the follow-up issue (https://github.com/getsentry/sentry/issues/54615)

src/remix/remix-wizard.ts
```diff
@@ -0,0 +1,122 @@
...
+  await traceStep('Initialize Sentry on server entry', async () => {
+    try {
+      await initializeSentryOnEntryServer(dsn, isTS, isV2);
+    } catch (e) {
+      clack.log.warn(`Could not initialize Sentry on server entry.
+  Please do it manually using instructions from https://docs.sentry.io/platforms/javascript/guides/remix/`);
+      debug(e);
+    }
+  });
+}
```
Lms24, line 122: l: Let's add an outro message at the end of the wizard (see sveltekit/next) so that it doesn't just end abruptly.

Lms24 (commented):
Looks great, thanks for figuring all of this out Onur! Really like the usage of recast whenever magicast doesn't work.

I still had a few comments but I think we're almost ready to get this in.

I also have a follow-up request but this should be done in a separate PR: SvelteKit and Next automatically create a sample page that throws some client/server errors. Can we do something like this in the Remix wizard, too? If you think this is doable, I'll create an issue to track it.

src/sourcemaps/tools/remix.ts
```diff
@@ -0,0 +1,64 @@
...
+    clack.log.info(
+      `Build your app with ${chalk.cyan('remix build --sourcemap')},
+      then upload your source maps using ${chalk.cyan(
+        'sentry-upload-sourcemaps',
+      )} cli tool.`,
+    );
+
+    await abortIfCancelled(
+      clack.select({
+        message: 'Did finish configuring your build and prod scripts?',
```
Lms24, line 58:
```suggestion
        message: 'Did you finish configuring your build and prod scripts?',
```

src/sourcemaps/tools/remix.ts
```diff
@@ -0,0 +1,64 @@
...
+      "Welcome back to the Source Maps wizard - we're almost done ;)",
+    );
+  } else {
+    clack.log.info(
+      `Build your app with ${chalk.cyan('remix build --sourcemap')},
+      then upload your source maps using ${chalk.cyan(
+        'sentry-upload-sourcemaps',
+      )} cli tool.`,
+    );
+
```
Lms24, line 55: m: Can we add a small snippet how they can adjust their build script in package.json to automatically call `sentry-upload-sourcemaps` after building?

Lms24 (commented):
Damn, almost missed that you also already created a sourcemaps wizard flow. Thanks a lot!

Lms24 (commented):
I just ran the wizard in a small remix app and found some more things (thanks for sticking with us here!):
 
* Can we log out a `clack.log.success` for each file we touched/inserted code? I think this increases transparency for users so that they are informed what the wizard is doing (especially for those who ran it without being in a version controlled environment). 

* Let's make sure we're adding the file ending in this log:
image

src/remix/sdk-setup.ts
onurtemizkan: It's possible, but I don't think it's very likely for the files we are modifying here.

onurtemizkan:
@Lms24, thanks for the review! Updated the PR.

src/remix/sdk-setup.ts
Lms24: Okay, then let's leave it as-is for now and revisit if it comes up.

src/sourcemaps/tools/remix.ts
```diff
@@ -0,0 +1,83 @@
...
+    clack.log.info(
+      `You can add ${chalk.cyan(
+        'sentry-upload-sourcemaps',
+      )} to your build script in ${chalk.cyan('package.json')} like this:
+${chalk.dim(`
+...
+"scripts": {
+  "build": "remix build --sourcemap && sentry-upload-sourcemaps"
+}
+...`)}
```
Lms24, line 66: Let's please directly use `console.log` here and adjust the colors to make the `build` line stand out. For reference, feel free to take a look at the other `tools` flows of the sourcemap wizard (for example, `tsc`). 
Reason: By directly using console.log, users can more easily copy/paste the code as no clack UI border characters are in the line.

Lms24 (commented):
Thanks for applying our feedback, Onur! Had one last comment then we can merge!

In preparation to merging, would you mind adding a note to `CHANGELOG.md`? In this repo, changelogs are semi-automatically generated. For each user-facing PR. we add an entry to CHANGELOG.md under (##Unreleased) and our release process will add it to the version entry for the version to be released. I think for this change, a small usage example would be nice. See the entry for the [SvelteKit wizard](https://github.com/getsentry/sentry-wizard/blob/08fe9aa7b34f04bf96ab3bd93acafd73df5c9791/CHANGELOG.md#L94-L103).

Lms24 (approved):
Perfect, thanks Onur!

PR #389, opened 2023-08-03
feat(android): Support for Android wizard

romtsn:
Adds basic support for Android wizard. 3 main steps:
1. Find build.gradle files with 'com.android.application' in them and add our Sentry Gradle plugin from https://docs.sentry.io/platforms/android/#configure
2. Find AndroidManifest.xml and patch it with our config from https://docs.sentry.io/platforms/android/#configure
3. Find Main Activity and patch it with our test snippet from https://docs.sentry.io/platforms/android/#verify

https://github.com/getsentry/sentry-wizard/assets/4999776/86cf0f92-fed0-4bde-b793-a957d7ad6dbb

### Next up
* A follow-up PR with enabling source context by default and adding `sentry.properties` for proguard mappings upload
* Add sentry-wizard to our brew tap

@Lms24 and @lforst would appreciate your review here and pardon for my typescript 🙈


.gitignore
```diff
@@ -14,7 +14,6 @@ build
 node_modules
 npm-debug.log
 ios
-android
```
romtsn, line 17: I'm not sure why it has been added here, probably because of RN (@HazAT?). Either we have to gitignore it or rename the folder, so idk what to do

src/android/android-wizard.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    const activityPatched = traceStep('Patch Main Activity', () =>
+      codetools.patchMainActivity(activityFile),
+    );
+    if (!activityPatched) {
+      clack.log.warn(
+        "Could not patch main activity. You'll have to manually verify the setup.\nPlease follow the instructions at https://docs.sentry.io/platforms/android/#verify",
+      );
+    }
+  }
```
Lms24, line 125: l: I'd recommend adding an outro message to the wizard and a small sentence how to verify the SDK setup. Example from the sveltekit wizard:

```
clack.outro(`
${chalk.green('Successfully installed the Sentry SvelteKit SDK!')}

You can validate your setup by starting your dev environment (`npm run dev`) and visiting "/sentry-example".

Check out the SDK documentation for further configuration:
https://docs.sentry.io/platforms/javascript/guides/sveltekit/
  `);
```

src/android/code-tools.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  if (fs.existsSync(javaSrcDir)) {
+    possibleActivityPath = `${path.join(
+      javaSrcDir,
+      ...packageNameParts,
+      ...activityNameParts,
+    )}.java`;
+    if (!fs.existsSync(possibleActivityPath)) {
+      // try kotlin if no java file
+      possibleActivityPath = possibleActivityPath.replace('.java', '.kt');
```
Lms24, line 50: l: Here we want to look up a file that could either end with `.java` or `.kt`, right? There's a utility function that does this for various JS file endings. Maybe we can refactor it to be more general and reuse it here: https://github.com/getsentry/sentry-wizard/blob/1d62aab0b4cedbfe84b1f20126831efae09568f2/src/utils/ast-utils.ts#L5-L14

Feel free to ignore, logaf-l, I just think it might simplify this function a little.

src/android/code-tools.ts
```diff
@@ -0,0 +1,157 @@
...
+export function patchMainActivity(activityFile: string | undefined): boolean {
+  if (!activityFile || !fs.existsSync(activityFile)) {
+    clack.log.warn('No main activity source file found in filesystem.');
+    Sentry.captureException('No main activity source file');
+    return false;
+  }
+
+  const activityContent = fs.readFileSync(activityFile, 'utf8');
+
+  if (/import io\.sentry\.Sentry;?/i.test(activityContent)) {
```
Lms24, line 95: m: Could there be multiple spaces here?

```suggestion
  if (/import\s+io\.sentry\.Sentry;?/i.test(activityContent)) {
```

src/android/code-tools.ts
```diff
@@ -0,0 +1,157 @@
...
+ *   super.onCreate(savedInstanceState)
+ *   // the snippet goes here <--
+ *   doSomething()
+ * }
+ * ```
+ *
+ * @param activityFile
+ * @returns true if successfully patched, false otherwise
+ */
+export function patchMainActivity(activityFile: string | undefined): boolean {
```
Lms24, line 86: No immediate action required but general comment RE code modifications/lookups: Using regex to check if Sentry is already used in the file or to look up/insert things is quite brittle. I think we still do this in some wizards but in others we try to parse and travers the AST. For example, to avoid flagging commented out Sentry imports as actual imports, etc. I don't think this is a necessity to change now (and frankly, I have no idea if AST parsers for Java/Kotlin exist for JS) but it's probably something to keep in mind if folks write in this in the future.

src/utils/clack-utils.ts
```diff
@@ -674,7 +685,7 @@ export async function getOrAskForProjectData(options: WizardOptions): Promise<{
     askForWizardLogin({
       promoCode: options.promoCode,
       url: sentryUrl,
-      platform: 'javascript-nextjs',
+      platform: platform,
```
Lms24, line 739: whoops, this must have been a leftover from when we extracted this function. Good catch! Would you mind adding the platform parameter to the function calls?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- feat(android): Add wizard support for Android (#389)
```
Lms24, line 5: l: When adding a new wizard snippet, we usually show how to invoke the wizard in the changelog (e.g. see entry for version 3.2.0)

.gitignore
Lms24: Hmm no idea either on why it is here but why do we need to remove it? Maybe we just change it to ./android to scope it to just that directory?

Lms24 (approved):
Looks good to me from the JS/Wizard perspective! I didn't test this yet with an actual Android project (I don't have a test project at hand...) but just gave it a semi thorough review. My comments are mostly around UI and JS things and some things I noticed that might become tricky (see regex comments). Feel free to ignore the logaf-l snippets as you see fit.

src/android/android-wizard.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  const appFile = await traceStep('Select App File', () =>
+    gradle.selectAppFile(buildGradleFiles),
+  );
+
+  // ======== STEP 1. Add Sentry Gradle Plugin to build.gradle(.kts) ============
+  clack.log.info(
+    `Adding ${chalk.bold(
+      'Sentry Gradle plugin',
+    )} to your app's build.gradle file.`,
```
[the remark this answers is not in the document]
romtsn: I've changed info/step messages to use `cyan`, but when it was successful, we already use `greenBright` for the entire message, so I just kept `bold` there, otherwise it looked like a rainbow :D 
image

src/android/android-wizard.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  const { selectedProject } = await getOrAskForProjectData(options, 'android');
+
+  // ======== STEP 2. Configure Sentry SDK via AndroidManifest ============
+  clack.log.info(
+    `Configuring Sentry SDK via ${chalk.bold('AndroidManifest.xml')}`,
+  );
+  const buildFile = path.basename(appFile);
+  const appDir = appFile.split(buildFile)[0];
```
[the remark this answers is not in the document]
romtsn: right, that's exactly what I needed 😅

src/android/android-wizard.ts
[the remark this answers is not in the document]
Lms24: Sorry for causing confusing, I agree we shouldn't create rainbows (leave that up to the rain) 😅 

What I meant was that I'd show the filename (in this case `build.gradle`) in cyan, not the plugin. Maybe this doesn't make sense in the Android wizard case - I'll leave this up to you. 

As for using colours we previously tried to adhere to a few simple (but unfortunately unwritten) rules. We should define this properly somewhere, to make creating wizards easier. I'll chop down some notes here - maybe we can add them to a better place somewhere later on:

There are definitely exceptions to these cases but generally:
* print major success messages in green
* use `clack.log.(success|error|warn|info)` with their respective styles
* print filenames in cyan

For showing copy/paste code instructions (e.g. see sourcemaps wizard) we
* show surronding (unchanged code) in `chalk.gray`
* show code that needs changes/added code in `chalk.greenbBright`+
* Directly `console.log` the code snippet so that it's as easy as possible to copy/paste (using `clack.*` prepends the clack UI characters in front of the code snippet).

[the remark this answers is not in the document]
romtsn: oh, sorry, I completely misread about filenames 🤦 I will change the filename to be in cyan, that makes sense!

src/android/code-tools.ts
romtsn: yep, good catch!

src/android/code-tools.ts
romtsn: Makes total sense! Actually, when I started with this, I wanted to modify AST for the gradle scripts, but I couldn't find any parsers for `.gradle`/`.gradle.kts` files and gave up on that idea. But I see now there are a few existing parsers at least for java/kotlin source files, so we could switch to that in the future. For now, however, I'd like to keep it simple and see how the current regex approach works out.

src/utils/clack-utils.ts
romtsn: I guess it was fine, because I basically parameterized this now to reuse on android, but it was only used on nextjs before that (I've added it in https://github.com/getsentry/sentry- wizard/pull/389/commits/8dff5049b55a3a7ba468a32990be37604d73aaf6).  It's also use in sourcemaps-wizard, but there we don't know the platform at that point, so I guess it's fine to keep it unchanged there?

romtsn:
@Lms24 thanks a lot for your thorough review 🙏 I've addressed all the points let me know if you wanna skim once again, otherwise I'll merge it by noon

Lms24 (approved):
Looks great now! Thanks for addressing my comments! :shipit:

Issue #390, opened 2023-08-04
Improve Vite Plugin Injection


Lms24:
Based on our collected information, plugin insertion fails for 3 reasons:

* The actual AST modification/injection went wrong
  * not much we can do here; let's continue to fall back to copy/paste 
* The vite config couldn't be found
  * this we can improve
* We bailed because the config already contained Sentry code
  * this we can also improve

```[tasklist]
### Improvements
- [ ] https://github.com/getsentry/sentry-wizard/pull/391
- [ ] https://github.com/getsentry/sentry-wizard/pull/392
```

PR #391, opened 2023-08-04
ref(sourcemaps): Handle no vite config found case

Lms24:
This PR improves one of the main reasons why our plugin injection approach currently fails. Previously, if we couldn't find a `vite.config.(js|ts|cjs|mjs)` file, we just bailed and fell back to copy paste instructions. This PR adds the following improvements:

* If we can find a vite.config file, inject the plugin (no changes here). Otherwise:
* Ask users if they have a config file. 
  * If not, create a new file with a standard vite config that contains our plugin and enables source map generation.
    * If this fails (i/o problems), we fall back to telling users to add a config and copy/paste code
  * If yes, ask for the path to the file, validate that we can access it and inject the plugin
* Whenever we attempt plugin injection and we fail, we continue to fall back to copy/paste. I thought about also creating a new file in this case but I think merging a new file into an existing config is not much better than copy/pasting code into an existing one. Happy to change this if reviewers have different opinions. 

ref #390

### DIFF
--- a/src/utils/string.ts
+++ b/src/utils/string.ts
@@ -1,0 +1,7 @@
+export function stripAnsii(str: string): string {
+  return str.replace(
+    // eslint-disable-next-line no-control-regex
+    /[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g,
+    '',
+  );
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased:
 
+- ref(sourcemaps): Handle no vite config found case (#391)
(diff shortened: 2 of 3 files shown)


src/utils/string.ts
```diff
@@ -0,0 +1,7 @@
+export function stripAnsii(str: string): string {
```
lforst, line 1: l: instead of stripping we could give the `getCopyPasteCodeSnippet` an additional param `colors` that will determine the output. But whatever you prefer!

lforst (approved):

src/utils/string.ts
Lms24: It's a good point. I had something similar in mind but abandoned it because stripping was faster and retained better readability of the snippet in our code 😅 However, it feels kind of like an after thought to the original function and somewhat hacky. So I'll change it.

PR #396, opened 2023-08-04
fix(apple): Don't remove other swift packages

brustolin:
Fixed a bug where adding Sentry as SPM removed other SPM pre installed.


AbhiPrasad (approved):

AbhiPrasad:
Should we cut a release today for this @brustolin @kahest?

PR #398, opened 2023-08-11
fix(sentry-cli-sourcemaps): Fix writing of build command

lforst:
In the case when the user has a script called `build` we would take its value instead of its key as the name of the build command - which is, as you would have guessed it, not right.


Lms24 (approved):
Oh damn... good catch!

PR #401, opened 2023-08-17
fix(cordova)  iOS fails to build when sentry.cli not defined

lucas-zimerman:
By default, Sentry.CLI is bundled with Sentry Cordova and can be accessed on the following path: `plugins/sentry-cordova/node_modules/@sentry/cli/bin/sentry-cli`

The current code breaks the build if `cli.executable` is not defined on sentry.properties, so this fix sets the default value when not defined to `plugins/sentry-cordova/node_modules/@sentry/cli/bin/sentry-cli`.

EDIT:

code build with and without cli.executable:

image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- fix(cordova): Fallback to the default Sentry CLI path if not defined. (#???)
+- fix(cordova): Fallback to the default Sentry CLI path if not defined. (#401)
 
 ## 3.9.2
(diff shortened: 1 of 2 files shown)


Lms24 (approved):
LGTM! 

(out of curiosity: Do you know why we use the sentry.properties file here as opposed to `.sentryclirc`? No action required, just interested)

lucas-zimerman:
> 

No idea to be honest, probably wasn't migrated to sentryclirc.

PR #406, opened 2023-08-28
feat(craft): Add brew target on craft

### LINKED ISSUE #346: Sentry Wizard available via brew
Developers on some platforms may be more comfortable working with brew, vs npm/npx/node.

The `@sentry/wizard` should be available via brew

The [sentry-cli is already for example](https://github.com/getsentry/homebrew-tools/blob/master/Formula/sentry-cli.rb).

romtsn:
Adds `brew` target to automate publishing of sentry-wizard to our custom brew tap. 

There's a change on craft needs to land before this one, to make it possible accessing checksums for sentry-wizard https://github.com/getsentry/craft/pull/488

Closes #346

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -12,7 +12,7 @@ targets:
         desc "The Sentry Wizard helps you set up your projects with Sentry"
         homepage "https://github.com/getsentry/sentry-wizard"
         url "https://registry.npmjs.org/@sentry/wizard/-/wizard-{{version}}.tgz"
-        sha256 "{{checksums.sentry-wizard-v{{version}}.tgz}}"
+        sha256 "{{checksums.sentry-wizard-vlatest__tgz}}"
         version "{{version}}"
         license "MIT"
 
--- a/.craft.yml
+++ b/.craft.yml
@@ -12,7 +12,7 @@ targets:
         desc "The Sentry Wizard helps you set up your projects with Sentry"
         homepage "https://github.com/getsentry/sentry-wizard"
         url "https://registry.npmjs.org/@sentry/wizard/-/wizard-{{version}}.tgz"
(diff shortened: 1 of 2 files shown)


HazAT (approved):

.craft.yml
```diff
@@ -3,6 +3,29 @@ changelogPolicy: auto
...
+        url "https://registry.npmjs.org/@sentry/wizard/-/wizard-{{version}}.tgz"
+        sha256 "{{checksums.sentry-wizard-vlatest__tgz}}"
+        version "{{version}}"
+        license "MIT"
+
+        depends_on "node"
+
+        def install
+          system "npm", "install", *Language::Node.std_npm_install_args(libexec)
+          bin.install_symlink Dir["#{libexec}/bin/*"]
```
lforst, line 23: I lack a lot of context. Do you mind quickly explaining what this does?

romtsn: Right, sorry, you can take a look at this PR which adds a new formula for sentry-wizard https://github.com/getsentry/homebrew-tools/pull/6 and the supporting Homebrew guide when publishing Node formulas https://docs.brew.sh/Node-for-Formula-Authors

romtsn: Citing the docs above

> Node modules should be installed to libexec. This prevents the Node modules from contaminating the global node_modules, which is important so that npm doesn’t try to manage Homebrew-installed Node modules.

Lms24 (approved):

lforst (approved):
Generally lgtm. We just need to be sure that when people run the wizard they actually run the latest version of it - otherwise they might get stale installation setups.

When we add this to the docs we may want to tell users to `brew update` beforehand.

.craft.yml
lforst: Makes sense. Thanks!

Issue #407, opened 2023-08-29
Wizard doesn't use provided DSN


jenskuhrjorgensen:
### Environment

`sentry-wizard -v`:
version: 3.10.0 | sentry-cli version: 1.75.2

"react-native": "0.72.4"

### Steps to Reproduce

1. `npx react-native@latest init SentryTest`
2. `sentry-wizard -i reactNative -p android`
3. Answer the four prompts (organization slug, project slug, DSN, auth token)
4. `npx react-native run-android --mode release`

### Expected Result

I expect the app to build, install and launch on an emulator or connected Android device. 

### Actual Result

The app crashes because there is no DSN provided. It turns out the wizard put in 'null' instead of the actual DSN that I provided during the wizard:

```
Sentry.init({ 
  dsn: 'null', 
});
```

I would have expected it to have added my DSN that I pasted during the Sentry Wizard because otherwise I don't know why the wizard would ask for it.

Stacktrace from LogCat:
```
2023-08-29 13:39:41.714 23227-23362 AndroidRuntime          pid-23227                            E  FATAL EXCEPTION: mqt_native_modules
                                                                                                    Process: com.sentrytest, PID: 23227
                                                                                                    java.lang.IllegalArgumentException: java.lang.IllegalArgumentException: Invalid DSN: No public key provided.
                                                                                                    	at io.sentry.Dsn.<init>(Dsn.java:89)
                                                                                                    	at io.sentry.Sentry.initConfigurations(Sentry.java:300)
                                                                                                    	at io.sentry.Sentry.init(Sentry.java:210)
                                                                                                    	at io.sentry.Sentry.init(Sentry.java:147)
                                                                                                    	at io.sentry.android.core.SentryAndroid.init(SentryAndroid.java:89)
                                                                                                    	at io.sentry.android.core.SentryAndroid.init(SentryAndroid.java:70)
                                                                                                    	at io.sentry.react.RNSentryModuleImpl.initNativeSdk(RNSentryModuleImpl.java:115)
                                                                                                    	at io.sentry.react.RNSentryModule.initNativeSdk(RNSentryModule.java:28)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
                                                                                                    	at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:188)
[...]

PR #409, opened 2023-08-31
fix: Support org auth tokens in old wizards

Lms24:
While working on #407 I noticed that the old wizards treat our new org tokens as invalid when they fall back to prompting users to enter a token. This PR adds an org token validation function which attempts to decode a token, based on the [org token RFC](https://github.com/getsentry/rfcs/blob/main/text/0091-ci-upload-tokens.md#token-format).

This does not affect the clack-based wizards (kit, next, sourcemaps, ios, android), as we don't fall back to user-entered values there.

### DIFF
--- a/lib/Steps/PromptForParameters.ts
+++ b/lib/Steps/PromptForParameters.ts
@@ -120,7 +120,16 @@ export class PromptForParameters extends BaseStep {
   }
 
   private _validateAuthToken(input: string): boolean | string {
-    if (!input.match(/[0-9a-f]{64}/g)) {
+    const isOrgToken = input.startsWith('sntrys_');
+
+    if (isOrgToken) {
+      if (!isValidOrgToken(input)) {
+        return 'Make sure you copied the correct auth token, it should start with sntrys_ and consist of >140 base64 characters';
+      }
+      return true;
+    }
+
+    if (!input.match(/(sntrys_)?[0-9a-f]{64}/g)) {
       return 'Make sure you copied the correct auth token, it should be 64 hex chars';
     }
     return true;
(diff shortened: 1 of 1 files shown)


lib/Steps/PromptForParameters.ts
```diff
@@ -120,7 +120,16 @@ export class PromptForParameters extends BaseStep {
   }
 
   private _validateAuthToken(input: string): boolean | string {
-    if (!input.match(/[0-9a-f]{64}/g)) {
+    const isOrgToken = input.startsWith('sntrys_');
+
+    if (isOrgToken) {
+      if (!isValidOrgToken(input)) {
+        return 'Make sure you copied the correct auth token, it should start with sntrys_ and consist of >140 base64 characters';
```
Lms24, line 127: Admittedly, this is a pretty bad hint message. @mydea can we say anything else about the structure of the token that would help users verify here?

Lms24: Ahh probably that the string needs to contain two `_` characters, no?

lforst, line 127: I would maybe just phrase it like this

```suggestion
        return 'Make sure you correctly copied your auth token.';
```

The additional info is not gonna help people I believe unless there is an exact number of characters we can give them.

lib/Steps/PromptForParameters.ts
```diff
@@ -149,3 +158,39 @@ export class PromptForParameters extends BaseStep {
...
+  const tokenParts = input.split('_');
+  if (tokenParts.length < 3) {
+    return false;
+  }
+
+  try {
+    const payload = tokenParts[1];
+    const decodedPayload = Buffer.from(payload, 'base64').toString();
+    const jsonPayload = JSON.parse(decodedPayload) as MaybeOrgAuthToken;
+    if (!jsonPayload.iat || !jsonPayload.url || !jsonPayload.org) {
```
lforst, line 188: will `iat` and `url` always be there? Could they be null?

lforst (approved):

lib/Steps/PromptForParameters.ts
Lms24: According to https://github.com/getsentry/rfcs/blob/main/text/0091-ci-upload-tokens.md#token-facts they should always be there

lib/Steps/PromptForParameters.ts
mydea: yes, it must have two `_`, that should work as a check!

mydea: I think mentioning that it has to start with `sntrys_` makes sense, as it is immediately obvious then when they copied something different.

lib/Steps/PromptForParameters.ts
mydea: yes, they should, spec wise. we seem to have an issue where `url` is sometimes empty when running this locally, but this is something that we have to fix backend-side (and should not affect sentry.io)

Issue #357, opened 2023-07-12
Generate Org Auth Token for Sentry Wizard


Lms24:
Right now, in the wizard we're still generating the old, user-scoped auth tokens. We recently introduced [Org-scoped auth tokens](https://github.com/getsentry/team-sdks/issues/6) which should be preferred over user-based tokens whenever possible. We therefore want to create them in the wizard when selecting a project. Unfortunately, due to the nature of the current project selection flow, this brings along some challenges:

### Problem
* The `SetupWizardView` in the Sentry backend currently obtains one list of projects across _all_ active orgs of a user. 
  * While doing so, it also obtains or generates a new token for the logged-in user 
  * Since we don't know _which_ org the user is going to select a project from in the wizard later, for which org should we generate a token?

Related problems:

- Since the introduction of Hybrid Cloud, the org and project lookup times out for larger orgs/project lists because we have to fan out into potentially multiple region silos to obtain the full list of all of the user's projects.
- Users can pre-select `--org` and `--project` (introduced in https://github.com/getsentry/sentry-wizard/pull/671) to not have to go through the entire list of projects. However, the current endpoint/view still loads all of the users' projects, even if they filtered it down to 1 in the wizard.
- UX-wise, scrolling through a long lists of projects with arrow keys in the Wizard CLI is very cumbersome for users with many projects

### Solution

We move the project selection step from the CLI to the browser UI. Instead of only logging in in the browser, users will also select their project afterwards. 

More specifically, we modify the `SetupWizard` React component that's loaded by the Django template to tell users to return to the wizard. If there's more than one org and project assigned to the user, the user needs to select their project inside the browser. After the selection, we make another request to the `SetupWizardView` which creates a new org token and updates the cache. During this entire time, the CLI (wizard) will wait until the project is selected and the backend created the corresponding org auth token. Once this is done, the CLI receives back the selected project data and token and it directly continues with the setup.

#### Advantages
* It solves the chicken/egg problem of org selection and org auth token creation.
* For users with many orgs, this will significantly improve the UX of selecting a project. Right now, scrolling through the list of projects in the CLI is less than ideal.

#### Disadvantages
* Users with >1 org or project need to do one step more in the browser
* We need to maintain the old flow (i.e. add this change as a new endpoint/view/react component) to maintain backwards-compatibility with older wizard versions.

#### Alternatives Considered

* Just generate a token for each of the user's orgs 
[...]

PR #414, opened 2023-09-01
feat(android): Add org, proj and authToken for proguard mappings/source context upload

romtsn:
_#skip-changelog_

* Refactored `addSentryCliRc` a bit to make it work for both `.sentryclirc` and `sentry.properties` (that's what we recommend for android/java users)
* Used that function^ to add auth.token locally and gitignore it
* Added a new gradle script config which enables source context upload and sets org and project slugs

Skipping changelog because it's gonna be part of #389 which is unreleased yet


src/utils/clack-utils.ts
```diff
@@ -30,6 +31,42 @@ interface WizardProjectData {
...
+  name: string;
+
+  likelyAlreadyHasAuthToken(contents: string): boolean;
+  tokenContent(authToken: string): string;
+
+  likelyAlreadyHasOrgAndProject(contents: string): boolean;
+  orgAndProjContent(org: string, project: string): string;
+}
+
+export class SourceMapsCliSetupConfig implements CliSetupConfig {
```
Lms24, line 45: m-h: I like the idea of the interface but I'd prefer to create objects instead of classes for the concrete implementations. We want to keep a functional approach in the `clack`-based (new) wizards and avoid using classes. 

Suggestion:
* Keep the `CliSetupConfig` interface
* Create objects for the concrete implementations that satisfy the interface
  * Either directly or by writing a `make[Concrete]CliSetupConfig` helper
 
Wdyt?

src/utils/clack-utils.ts
```diff
@@ -30,6 +31,42 @@ interface WizardProjectData {
...
 
+export interface CliSetupConfig {
+  filename: string;
+  name: string;
+
+  likelyAlreadyHasAuthToken(contents: string): boolean;
+  tokenContent(authToken: string): string;
+
+  likelyAlreadyHasOrgAndProject(contents: string): boolean;
+  orgAndProjContent(org: string, project: string): string;
```
Lms24, line 42: l: Given that we want to move towards org-scoped auth tokens, we should probably split this up to org and project. Reason: With org-auth tokens our wizards don't need to set the org slug anymore. Sentry CLI should already support this. 

Feel free to skip this though because afaik no wizard actually leverages this at the moment. The Sentry backend returns org auth tokens for users who only belong to one org. For users with multiple orgs, we still return old auth tokens, due to the chicken-egg problem described in #357.

src/utils/clack-utils.ts
romtsn: sounds good, I was just not sure which approach is better. Will change that 👍

romtsn: Addressed in https://github.com/getsentry/sentry-wizard/pull/414/commits/4473bd5cee80a36300e703769a5ae240172eff60

src/utils/clack-utils.ts
romtsn: yeah I think I'm gonna skip this for now, I think it better fits PR where we switch to use org auth tokens

Lms24 (approved):
LGTM, thanks for addressing my feedback!

PR #415, opened 2023-09-01
feat(sourcemaps): Provide exit path if there's no need to upload sourcemaps

### LINKED ISSUE #397: Provide an exit path for users who don't need to upload source maps
If users who don't bundle, transpile, minify or build their JS in any way end up in the wizard, the tool selection step currently doesn't offer any viable path for them. Selecting "None of the Above" will start the CLI flow which is unnecessary. Instead, we should provide an exit path for them at this point as they don't need to configure source maps. 

**Proposed solution:**

split “None of the above” into
* I use another build tool (=> Sentry CLI)
* I don’t minify/bundle/transpile my code (=> exit as there’s nothing to do)

Lms24:
closes #397

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ brew update
 brew install getsentry/tools/sentry-wizard
 ```
 
-- feat(sourcemaps): Provide exit path if there's no need to upload sourcemaps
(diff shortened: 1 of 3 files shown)


lforst (approved):
seems good 👍

PR #417, opened 2023-09-04
feat: Add bun package manager support

HazAT:
Adds support to pick Bun.js first as the preferred package manager.


Lms24 (commented):
Thanks! 
Can we add the detection and installation to the newer wizards, too?

* Detection/selection:
   https://github.com/getsentry/sentry-wizard/blob/9a37913697f26cb712a354aac0d938fc0b872f8a/src/utils/clack-utils.ts#L463-L497 

* Installation:
   https://github.com/getsentry/sentry-wizard/blob/9a37913697f26cb712a354aac0d938fc0b872f8a/src/utils/clack-utils.ts#L192-L198

src/utils/clack-utils.ts
```diff
@@ -185,17 +192,11 @@ export async function installPackage({
   sdkInstallSpinner.start(
     `${alreadyInstalled ? 'Updating' : 'Installing'} ${chalk.bold.cyan(
       packageName,
-    )} with ${chalk.bold(packageManager)}.`,
+    )} with ${chalk.bold(packageManager.getName())}.`,
   );
 
   try {
-    if (packageManager === 'yarn') {
```
HazAT, line 192: the `PackageManager` doesn't install `@latest` - I am not sure if we should do this but if there was a good reason for it, we might add it as default.

lib/Helper/PackageManager.ts
```diff
@@ -5,6 +5,9 @@ import * as path from 'path';
 import { promisify } from 'util';
 
 export function getPackageManagerChoice(): PackageManager | null {
```
lforst, line 7: If we use it I would rename this function to something like `detectPackageManager` or `detectPackageManager`.

lforst (commented):
Is there a reason we're moving to classes again?

In general i don't think we like classes anymore and also to me it seems like it's complicating things again.

Lms24 (commented):
Sorry for not making this clear in my initial review but I'd prefer it if we moved away from classes, given that the new wizards by choice only use helper functions to keep things simple. Don't want to dwell on this too much so my suggestion would be to revert the refactor change and add extend the helper functions `clack-utils` in addition.

HazAT:
OK, will refactor, just used what was there already

HazAT:
So it's refactored now, only React Native is not using clack now - that's why those functions are in a seperated file

src/utils/package-manager.ts
```diff
@@ -0,0 +1,61 @@
...
+export const packageManagers = [bun, yarn, pnpm, npm];
+
+export function detectPackageManger(): PackageManager {
+  for (const packageManager of packageManagers) {
+    if (fs.existsSync(path.join(process.cwd(), packageManager.lockFile))) {
+      return packageManager;
+    }
+  }
+  // We make the default NPM - it's weird if we don't find any lock file
+  return npm;
```
Lms24, line 53: m: I think the problem here are monorepos. For monorepo users, we expect folks to run the wizard in the respective package and not in the repo root. However, the lock file will most likely be in the root. 

I'd rather return `undefined`  and fall back to asking users about their package managers if we're actively installing packages. To print instructions it's totally fine IMO to fall back to NPM. Suggestion:
* `detectPackageManager` returns `undefined` if we don't find a lock file
* `detectPackageManagerWithFallback` (or something along these lines) returns NPM as the default. 

The alternative would be to be smarter about looking up a lock file but I think this is not trivial to solve reliably. Maybe we try this in a follow-up PR?

Lms24 (commented):
Thanks for refactoring! I like it a lot more now. Still have a concern about the default value.

src/utils/package-manager.ts
Lms24: Other idea: We can use similar logic to what the `detect-package-manager` package does: 

https://www.npmjs.com/package/detect-package-manager

1. check lock files
2. if none found, try to exec package manager commands
3. if none found, fall back to npm

wdyt?

HazAT:
@Lms24 I changed it not to default to `npm` in the main function and ask the user instead.
I didn't want to try to invoke because I feel like usually people have more than one installed so asking is better imho.

Lms24 (approved):
Thanks!

PR #422, opened 2023-09-04
feat(apple): Use ".sentryclirc" for auth instead of hard coding it

brustolin:
The wizard will create a `.sentryclirc` configuration file instead of writing the auth token in the build phase and fastlane.


Lms24 (approved):

Lms24:
Good catch and important change! I'm gonna add this to the wizard guidelines that we should never hard-code the auth token into users' files.

PR #432, opened 2023-09-07
feat(sourcemaps): Automatically insert Sentry Webpack plugin

### LINKED ISSUE #418: Automatically insert plugin into Webpack config
ref: https://github.com/orgs/getsentry/projects/114/views/1?pane=issue&itemId=32715986

Lms24:
This PR adds automatic Webpack plugin insertion to the sourcemaps wizard's webpack flow.

This changes the webpack flow from always showing copy/paste instructions to:
* check if a webpack config exists
  * if not automatically found, ask if users have one
    * if not, create a new webpack config with the Sentry related code
    * if yes, ask them to enter the path to their config
* modify existing config
  * add require call for webpack plugin
  * enable source map generation
  * insert the plugin
* if any if these processes fail, we fall back to showing the copy/paste instructions
 
Because this is the second time that we're adding auto insertion I took the opportunity to extract some common helper functions which we can reuse in both flows and to finally add some tests. I mainly concentrated on testing important AST modifications and helper functions as these are the most critical parts of the flow. 

This PR only handles CJS webpack configs at the moment. We can add ESM support later on, shouldn't be too much work. However, I'd argue that CJS is probably the most common config form. 

closes #418

### DIFF
--- a/src/sourcemaps/tools/webpack.ts
+++ b/src/sourcemaps/tools/webpack.ts
@@ -28,6 +28,7 @@ import {
 } from './types';
 
 import { findFile, hasSentryContentCjs } from '../../utils/ast-utils';
+import { debug } from '../../utils/debug';
 
 const getCodeSnippet = (
   options: SourceMapUploadToolConfigurationOptions,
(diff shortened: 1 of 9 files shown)


src/utils/clack-utils.ts
```diff
@@ -797,3 +798,141 @@ async function askForProjectSelection(
...
+ * code in their file. This is helpful if automatic insertion failed or is not possible/feasible.
+ *
+ * @param filename the name of the file to which the code snippet should be applied.
+ * If a path is provided, only the filename will be used.
+ * @param codeSnippet the snippet to be printed.
+ * Make sure to follow the diff-like format of highlighting lines that require changes
+ * and showing unchanged lines in gray.
+ *
+ * TODO: Link to wizard spec (develop) once it is live
+ * TODO: refactor copy paste instructions across different wizards to use this function.
```
Lms24, line 867: will do this in a follow-up PR as this one is already (too) big

src/sourcemaps/tools/webpack.ts
```diff
@@ -47,21 +89,297 @@ export const configureWebPackPlugin: SourceMapUploadToolConfigurationFunction =
...
+      }
+    }
+
+    const exportStmt = getCjsModuleExports(program);
+    if (!exportStmt) {
+      // We only care about CJS at the moment since it's probably the most widely used format for webpack configs.
+      clack.log.warn(
+        `Could not find module.exports = { /* config */ } in ${prettyConfigFilename}.
+This is fine, please follow the instructions below.`,
+      );
```
lforst, line 166: l/m: I don't think we should show this log message (or any of the log messages that we couldn't find something in the AST) and I think we should just go for the fallback without saying anything. Will people care about these messages? Otherwise, they're just noise. Wdyt? (we should still do telemetry tho)

src/sourcemaps/tools/webpack.ts
```diff
@@ -47,21 +89,297 @@ export const configureWebPackPlugin: SourceMapUploadToolConfigurationFunction =
...
+      return false;
```
lforst, line 167: m: I think we should set a tag for "ast-mod-fail-reason" here.

src/sourcemaps/tools/webpack.ts
```diff
@@ -47,21 +89,297 @@ export const configureWebPackPlugin: SourceMapUploadToolConfigurationFunction =
...
   };
+
+/**
+ * Modifies a webpack config file to enable source map generation and add the Sentry webpack plugin
+ * exported only for testing
+ */
+export async function modifyWebpackConfig(
+  webpackConfigPath: string,
+  options: SourceMapUploadToolConfigurationOptions,
+): Promise<boolean> {
```
lforst, line 124: l: Do you think it would make sense to split this function into `shouldModifyWebpackConfig(): boolean` and `modifyWebpackConfig(): void`?

src/sourcemaps/tools/webpack.ts
```diff
@@ -47,21 +89,297 @@ export const configureWebPackPlugin: SourceMapUploadToolConfigurationFunction =
...
-        message: 'Did you copy the snippet above?',
-        options: [{ label: 'Yes, continue!', value: true }],
-        initialValue: true,
-      }),
-    );
+    if (successfullyAdded) {
+      Sentry.setTag('ast-mod', 'success');
+    } else {
+      Sentry.setTag('ast-mod', 'fail');
+      await showCopyPasteInstructions(
```
lforst, line 108: man I really like these utilities we got going here!

src/utils/clack-utils.ts
```diff
@@ -797,3 +798,141 @@ async function askForProjectSelection(
...
+export async function createNewConfigFile(
+  filepath: string,
+  codeSnippet: string,
+  moreInformation?: string,
+): Promise<boolean> {
+  const prettyFilename = chalk.cyan(path.basename(filepath));
+  try {
+    await fs.promises.writeFile(filepath, codeSnippet);
+
+    Sentry.setTag('created-new-config', 'success');
```
lforst, line 919: m: I am wondering if these tags make sense in the case we create multiple configs inside one wizard run but maybe this is a "future us" problem.

src/utils/clack-utils.ts
```diff
@@ -797,3 +798,141 @@ async function askForProjectSelection(
...
+ *
+ * @param filepath path to the new config file
+ * @param codeSnippet the snippet to be inserted into the file
+ * @param moreInformation (optional) the message to be printed after the file was created
+ * For example, this can be a link to more information about configuring the tool.
+ *
+ * @returns true on sucess, false otherwise
+ */
+export async function createNewConfigFile(
+  filepath: string,
```
lforst, line 911: l/m: I am always wondering if we should be very specific here and say we only take an absPath. Otherwise it's always a bit of guessing what you should provide and where the file will end up.

If we require an abs path we could also always print the filepath in the "relative from cwd" format.

src/sourcemaps/tools/webpack.ts
```diff
@@ -47,21 +89,297 @@ export const configureWebPackPlugin: SourceMapUploadToolConfigurationFunction =
...
-    );
+    const webpackConfigPath =
+      findFile(path.resolve(process.cwd(), 'webpack.config')) ??
+      (await askForToolConfigPath('Webpack', 'webpack.config.js'));
 
-    // Intentially logging directly to console here so that the code can be copied/pasted directly
-    // eslint-disable-next-line no-console
-    console.log(getCodeSnippet(options));
+    const successfullyAdded = webpackConfigPath
+      ? await modifyWebpackConfig(webpackConfigPath, options)
```
lforst, line 97: l/m: Should we log something along the lines of "Modified your config. Please double check whether it makes sense?" when we modified the config?

test/utils/ast-utils.test.ts
```diff
@@ -35,10 +40,40 @@ describe('AST utils', () => {
...
     ])(
-      "reutrns false for modules without a valid '@sentry/' import",
+      "returns false for modules without a valid '@sentry/' import",
       (code) => {
         expect(hasSentryContent(parseModule(code))).toBe(false);
       },
     );
   });
+
+  describe('hasSentryContentCjs', () => {
```
lforst, line 50: l/m: Do you think it makes sense to think about the esm case? Maybe we can catch both by just looking for "@sentry/vite-plugin" in the file?

lforst (approved):
I have a few thoughts but nothing majorly blocking. In any case: nice job!

src/sourcemaps/tools/webpack.ts
Lms24: good idea. I think it makes sense in both cases (modified and added) because if we added a new file, users have to most likely adjust their current way of building.

src/sourcemaps/tools/webpack.ts
Lms24: Fair, I agree, it's a little too verbose here. I'll replace it with a `debug` call. Might still be useful to keep this around for future debugging.

src/utils/clack-utils.ts
Lms24: Hmm that's a good point. Right now, we only call it once but lemme refactor this to take out the tags setting and add to the docs that the function caller is responsible for collecting telemetry on this.

src/utils/clack-utils.ts
Lms24: Good call, at least for this helper that actually writes to a file. I added a check and a test.

test/utils/ast-utils.test.ts
Lms24: Yeah, we can check for a (String)Literal starting with `@sentry/`. Tests seem to pass with this for esm and cjs. I'll change it.

test/utils/clack-utils.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    const code = `module.exports = {/*config...*/}`;
+
+    const result = await createNewConfigFile(filename, code);
+
+    expect(result).toBe(false);
+    expect(writeFileSpy).toHaveBeenCalledWith(filename, code);
+    expect(clackMock.log.warn).toHaveBeenCalledTimes(1);
+  });
+
+  it('returns fals if the passed path is not absolute', async () => {
```
lforst, line 134:
```suggestion
  it('returns false if the passed path is not absolute', async () => {
```

lforst (approved):

PR #435, opened 2023-09-08
enh(android): Add more telemetry

romtsn:
* Add more telemetry for the Android wizard
* Validate and ask for the gradle file if a wrong one was given in a custom build logic scenario

### DIFF
--- a/src/android/gradle.ts
+++ b/src/android/gradle.ts
@@ -45,20 +45,17 @@ export async function selectAppFile(
 
   if (appFiles.length === 0) {
     Sentry.setTag('custom-build-logic', true);
-    let appFile = await abortIfCancelled(
+    const appFile = await abortIfCancelled(
       clack.text({
         message: `Unable to find your app's directory. 
-        Please enter the relative path to your app's build.gradle file from the root project (e.g. "app/build.gradle.kts")`,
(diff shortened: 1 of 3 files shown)


src/android/gradle.ts
```diff
@@ -45,12 +45,20 @@ export async function selectAppFile(
...
   if (appFiles.length === 0) {
     Sentry.setTag('custom-build-logic', true);
-    const appFile = await abortIfCancelled(
+    let appFile = await abortIfCancelled(
       clack.text({
         message: `Unable to find your app's directory. 
         Please enter the relative path to your app's build.gradle file from the root project (e.g. "app/build.gradle.kts")`,
       }),
     );
+    while (!appFile.includes('.gradle') || !fs.existsSync(appFile)) {
```
Lms24, line 54: l: can we solve this with clack's [`validate` option](https://github.com/natemoo-re/clack/tree/main/packages/prompts#text)?

Lms24 (approved):
Nice!

makes me think we should use `captureException` like this our wizards, too.

src/android/gradle.ts
romtsn: nice, didn't know that existed

PR #440, opened 2023-09-11
fix(android): Add support for unusual import statements

markushi:
Sometimes injecting the sentry import statements fails due to imports like

````java
import static a.b.c;
import x.y.z as D;
import a.b.*;
````

This PR adds support for that and some basic unit tests.


romtsn (approved):
Great catch, thanks 💯

Lms24 (approved):

PR #442, opened 2023-09-12
fix: upgrade glob from 7.1.6 to 7.2.3

mayank1513:
upgrade glob from 7.1.6 to 7.2.3.


Lms24:
Hi @mayank1513 thanks for opening this PR! What's the reason for this change?

Lms24:
Again, thanks for opening this PR! I'm going to close it though because we manage dependencies and versions ourselves when we see a concrete need to upgrade. In this case, I don't see a concrete reason and would rather to avoid potential breakage (low risk but still). Please feel free to contribute to our projects in other ways and thank you for understanding.

PR #443, opened 2023-09-12
fix: upgrade @sentry/cli from 1.72.0 to 1.75.2

mayank1513:
@sentry/cli from 1.72.0 to 1.75.2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "dependencies": {
     "@clack/core": "0.3.3",
     "@clack/prompts": "0.7.0",
-    "@sentry/cli": "^1.72.0",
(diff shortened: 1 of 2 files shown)


Lms24:
Hi @mayank1513 thanks for opening this PR! What's the reason for this change?

mayank1513:
Remove tech debt

Lms24:
Again, thanks for opening this PR! I'm going to close it though because we manage dependencies and versions ourselves when we see a concrete need to upgrade. In this case, I don't see a concrete reason and would rather to avoid potential breakage (low risk but still). Please feel free to contribute to our projects in other ways and thank you for understanding.

PR #455, opened 2023-09-14
feat(sveltekit): Add telemetry collection

### LINKED ISSUE #451: Add telemetry to SvelteKit wizard

Lms24:
This PR adds telemetry collection to the SvelteKit wizard. Concretely:

* Transaction and steps via `withTelemetry` and `traceStep`:
image
* capture minimal exceptions around known failure points
* set tags around
  * Svelte and Kit versions
  * Decisions to continue if precondition checks fail
  * Whether files were merged (AST mods) or newly created
  * Whether file modifications were successful

closes #451

### DIFF
--- a/src/sveltekit/sdk-setup.ts
+++ b/src/sveltekit/sdk-setup.ts
@@ -63,7 +63,7 @@ export async function createOrMergeSvelteKitFiles(
   const { dsn } = projectInfo;
 
   Sentry.setTag(
-    'server-hooks-file-strategy',
+    'client-hooks-file-strategy',
     originalClientHooksFile ? 'merge' : 'create',
   );
   if (!originalClientHooksFile) {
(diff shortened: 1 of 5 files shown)


src/sveltekit/sdk-setup.ts
```diff
@@ -59,19 +62,25 @@ export async function createOrMergeSvelteKitFiles(
 
   const { dsn } = projectInfo;
 
+  Sentry.setTag(
+    'server-hooks-file-strategy',
```
lforst, line 66: I think the labels for these tags are swapped.

lforst (dismissed):

lforst (approved):

src/sveltekit/sdk-setup.ts
Lms24: Good catch, thanks!

PR #458, opened 2023-09-14
feat(nextjs): Add telemetry collection to NextJS wizard

### LINKED ISSUE #450: Add telemetry to NextJS wizard

Lms24:
This PR adds telemetry data collection to the NextJS wizard. Specifically:

- Transaction and spans:
image
- tags around 
  - Next JS major versions
  - SDK already installed
  - SDK configuration methods and results
  - Detected package manager (this tag will be added to all wizards with telemetry)

This PR is unfortunately a little messy because I decided to pull a lot of the `traceStep` calls into the helper functions rather than wrapping the individual calls to the helpers. IMO this is better as we can reuse already existing manual instrumentation when we use these helpers. Also, if telemetry is disabled or the helper is used in a wizard that doesn't support telemetry, the `traceStep` wrapper will no-op anyway. 

closes #450

### DIFF
--- a/src/android/android-wizard.ts
+++ b/src/android/android-wizard.ts
@@ -165,9 +165,8 @@ async function runAndroidWizardWithTelemetry(
       'sentry.properties',
     )} file.`,
   );
-  await traceStep('Add SentryCli Config', () =>
-    addSentryCliConfig(authToken, proguardMappingCliSetupConfig),
-  );
+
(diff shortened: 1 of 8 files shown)


Lms24:
Thought: We can bump the SDK and test the new span APIs. Will open a separate PR for this.

AbhiPrasad (approved):

PR #466, opened 2023-09-21
fix(remix): Use `captureRemixServerException` inside `handleError`.

onurtemizkan:
Fixed Remix v2's `handleError` template, that was mistakenly invoking `Sentry.captureRemixErrorBoundaryError`

Replaced it with `Sentry.captureRemixServerException`

And as it's optional, removed manual handling of non-error capturing from the template.

### DIFF
--- a/src/remix/templates.ts
+++ b/src/remix/templates.ts
@@ -5,11 +5,7 @@ export const ERROR_BOUNDARY_TEMPLATE_V2 = `const ErrorBoundary = () => {
 };
 `;
 
-export const HANDLE_ERROR_TEMPLATE_V2 = `function handleError(error) {
-  if (error instanceof Error) {
-    Sentry.captureRemixErrorBoundaryError(error);
-  } else {
-    Sentry.captureException(error);
-  }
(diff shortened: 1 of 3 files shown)


src/remix/templates.ts
```diff
@@ -5,11 +5,7 @@ export const ERROR_BOUNDARY_TEMPLATE_V2 = `const ErrorBoundary = () => {
...
 `;
 
-export const HANDLE_ERROR_TEMPLATE_V2 = `function handleError(error) {
-  if (error instanceof Error) {
-    Sentry.captureRemixErrorBoundaryError(error);
-  } else {
-    Sentry.captureException(error);
-  }
+export const HANDLE_ERROR_TEMPLATE_V2 = `function handleError(error, { request }) {
+  Sentry.captureRemixServerException(error, 'remix.server', { request });
```
wilcoschoneveld, line 9: @onurtemizkan @AbhiPrasad 

this won't work;

needs to be:
`Sentry.captureRemixServerException(error, 'remix.server', request);`

without `{ }` around `request`

onurtemizkan: Thanks for reporting @wilcoschoneveld! Opened a fix PR. #469

PR #467, opened 2023-09-22
feat(Apple):More Telemetry

### LINKED ISSUE #452: iOS wizard success telemetry
The iOS wizard will attempt the following.

- [x] Update your AppDelegate or SwiftUI App Initializer with the default Sentry configuration and an example error
- [x] Add a new Upload Debug Symbols phase to your xcodebuild build script
- [x] Create .sentryclirc with an auth token to upload debug symbols (this file is automatically added to .gitignore)
- [x] When you're using Fastlane, it will add a Sentry lane for uploading debug symbols
- [x] We should have a tag for each attempted, and on the tag if it was a success, if not, if not ideally why
  - the why should be categorical ideally and not include any PII

brustolin:
Added more tags to track the process.

Closes #452 

_#skip-changelog_

### DIFF
--- a/src/apple/apple-wizard.ts
+++ b/src/apple/apple-wizard.ts
@@ -28,7 +28,6 @@ import {
   confirmContinueIfNoOrDirtyGitRepo,
   getOrAskForProjectData,
 } from '../utils/clack-utils';
-import { has } from 'lodash';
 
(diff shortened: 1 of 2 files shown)


Lms24 (approved):

smeubank:
looks good!

PR #472, opened 2023-09-28
fix(remix): Instrument existing root ErrorBoundary

onurtemizkan:
In the current version, if there is an `ErrorBoundary` exported inside root route, Remix wizard treats it as already instrumented.

This PR adds support for auto-instrumenting existing root error boundaries.
This is also required to properly instrument Hydrogen apps.


AbhiPrasad (approved):

AbhiPrasad:
we can merge after changelog item!

AbhiPrasad:
Oh wait non of the CI items are required 🤔 

we should fix that

PR #473, opened 2023-09-28
feat(rn): Update React Native wizard to use clack prompts

krystofwoldrich:
This PR updates the RN Wizard implementation to use clack for prompts and shared tools from `src/**/*`. This PR aims to change the wizard behavior as little as possible and only focuses on the clack changes. For this reason, telemetry was not implemented and is disabled at the moment. (Will be added in follow up PR.)

`lib/**/*` RN implementation was completely removed.

**This is a breaking change for RN Wizard**

- `--skip-connect` is not supported anymore
- `--quite` is not supported anymore
- `--platform` is not supported anymore (the SDK check is android, ios are present and patch them if they are)
- `--signup` is not supported anymore (the wizard new impl asks if the user has an account)

Example of successful install flow:
Screenshot 2023-09-29 at 15 08 16

Example of uninstall flow:
Screenshot 2023-09-29 at 15 15 29

Example of pod install error:
Screenshot 2023-09-29 at 14 08 25


src/apple/cocoapod.ts
```diff
@@ -40,21 +41,31 @@ export async function addCocoaPods(projPath: string): Promise<boolean> {
...
+export async function podInstall(dir = '.') {
+  const installSpinner = clack.spinner();
+  installSpinner.start("Running 'pod install'. This may take a few minutes...");
 
   try {
-    await bash.execute('pod install --silent');
-    loginSpinner.stop('Running "pod install"');
+    await bash.execute(`cd ${dir} && pod repo update`);
+    await bash.execute(`cd ${dir} && pod install --silent`);
+    installSpinner.stop('Pods installed.');
```
Lms24, line 58: l/q: do we need to `cd` back to the previous dir or are we fine?

src/react-native/react-native-wizard.ts
```diff
@@ -0,0 +1,256 @@
...
+  const xcodeProjectPath = getFirstMatchedPath(XCODE_PROJECT);
+  if (!xcodeProjectPath) {
+    clack.log.warn(
+      `Could not find Xcode project file using ${chalk.bold(XCODE_PROJECT)}.`,
+    );
+    return;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+  const xcodeProject = xcode.project(xcodeProjectPath);
```
Lms24, line 201: l: just wondering if there's a @types/xcode package or some other way of getting types for `xcode` so that we can get rid of the eslint ignores?

src/react-native/react-native-wizard.ts
```diff
@@ -0,0 +1,256 @@
...
+function addSentryInit({ dsn }: { dsn: string }) {
+  const prefixGlob = '{.,./src}';
+  const suffixGlob = '@(j|t|cj|mj)s?(x)';
+  const universalGlob = `App.${suffixGlob}`;
+  const jsFileGlob = `${prefixGlob}/+(${universalGlob})`;
+  const jsPath = getFirstMatchedPath(jsFileGlob);
+  if (!jsPath) {
+    clack.log.warn(
+      `Could not find main App file using ${chalk.bold(jsFileGlob)}.`,
+    );
```
Lms24, line 135: l: If this happens, do users need to do something manually? If yes, can we tell them what to do (add a link to docs or provide copy/paste code)?

src/react-native/react-native-wizard.ts
```diff
@@ -0,0 +1,256 @@
...
+    return;
+  }
+  const jsRelativePath = path.relative(process.cwd(), jsPath);
+
+  const js = fs.readFileSync(jsPath, 'utf-8');
+  if (
+    doesJsCodeIncludeSdkSentryImport(js, { sdkPackageName: RN_SDK_PACKAGE })
+  ) {
+    clack.log.warn(`${chalk.bold(jsRelativePath)} already includes Sentry.`);
```
Lms24, line 144: l: Basically same question here. Let's at least tell them that we're skipping adding the init call.

src/react-native/react-native-wizard.ts
```diff
@@ -0,0 +1,256 @@
...
+  const projectsIssuesUrl = `${project.organization.links.organizationUrl}/issues/?project=${project.id}`;
+
+  clack.note(`To make sure everything is set up correctly, put the following code snippet into your application.
+The snippet will create a button that, when tapped, sends a test event to Sentry.
+
+After that check your project issues:
+
+${chalk.cyan(projectsIssuesUrl)}
+
+<Button title='Try!' onPress={ () => { Sentry.captureException(new Error('First error')) }}/>`);
```
Lms24, line 172: l: To make copy/pasting easier, wdyt about using the `showCopyPasteInstructions` and `makeCodeSnippet` helpers? Doesn't strictly have to happen though, so feel free to disregard.

src/react-native/react-native-wizard.ts
```diff
@@ -0,0 +1,256 @@
...
+  await addSentryCliConfig(authToken, {
+    ...propertiesCliSetupConfig,
+    name: 'source maps and iOS debug files',
+    filename: 'android/sentry.properties',
+  });
+
+  const appBuildGradlePath = getFirstMatchedPath(APP_BUILD_GRADLE);
+  if (!appBuildGradlePath) {
+    clack.log.warn(
+      `Could not find Android ${chalk.bold(
```
Lms24, line 226: nit: Not just here but generally, I'd prefer filenames being `chalk.cyan` as defined [here](https://develop.sentry.dev/sdk/setup-wizards/#colors). Not a blocker though.

src/react-native/react-native-wizard.ts
```diff
@@ -0,0 +1,256 @@
...
+      `Android ${chalk.bold('app/build.gradle')} file already includes Sentry.`,
+    );
+    return;
+  }
+
+  const patchedAppBuildGradle = addRNSentryGradlePlugin(appBuildGradle);
+  if (doesAppBuildGradleIncludeRNSentryGradlePlugin(patchedAppBuildGradle)) {
+    clack.log.success(
+      chalk.greenBright(
+        `Added Sentry RN Gradle Plugin to ${chalk.bold('app/build.gradle')}.`,
```
Lms24, line 247: nit (sorry, feel free to ignore): Basically continuing my comment around colors, wdyt about removing `greenBright` and making the filename `cyan`? Our coloring guides aren't too specific around when to make a success message entirely green and when `clack.sucess` is enough. IMO it's okay to e.g. only make the final success message `greenBright` but feel free to go with whatever you prefer. (We're not super consistent here, yet, anyway.)

src/react-native/react-native-wizard.ts
```diff
@@ -0,0 +1,256 @@
...
+export const SUPPORTED_RN_RANGE = '>=0.69.0';
+
+export async function runReactNativeWizard(
+  options: ReactNativeWizardOptions,
+): Promise<void> {
+  options.telemetryEnabled = false;
+
+  if (options.uninstall) {
+    return runReactNativeUninstall(options);
+  }
```
Lms24, line 61: Ahh this is the first clack-based wizard with an uninstall functionality. I like the simplicity of this approach!

src/react-native/javascript.ts
```diff
@@ -0,0 +1,22 @@
+export function addSentryInitWithSdkImport(
```
Lms24, line 1: l: comment is for both function in this file: The regex approach is generally fine and we can use it as a starting point. If we experience problems, I'd recommend using `recast` and recast builders to insert the init call or check for Sentry code. For the latter, we have a helper in `ast-utils.ts`

Lms24 (approved):
Great work! I had a couple of comments but nothing that I'd consider blocking. Feel free to disregard my comments regarding coloring or UI, I'm just trying to adhere to our develop specifications. We can also just iterate on this later.  
(sorry for the individual reviews instead of one w/ many comments, seems I clicked the wrong button in VSCode every time 😅 )

src/react-native/react-native-wizard.ts
krystofwoldrich: Sadly there is none -> https://github.com/apache/cordova-node-xcode/issues/64, but I think for the few functions we could add the types ourselves. Maybe next PR.

src/react-native/react-native-wizard.ts
krystofwoldrich: I've missed that but it looks better than my current solution. I'll use it.

src/react-native/react-native-wizard.ts
krystofwoldrich: No worries, the inconsistency is why I made all the messages green, as otherwise, only some were green which draws attention to them like they are more important but in this specific case it wasn't true.

src/react-native/react-native-wizard.ts
krystofwoldrich: Thanks.

src/react-native/javascript.ts
krystofwoldrich: Yeah, when enlarging the JS patch in the future I would definitely make use of operations over the ast, but in this PR I focused on the clack changes so I keep the core functionality impl the same.

src/apple/cocoapod.ts
krystofwoldrich: we don't have to, it changes pwd only for the current command,
thanks for the question, that could be an annoying bug if it would change the pwd of the wizard process

src/react-native/react-native-wizard.ts
krystofwoldrich: I'll add copy paste code.

src/react-native/react-native-wizard.ts
krystofwoldrich: I adjusted the message

PR #479, opened 2023-10-05
fix(reactnative): Save Sentry URL, Organization and Project to `sentry.properties`

krystofwoldrich:
Fixes missing url, org and project in RN sentry.properties

- bug introduced in https://github.com/getsentry/sentry-wizard/pull/473

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - ref(reactnative): Use clack prompts and share common code (dirty repo, login) (#473)
 - feat(reactnative): Add telemetry (#477)
+- fix(reactnative): Save Sentry URL, Organization and Project to `sentry.properties` (#479)
 
 ## 3.15.0
 
--- a/lib/Steps/ChooseIntegration.ts
+++ b/lib/Steps/ChooseIntegration.ts
@@ -65,7 +65,7 @@ export class ChooseIntegration extends BaseStep {
         break;
       case Integration.reactNative:
       default:
-        integration = new ReactNative(sanitizeUrl(this._argv));
+        integration = new ReactNative(this._argv);
         break;
     }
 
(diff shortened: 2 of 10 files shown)


Lms24 (commented):
Sorry if I'm missing the obvious but just to clarify:
* org, project and url are added to `sentry.properties`
* only the auth token is added to `.sentrycli.rc`
correct? 

Asking because, we want to add the auth token to one file (which we .gitignore) and url, org, project to another one (which is safe to commit to a repo).

krystofwoldrich:
@Lms24 Auth token, org, project, and URL are all in `sentry.properties`. (That's how the old wizard behaved. But I've missed the org, project and URL properties during the clack update)

RN has `android/sentry.properties` and `ios/sentry.properties`.

I'll check if the split works with the RN Sentry plugins (it should), as then we can ignore the file with the auth token only: 
- org, project and url are added to sentry.properties
- only the auth token is added to .sentryclirc

Lms24:
Awesome, thanks for checking @krystofwoldrich! If it turns out to be problematic, let me know. Given that this is how it worked in the old wizard, I'd be fine if we continue with this for now if it's not easily fixable. Generally, though, I'd like to decrease the chances of users (accidentally) committing the auth token to their repos.

krystofwoldrich:
@Lms24 So splitting the auth token and project org URL into two files would require changes in CLI or RN Plugins first.

Lms24:
> would require changes in CLI

In how far? Do you mean `.sentryclirc` and `sentry.properties` can't co-exist? I'm a bit confused because we used to have both files in Next and SvelteKit and this worked (we moved away from `sentry.properties` though because it made little sense in JS). 

Another thing we do in the source maps wizard is configuring a Sentry CLI call with `--org`, `--project`, etc. params. Meaning the only thing we write into a config file is the auth token. The rest is just passed to this CLI call which we serialize into an npm script. Not sure if something like this would be feasible?

Lms24 (approved):
Anyway, given that right now, the new RN wizard is missing this part entirely, I don't wanna block things here. However, I'd recommend adding a `clack.warn` message telling users to avoid committing the auth token to their repo (we do this in the source maps wizard for example).

PR #481, opened 2023-10-17
fix(Cordova)  skip dynamic libraries on Cordova

lucas-zimerman:
The current script is only required when debugging on simulators, when building/deploying to real devices, the Library becomes dynamic and it's no longer required to strip the SDK.

I chose to check the `fat` library because it seems to be the safest field to be used that is language agnostic. When the file is not `fat` this strip code isn't required and so it's skipped.

Tests were done manually on a Local Sample, tested with Simulators, plugged to a real device and also deployed to the AppStore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


krystofwoldrich (approved):
Looks good.

Issue #487, opened 2023-11-02
Update Wizard for Vite in Remix 2.2


smeubank:
### Description:

Starting with Remix 2.2 has unstable support for Vite, meaning we may need to consider that Remix users may be using Vite already with this version, and at some point in the future it might be the default.

### Requirements:

Remix and source maps wizard(s) should consider this nuance and be able to react accordingly, whether that be with an initial question to the user, or detection based on some combination of dependencies, versions, and config files.

https://remix.run/blog/remix-heart-vite

### Deliverables:

- [x] Remix Wizard Update (https://github.com/getsentry/sentry-wizard/pull/495)
- [x] Source maps wizard update (https://github.com/getsentry/sentry-wizard/pull/495)
- [ ] Remix Docs update

Lms24:
Reopening because this got auto closed via #495. I belive, docs are still missing, correct @onurtemizkan ?

Lms24:
Oh, nevermind, the docs are tracked in https://github.com/getsentry/sentry-javascript/issues/9421. Closing this one then. (but please still do it anyway)

PR #495, opened 2023-11-13
feat(remix): Add Vite support.

onurtemizkan:
Resolves: #487

Adds Vite sourcemaps upload support to Remix wizard.

Also:

- Runs `npx run reveal` when `entry.client` / `entry.server` files do not exist in the project.
- Fixes a bug (using `project.name` instead of `project.slug`) while populating build upload script with the `project` argument.

### DIFF
--- a/src/remix/sdk-setup.ts
+++ b/src/remix/sdk-setup.ts
@@ -8,6 +8,7 @@ import type { ProxifiedModule } from 'magicast';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as url from 'url';
+import * as childProcess from 'child_process';
 
 // @ts-expect-error - clack is ESM and TS complains about that. It works though
 import clack from '@clack/prompts';
@@ -22,6 +23,7 @@ import { instrumentRootRouteV1 } from './codemods/root-v1';
 import { instrumentRootRouteV2 } from './codemods/root-v2';
 import { instrumentHandleError } from './codemods/handle-error';
 import { getPackageDotJson } from '../utils/clack-utils';
+import { configureVitePlugin } from '../sourcemaps/tools/vite';
 
 export type PartialRemixConfig = {
   unstable_dev?: boolean;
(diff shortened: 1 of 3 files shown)


src/remix/sdk-setup.ts
```diff
@@ -36,6 +38,52 @@ export type PartialRemixConfig = {
...
+  selfHosted: boolean,
+): Promise<void> {
+  await configureVitePlugin({
+    orgSlug: selectedProject.organization.slug,
+    projectSlug: selectedProject.slug,
+    url: sentryUrl,
+    selfHosted,
+    authToken,
+  });
+}
```
Lms24, line 86: l: Can we just directly call `configureVitePlugin`? Maybe I missed something but looks like we only call this once. WDYT?

src/remix/sdk-setup.ts
```diff
@@ -36,6 +38,52 @@ export type PartialRemixConfig = {
...
+    );
+  } else {
+    clack.log.info(
+      `Couldn't find entry files in your project. Trying to run ${chalk.cyan(
+        REMIX_REVEAL_COMMAND,
+      )}...`,
+    );
+
+    childProcess.execSync(REMIX_REVEAL_COMMAND, {
+      stdio: 'inherit',
```
Lms24, line 65: l: Can we pipe the output from the command and wrap it in a `clack` call? Currently this somewhat breaks the UI:
image

Lms24, line 65: I don't know how `remix reveal` works but should we check for the return value or try/catch it? Just to ensure we don't crash but e.g. link to the manual setup if something goes wrong.

Lms24 (commented):
This looks good, thank you Onur!

Just have a couple of small comments.

Also one question: I saw that the "remix reveal" command adds some stuff to the server entry file around handling bot vs human requests. Is this what's present if there was no entry file at all (aka the sensible defaults they talk about) or is this something in addition? I just want to make sure that people do not confuse this with Sentry-relevant code.

src/remix/sdk-setup.ts
onurtemizkan: Yes, makes sense 👍 updated -> [177ab2b](https://github.com/getsentry/sentry-wizard/pull/495/commits/177ab2b598bdce4853138b4e04bce989c3f5f9db)

src/remix/sdk-setup.ts
onurtemizkan: This is handled on the `remix-wizard.ts`. This will throw on any non-zero exit code, which will be caught on the upper level.

onurtemizkan: Updated 👍 -> [a184b9f](https://github.com/getsentry/sentry-wizard/pull/495/commits/a184b9f1915ef27ef8a309d69d3afeb39c8f5101)

onurtemizkan:
>  I saw that the "remix reveal" command adds some stuff to the server entry file around handling bot vs human requests. Is this what's present if there was no entry file at all (aka the sensible defaults they talk about) or is this something in addition? I just want to make sure that people do not confuse this with Sentry-relevant code.

As far as I know, those are also a part of their sensible defaults. I have seen them on default Remix application templates too.

src/remix/sdk-setup.ts
Lms24: Ohh right, I missed this, thanks!

Lms24 (approved):
Great, LGTM now, thanks!

Lms24 (commented):
Oh, please add an entry to `CHANGELOG.md` under the `## Unreleased` section.

PR #496, opened 2023-11-13
feat(nextjs): Add instructions for custom _error page

lforst:
Following https://github.com/getsentry/sentry-docs/pull/8479 we want the wizard to provide you instructions on how to instrument React rendering errors.

### DIFF
--- a/src/nextjs/nextjs-wizard.ts
+++ b/src/nextjs/nextjs-wizard.ts
@@ -23,15 +23,18 @@ import {
 } from '../utils/clack-utils';
 import { SentryProjectData, WizardOptions } from '../utils/types';
 import {
+  getFullUnderscoreErrorCopyPasteSnippet,
   getNextjsConfigCjsAppendix,
   getNextjsConfigCjsTemplate,
   getNextjsConfigEsmCopyPasteSnippet,
   getNextjsSentryBuildOptionsTemplate,
   getNextjsWebpackPluginOptionsTemplate,
   getSentryConfigContents,
+  getSentryDefaultUnderscoreErrorPage,
   getSentryExampleApiRoute,
   getSentryExampleAppDirApiRoute,
   getSentryExamplePageContents,
+  getSimpleUnderscoreErrorCopyPasteSnippet,
 } from './templates';
 import { traceStep, withTelemetry } from '../telemetry';
 import { getPackageVersion, hasPackageInstalled } from '../utils/package-json';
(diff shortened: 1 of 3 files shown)


src/nextjs/nextjs-wizard.ts
```diff
@@ -83,6 +86,109 @@ export async function runNextjsWizardWithTelemetry(
...
+      clack.log.success(
+        `Created ${chalk.bold(path.join(...pagesLocation, '_error.jsx'))}.`,
+      );
+    } else if (
+      fs
+        .readFileSync(
+          path.join(process.cwd(), ...pagesLocation, underscoreErrorPageFile),
+          'utf8',
+        )
+        .includes('getInitialProps')
```
Lms24, line 135: l: This could lead to a false positive for commented out code but I guess it's fine

src/nextjs/nextjs-wizard.ts
```diff
@@ -83,6 +86,109 @@ export async function runNextjsWizardWithTelemetry(
...
+        )}:`,
+      );
+
+      // eslint-disable-next-line no-console
+      console.log(
+        getFullUnderscoreErrorCopyPasteSnippet(
+          underscoreErrorPageFile === '_error.ts' ||
+            underscoreErrorPageFile === '_error.tsx',
+        ),
+      );
```
Lms24, line 174: same question here

src/nextjs/nextjs-wizard.ts
```diff
@@ -83,6 +86,109 @@ export async function runNextjsWizardWithTelemetry(
...
+      clack.log.info(
+        `It seems like you already have a custom error page.\n\nPlease put the following function call in the ${chalk.bold(
+          'getInitialProps',
+        )}\nmethod of your custom error page at ${chalk.bold(
+          path.join(...pagesLocation, underscoreErrorPageFile),
+        )}:`,
+      );
+
+      // eslint-disable-next-line no-console
+      console.log(getSimpleUnderscoreErrorCopyPasteSnippet());
```
Lms24, line 146: l/m: Should we use the copy/paste instruction helper function from `clack-utils.ts` instead?

lforst: Oh, I forgot this existed!

src/nextjs/nextjs-wizard.ts
lforst: I saw this more as a heuristic of "are people aware of getInitialProps and what it does". If they have it commented out somewhere, they're likely aware of it's existence and can google on how to use it. If it's not there at all people might not know it and then we need to have a little bit more hand-holding.

src/nextjs/nextjs-wizard.ts
lforst: Now that I look at it, I think I'd like to have a slightly different message than in the helper. Are you okay with me leaving it like this or should I make adjustments?

Lms24 (approved):

PR #499, opened 2023-11-16
feat(reactnative): Use Xcode scripts bundled with Sentry RN SDK

krystofwoldrich:
Since RN SDK 5.11.0 we ship Xcode scripts which replace directly patch code into the build phases of the RN Xcode project.

This PR updates the wizard to detect the SDK version which shipps with the scripts and use them to create the Xcode build phases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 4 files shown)


Lms24 (approved):
LGTM from general wizard perspective. Can't say much about the correctness for RN though.

PR #500, opened 2023-11-16
fix(wizard): Update wizard API data type and issue stream url creation

krystofwoldrich:
This PR updates the wizard API types to reflect recent changes.

Updates API data usage. And issue stream url creating. Adds helper function for the URL.

- Fix RN issue stream url
- Fix remix builds script -> use project slag
- Remove `project-platform` tag as the value is not send by the API

Context:
- https://github.com/getsentry/sentry/pull/59153

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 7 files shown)


Lms24 (approved):
Thanks!

PR #501, opened 2023-11-16
feat(reactnative): Make `pod install` step optional

krystofwoldrich:
This PR adds dialog before running `pod install`. 

This step can take minutes and is prone to fail. This way users can avoid it and run it later, when their environment is setup for it.

### DIFF
--- a/src/react-native/react-native-wizard.ts
+++ b/src/react-native/react-native-wizard.ts
@@ -377,13 +377,14 @@ async function patchAndroidFiles(config: RNCliSetupConfigContent) {
 
 async function confirmPodInstall(): Promise<boolean> {
   return traceStep('confirm-pod-install', async () => {
-    const continueWithPodInstall: boolean = await abortIfCancelled(
+    const continueWithPodInstall = await abortIfCancelled(
       clack.select({
         message:'Do you want to run `pod install` now?',
         options: [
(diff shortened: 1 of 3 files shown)


brustolin (approved):
LGTM

src/react-native/react-native-wizard.ts
```diff
@@ -374,3 +374,20 @@ async function patchAndroidFiles(config: RNCliSetupConfigContent) {
     chalk.green(`Android ${chalk.cyan('app/build.gradle')} saved.`),
   );
 }
+
+async function confirmPodInstall(): Promise<boolean> {
+  return traceStep('confirm-pod-install', async () => {
+    const continueWithPodInstall = await abortIfCancelled(
+      clack.select({
+        message:'Do you want to run `pod install` now?',
```
Lms24, line 382: l: Should we hint something like "This might take several minutes"? 

Feel free to disregard if this is this doesn't make sense, I obviously lack a lot of context here

Lms24 (approved):

src/react-native/react-native-wizard.ts
krystofwoldrich: Yes, makes sense, we have a similar message at the spinner for the pods install.

PR #504, opened 2023-11-25
feat(remix): Add instrumentation step for Express server adapters.

onurtemizkan:
Adds support for instrumenting Express server adapters in Remix projects.

Summary:

- Checks if `@remix/express` is in the dependency list.
- If so, attempts to find where it's imported (does not perform a deep file traversal, there's a list of potential paths)
- Wraps `createRequestHandler` function and updates its usages.
- If Vite is used, attempts to update `build` parameter for the dev mode. (Ref: https://github.com/getsentry/sentry-javascript/issues/9500#issuecomment-1812309813)


AbhiPrasad (approved):
lets get the changelog conflict updated and we can merge!

Issue #498, opened 2023-11-15
Add Sentry Metro Plugin for Expo 


krystofwoldrich:
- Subtask of https://github.com/getsentry/sentry-wizard/issues/285

Wizard should add Sentry Metro Plugin to existing config or create a new one (https://docs.expo.dev/guides/customizing-metro/) and add the plugin.

https://docs.sentry.io/platforms/react-native/sourcemaps/#manual-upload-with-hermes

```js
const { getDefaultConfig, mergeConfig } = require("@react-native/metro-config");
const {
  createSentryMetroSerializer,
} = require("@sentry/react-native/dist/js/tools/sentryMetroSerializer");

const config = {
  serializer: {
    customSerializer: createSentryMetroSerializer(),
  },
};

const m = mergeConfig(getDefaultConfig(__dirname), config);
module.exports = m;
```

PR #505, opened 2023-11-29
feat(react-native): Add support for Expo managed projects

krystofwoldrich:
- closes https://github.com/getsentry/sentry-wizard/issues/498

This PR adds support for Expo project (with and without native dirs).
The support includes:
- patch `app.json` configuration file with `@sentry/react-native/expo`
- create or patch `metro.config.json` with `@sentry/react-native/metro` `getSentryExpoConfig`
- patch Expo Xcode Build Phase with `sentry-xcode.sh` 
- add `.env.local` with Sentry Auth Token
- gitignore `.env.local`


src/react-native/metro.ts
```diff
@@ -20,33 +20,61 @@ import * as recast from 'recast';
...
+  const doesConfigExist = fs.existsSync(metroConfigPath);
+
+  if (!doesConfigExist && isExpoManagedProject) {
+    return await createExpoMinimalMetroConfigWithSentry(metroConfigPath);
+  }
+
+  if (!doesConfigExist) {
+    Sentry.setTag('metro-config-path', 'not-found');
+    return await showInstructions();
+  }
```
lucas-zimerman, line 46: what if we invert the if condition? (I can't make suggestions because i believe im not part of this project
```typescript
  const missingMetroConfig = !fs.existsSync(metroConfigPath);

  if (missingMetroConfig && isExpoManagedProject) {
    return await createExpoMinimalMetroConfigWithSentry(metroConfigPath);
  }

  if (missingMetroConfig ) {
    Sentry.setTag('metro-config-path', 'not-found');
    return await showInstructions();
  }
```

lucas-zimerman, line 46:
```suggestion
  const missingMetroConfig = !fs.existsSync(metroConfigPath);

  if (missingMetroConfig && isExpoManagedProject) {
    return await createExpoMinimalMetroConfigWithSentry(metroConfigPath);
  }

  if (missingMetroConfig) {
    Sentry.setTag('metro-config-path', 'not-found');
    return await showInstructions();
  }
```

src/utils/package-json.ts
```diff
@@ -2,6 +2,7 @@ export type PackageDotJson = {
   scripts?: Record<string, string | undefined>;
   dependencies?: Record<string, string>;
   devDependencies?: Record<string, string>;
+  main?: string;
```
lucas-zimerman, line 5: NIT: could we rename `main` to `mainPath` or `mainSourceFilePath`? just to be more clear that the content of main will be the path to the main source file.

src/react-native/react-native-wizard.ts
```diff
@@ -104,6 +125,19 @@ export async function runReactNativeWizardWithTelemetry(
 
   const packageJson = await getPackageDotJson();
 
+  const isExpoManaged = isExpoManagedProject(packageJson);
```
lucas-zimerman, line 128: Suggestion: We should avoid using the read package.json, instead we should reread it, to avoid using an outdated reference of it when a write operation is done. 

This may not be happening here in this PR but I am adding a comment since i noticed the same packageJson being passed to multiple locations in this PR.

For context: https://github.com/getsentry/sentry-wizard/pull/493

src/utils/package-json.ts
krystofwoldrich: I can add JSDoc, but the type represents the structure of `package.json` so we can't choose the keys.

src/react-native/react-native-wizard.ts
krystofwoldrich: Yeah makes sense, I removed passing the package JSON and made it an explicit metro package name, as that's the only value needed.

krystofwoldrich:
Yes, related docs PR here.

- https://github.com/getsentry/sentry-wizard/pull/505

krystofwoldrich:
DO NOT MERGE!

The current implementation overwrites Expo default `customSerializer`. Build of app still works, but we unexpectedly remove some processing of the bundle.

Lms24:
@krystofwoldrich is this ready for a review? (no rush, just trying to catch up with wizard stuff after PTO 😅)

src/react-native/expo-env-file.ts
```diff
@@ -0,0 +1,53 @@
...
+  const newContent = `#DO NOT COMMIT THIS\nSENTRY_AUTH_TOKEN=${options.authToken}\n`;
+
+  const added = await addToGitignore(EXPO_ENV_LOCAL_FILE);
+  if (added) {
+    clack.log.success(
+      `Added ${chalk.cyan(EXPO_ENV_LOCAL_FILE)} to .gitignore.`,
+    );
+  } else {
+    clack.log.error(
+      `Could not add ${EXPO_ENV_LOCAL_FILE} to .gitignore, please add it to not commit your auth key.`,
```
Lms24, line 23: l (feel free to ignore): we might wanna make the filename `chalk.cyan`

krystofwoldrich:
@Lms24 Thank you for the message, yes, since the morning it's ready for review.

Lms24 (approved):
LGTM from general wizard perspective! I couldn't test it though as I don't have an RN test environment, so deferring this to other reviewers/you :) 

All good to merge from my end!

src/react-native/expo-env-file.ts
krystofwoldrich: Make sense, also the sucess message has it, thanks.

PR #506, opened 2023-11-29
feat(nextjs): Add instructions on how to add a `global-error` page to Next.js App Router

lforst:
Adds instructions on how to add a global-error file to the App router for instrumentation.

### DIFF
--- a/src/nextjs/nextjs-wizard.ts
+++ b/src/nextjs/nextjs-wizard.ts
@@ -24,12 +24,14 @@ import {
 import { SentryProjectData, WizardOptions } from '../utils/types';
 import {
   getFullUnderscoreErrorCopyPasteSnippet,
+  getGlobalErrorCopyPasteSnippet,
   getNextjsConfigCjsAppendix,
   getNextjsConfigCjsTemplate,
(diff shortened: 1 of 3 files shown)


src/nextjs/nextjs-wizard.ts
```diff
@@ -189,6 +191,84 @@ export async function runNextjsWizardWithTelemetry(
...
+
+    if (!globalErrorPageFile) {
+      await fs.promises.writeFile(
+        path.join(process.cwd(), ...appDirLocation, 'global-error.jsx'),
+        getSentryDefaultGlobalErrorPage(),
+        { encoding: 'utf8', flag: 'w' },
+      );
+
+      clack.log.success(
+        `Created ${chalk.bold(
```
Lms24, line 237: l: should we use `chalk.cyan` here instead? I think we mostly use it when mentioning file names. But not sure how consistent we are with it in Nextjs, so feel free to disregard

Lms24 (approved):
Nice!

src/nextjs/nextjs-wizard.ts
lforst: oh you're right. Will update.

Issue #503, opened 2023-11-21
Add command line prompt for generating default error pages from wizard


AbhiPrasad:
Based on feedback from node.js team - can we confirm that we want to generate default error page?

Lms24:
Checked back with mobile folks for other wizards - this doesn't apply to them. Closing this issue as completed.

PR #515, opened 2023-12-12
feat(nextjs): Ask for confirmation before creating example page

Lms24:
This PR adds a prompt asking if the wizard should create an example page to test the Sentry setup.
Only if the user replies with yes, we'll create the example route and mention it in the outro.
Adding this prompt should decrease the likelihood of users accidentally comitting the example files.

image

Will follow up with adding the adjustment to other wizards in future PRs once we agreed on wording.

ref https://github.com/getsentry/sentry-wizard/issues/503

### DIFF
--- a/src/nextjs/nextjs-wizard.ts
+++ b/src/nextjs/nextjs-wizard.ts
@@ -271,11 +271,7 @@ export async function runNextjsWizardWithTelemetry(
     }
   });
 
-  const shouldCreateExamplePage = await traceStep(
-    'ask-create-example-page',
-    askShouldCreateExamplePage,
-  );
-
+  const shouldCreateExamplePage = await askShouldCreateExamplePage();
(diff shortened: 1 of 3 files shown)


src/nextjs/nextjs-wizard.ts
```diff
@@ -415,14 +424,14 @@ async function createOrMergeNextJsFiles(
     if (nextConfigJsExists) {
       Sentry.setTag('next-config-strategy', 'modify');
 
-      const nextConfgiJsContent = fs.readFileSync(
+      const nextConfigJsContent = fs.readFileSync(
```
Lms24, line 427: just some typo fixes (new VSCode extension 😅 )

AbhiPrasad (approved):

lforst (approved):
nice 👌

Issue #519, opened 2023-12-20
Deprecated  `configureScope` is used in a few templates


roottool:
## Overview

Deprecated `configureScope` is included in two templates. Currently, we are recommended to use `getCurrentScope`.

> Deprecate `configureScope` in favor of using `getCurrentScope()`
> https://github.com/getsentry/sentry-javascript/blob/812d10ad815773ca3c584b6187388c2f208a69ce/MIGRATION.md#deprecate-configurescope-in-favor-of-using-getcurrentscope

## Expected behavior

`getCurrentScope` is used in place of `configureScope`.

## Actual behavior

`configureScope` is used.

- https://github.com/getsentry/sentry-wizard/blob/a16b3cd4b0022644be4c33a77fb01d913c8586e4/src/nextjs/templates.ts#L223
- https://github.com/getsentry/sentry-wizard/blob/a16b3cd4b0022644be4c33a77fb01d913c8586e4/src/sveltekit/templates.ts#L70

Lms24:
Hey @roottool thanks for raising this! We need to adjust the templates here - you're absolutely right. I added it to our v8 TODO list and will expedite fixing this so that we don't ship soon-to-be outdated code to our users.

PR #520, opened 2023-12-20
fix(nextjs,sveltekit): Replace deprecated Sentry API calls in example page templates

Lms24:
fixes https://github.com/getsentry/sentry-wizard/issues/519

We had `configureScope` and `startTransaction` calls in the example page templates. Given we've deprecated the former and will soon deprecate the other, this PR updates the templates to use the new `startSpan` API instead.

Also shows how much simpler the new API makes this example.

### DIFF
--- a/src/nextjs/templates.ts
+++ b/src/nextjs/templates.ts
@@ -216,8 +216,9 @@ export default function Page() {
             margin: "18px",
           }}
           onClick={async () => {
-            Sentry.startSpan({
+            await Sentry.startSpan({
               name: 'Example Frontend Span',
+              op: 'test'
             }, async () => {
               const res = await fetch("/api/sentry-example-api");
               if (!res.ok) {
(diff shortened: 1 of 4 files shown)


src/nextjs/templates.ts
```diff
@@ -216,22 +216,15 @@ export default function Page() {
...
-
-            Sentry.configureScope((scope) => {
-              scope.setSpan(transaction);
-            });
-
-            try {
+            await Sentry.startSpan({
+              name: 'Example Frontend Span',
+              op: 'test'
+            }, async () => {
```
lforst, line 222: We could remove the async on the onClick handler and also the await before Sentry.startSpan

lforst (approved):

src/nextjs/templates.ts
Lms24: good point, removed it in both templates

Issue #405, opened 2023-08-26
Loading Wizard failed. Please try again in a few minutes and let us know if this issue persists


zhutoutoutousan:
### Environment

Node Version: v18.14.0
Windows 11 22H2

### Steps to Reproduce

1. Use `turborepo` to create a package based monorepo with pnpm workspace as manager
2. Follow the official sentry nextjs pendo, execute `npx @sentry/wizard -i nextjs --debug` in {TURBOREPO_PATH}/app/{DESIRED_APP_PATH}/ where `package.json` resides

### Expected Result

`Loading Wizard` should succeed and life should be right as rain 🍡 

### Actual Result
```
PS {DESIRED_APP_PATH}> npx @sentry/wizard -i nextjs --debug
Need to install the following packages:
  @sentry/wizard@3.10.0
Ok to proceed? (y) Y
{
        "_": [],
        "i": "nextjs",
        "integration": "nextjs",
        "debug": true,
        "uninstall": false,
        "skip-connect": false,
        "skipConnect": false,
        "quiet": false,
        "s": false,
        "signup": false,
        "disable-telemetry": false,
        "disableTelemetry": false,
        "$0": "C:\\Users\\Administrator\\AppData\\Local\\npm-cache\\_npx\\a8388072043b4cbc\\node_modules\\@sentry\\wizard\\dist\\bin.js"
}
Running Sentry Wizard...
version: 3.10.0 | sentry-cli version: 1.75.2
Sentry Wizard will help you to configure your project
Thank you for using Sentry :)
Skipping connection to Sentry due files already patched

{
        "integration": {
                "_argv": {
                        "_": [],
                        "i": "nextjs",
                        "integration": "nextjs",
                        "debug": true,
                        "uninstall": false,
                        "skip-connect": false,
                        "skipConnect": false,
                        "quiet": false,
                        "s": false,
                        "signup": false,
                        "disable-telemetry": false,
                        "disableTelemetry": false,
                        "$0": "C:\\Users\\Administrator\\AppData\\Local\\npm-cache\\_npx\\a8388072043b4cbc\\node_modules\\@sentry\\wizard\\dist\\bin.js"
                },
                "_isDebug": true
        }
}


{
        "integration": {
                "_argv": {
                        "_": [],
                        "i": "nextjs",
                        "integration": "nextjs",
                        "debug": true,
                        "uninstall": false,
                        "skip-connect": false,
                        "skipConnect": false,
                        "quiet": false,
[...]

PR #524, opened 2024-01-10
ref(wizard): Print error object if wizard endpoint API request failed

Lms24:
Seems like #405 is happening again and I have no idea what might be causing the errors here. 
Since I cannot reproduce this myself, let's log out the error so that users can report more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- ref(wizard): Print error object if wizard endpoint API request failed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- ref(wizard): Print error object if wizard endpoint API request failed
```
```suggestion
- ref(wizard): Print error object if wizard endpoint API request failed (#524)
```

lforst (approved):

PR #533, opened 2024-02-06
ref(remix): Replace `BrowserTracing` with `browserTracingIntegration`

Lms24:
This concludes removing all deprecated integration classes from the wizard with SDK version 7.100.0


AbhiPrasad (approved):
it's too late to be making PRs 😄

Issue #535, opened 2024-02-13
Add `spotlight: true` comment to relevant `Sentry.init` snippets  


Lms24:
Our wizard should inject a commented `spotlight: true` flag to enable sending events from the SDK to the spotlight sidecar:

```js
Sentry.init({
  dsn: '...',
  //...
  // uncomment this to enable Spotlight (https://spotlightjs.com):
  // spotlight: true
})
```

For now, the following platforms come to mind:
* NextJS server-side
* SvelteKit server-side
* Remix server-side

Client-side is taken care of by the spotlight overlay (which injects the integration into the Sentry SDK). 

Also
* React Native (?)
* Electron (?)
* Not sure about the state in iOS/Android

Shubhdeep12:
Hey @Lms24 should we do this? For the platforms where we spotlight is supported.

Lms24:
Hey, @Shubhdeep12 yes, let's give this a try! If you take a look at this, let's please first tackle the JS frameworks with one PR per framework (to keep things a bit easier to review and revert if necessary). Let me know if you have questions around the wizard.

Shubhdeep12:
Sure, i'll do it in the same way.

Shubhdeep12:
Hi @Lms24 Added the PRs

Please check.
Thanks

Lms24:
Hey @Shubhdeep12 sorry for the late reaction to this PR! I missed that you reworked it and didn't look at it earlier. 

Here's the thing: In the meantime, we implemented feature selection in the wizard, with the plan to eventually also make spotlight a selectable feature. Then we don't have to worry about the comment shenanigans and can actually enable spotlight on user selection. I would argue though that we should tackle this separately in a new issue when the time is ready for this. If we were to rebase and re-integrate this PR, we'd have to change a signficant part of the Remix SDK init call creation logic. 

Therefore, for the moment, I'm going to close this PR.

PR #545, opened 2024-03-06
feat(Next.js): Added Spotlight option in server config

Shubhdeep12:
Added comment to add spotlight in Sentry.init for next.js server config

_Note: We did not include the installation and setup of Spotlight because it is not added by default, and it is also not mandatory._

#535

### DIFF
--- a/src/nextjs/templates.ts
+++ b/src/nextjs/templates.ts
@@ -145,7 +145,7 @@ export function getSentryConfigContents(
 
   // uncomment the line below to enable Spotlight (https://spotlightjs.com)
   // spotlight: process.env.NODE_ENV === 'development',
-  `
+  `;
   }
 
   // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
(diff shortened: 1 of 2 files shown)


src/nextjs/templates.ts
```diff
@@ -151,7 +160,9 @@ Sentry.init({
   tracesSampleRate: 1,
 
   // Setting this option to true will print useful information to the console while you're setting up Sentry.
-  debug: false,${additionalOptions}
+  debug: false,
+  ${additionalOptions}
+  ${spotlightOption}
```
lforst, line 165: Looking at this change superficially I think this will mess up the output.

Shubhdeep12: Hi @lforst this was adding some extra empty lines, which I just fixed.

Is there any specific issue that you are mentioning here? which I am missing?

lforst (approved):

src/nextjs/templates.ts
lforst: That was what I was referring to. Thanks for fixing it!

Shubhdeep12:
> Hi, prettier is failing and also could you add a changelog entry? Thanks!

Hi @lforst it should be fixed now.

PR #546, opened 2024-03-06
feat(SvelteKit): Added Spotlight option in server config

Shubhdeep12:
added comment to add spotlight in Sentry.init for svelteKit server hooks config

#535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,12 @@
 
 ## Unreleased
 
-- Add note about `tunnelRoute` and Next.js middleware incompatibility (#544)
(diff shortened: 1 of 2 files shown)


Lms24 (approved):
Thanks! So there's a second part to this but I'm afraid it's not super trivial. We only add the Sentry.init call from `templates.ts` if there's no prior `hooks.server.ts` file. In case there is, we try and add our Sentry code to the existing code using AST manipulation. 

I briefly tried adding a comment into the init call options object but it completely messed up file formatting. So for now, let's just go with this in the template. We can revisit this later.

Long term, I'd like to streamline of our AST manipulation tools and remove magicast in favour of directly using recast. This should give us more freedom to make such changes ideally.

PR #547, opened 2024-03-06
feat(Remix): Added Spotlight option in server config

Shubhdeep12:
Added comment to add spotlight in Sentry.init for remix server config

#535

### DIFF
--- a/src/remix/sdk-setup.ts
+++ b/src/remix/sdk-setup.ts
@@ -102,12 +102,14 @@ function insertServerInitCall(
   dsn: string,
   originalHooksMod: ProxifiedModule<any>,
 ) {
-  const initCall = builders.functionCall('Sentry.init', {
-    dsn,
-    tracesSampleRate: 1.0,
-    // uncomment the line below to enable Spotlight (https://spotlightjs.com)
-    // spotlight: process.env.NODE_ENV === 'development',
-  });
+  const code = [
+    "Sentry.init({",
+    `  dsn: "${dsn}",`,
+    "  tracesSampleRate: 1.0,",
+    "  // uncomment the line below to enable Spotlight (https://spotlightjs.com)",
+    "  // spotlight: process.env.NODE_ENV === 'development',",
+    "});"
+  ].join("\n");
 
   const originalHooksModAST = originalHooksMod.$ast as Program;
 
@@ -118,7 +120,7 @@ function insertServerInitCall(
     0,
     // @ts-expect-error - string works here because the AST is proxified by magicast
     // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
-    generateCode(initCall).code,
+    code,
   );
 }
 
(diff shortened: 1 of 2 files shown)


src/remix/sdk-setup.ts
```diff
@@ -105,6 +105,8 @@ function insertServerInitCall(
   const initCall = builders.functionCall('Sentry.init', {
     dsn,
     tracesSampleRate: 1.0,
+    // uncomment the line below to enable Spotlight (https://spotlightjs.com)
+    // spotlight: process.env.NODE_ENV === 'development',
```
Lms24, line 109: Have you tested this in a Remix application? I don't think this works as you'd expect. Unless I'm missing something, this is a pure code comment and won't be picked up by the function call builder that generates the AST for the Sentry.init call. 

It's very closely related to what I wrote in #546 about AST manipulation. Maybe we need to find a way to make this work rather sooner than later. 

What I was trying to do earlier when reviewing #546 was using a recast builder to manually build the AST. It allows me to insert comments into an AST node like the options object. If you wanna take a look at this, we already do stuff like this in `ast-utils.ts`: https://github.com/Shubhdeep12/sentry-wizard/blob/088ffb2452d0a7a00521282a3f5de59d53d2d832/src/utils/ast-utils.ts#L121-L152

Maybe we can use this here, too. So basically change the magiacst builder to a `recast.types.builder` and manually construct things. But we have to find a way to preserve code formatting (as well as possible) when the code is generated.

Lms24 (changes requested):

src/remix/sdk-setup.ts
Shubhdeep12: Got it, looks like magicast builder is removing the comments. my bad, I think I missed checking this. 
let me try the suggested sol.

Shubhdeep12: Hi @Lms24 not sure if we can add comments in objects in recast too using builder functions. 
But with magicast and recast, there is this way to add comments viz. writing code itself instead of the builder function.
I tried to put the code directly and then parsed it. It's working fine and not removing the comments now.

Any input here? Can we do this way?
Screenshot from 2024-03-18 09-03-55

Lms24: Hey @Shubhdeep12 yup, if that gets the job done, let's do it 👍

Shubhdeep12: pushed the fix @Lms24

Lms24:
Hey @Shubhdeep12 sorry for the late reaction to this PR! I missed that you reworked it and didn't look at it earlier. 

Here's the thing: In the meantime, we implemented feature selection in the wizard, with the plan to eventually also make spotlight a selectable feature. Then we don't have to worry about the comment shenanigans and can actually enable spotlight on user selection. I would argue though that we should tackle this separately in a new issue when the time is ready for this. If we were to rebase and re-integrate this PR, we'd have to change a signficant part of the Remix SDK init call creation logic. 

Therefore, for the moment, I'm going to close this PR.  

(Going through issues and PRs as part of Sentry's triaging week)

PR #548, opened 2024-03-06
feat(React Native): Added Spotlight option in config

Shubhdeep12:
Added comment to add spotlight in Sentry.init for react-native config

#535


krystofwoldrich (approved):
Excuse the delay, I've missed the notification.

Yes, it looks good.

PR #556, opened 2024-04-10
feat(nextjs): Ask users about `tunnelRoute` option

### LINKED ISSUE #554: Callout Next.js Tunnel option and prompt if it should be enabled by default or not
We should add this as a prompt to the setup wizard to ensure users are not surprised.

Lms24:
This PR adds a new step to the NextJS wizard where we ask if users want to register the `tunnelRoute` option. Reason: Previously we set this by default but recently learned that this causes increased Vercel bills for our users due to URL rewrites being a paid feature of Vercel.

image

image

closes #554

### DIFF
--- a/src/nextjs/nextjs-wizard.ts
+++ b/src/nextjs/nextjs-wizard.ts
@@ -718,7 +718,7 @@ async function createExamplePage(
 
 /**
  * Ask users if they want to set the tunnelRoute option.
- * We can't set this by default because it increases hosting bills due to vercel charging for server-side rewrites.
+ * We can't set this by default because it potentially increases hosting bills.
  * It's valuable enough to for users to justify asking the additional question.
  */
 async function askShouldSetTunnelRoute() {
@@ -731,7 +731,7 @@ async function askShouldSetTunnelRoute() {
           {
             label: 'Yes',
             value: true,
-            hint: 'Can increase your hosting bill if you use Vercel',
+            hint: 'Can increase your server load and hosting bill',
           },
           {
             label: 'No',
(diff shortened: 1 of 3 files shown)


HazAT (approved):
👍

src/nextjs/nextjs-wizard.ts
```diff
@@ -701,3 +715,40 @@ async function createExamplePage(
...
+            value: true,
+            hint: 'Can increase your hosting bill if you use Vercel',
+          },
+          {
+            label: 'No',
+            value: false,
+            hint: 'Browser errors and events might be blocked by ad blockers before being sent to Sentry',
+          },
+        ],
+        initialValue: true,
```
HazAT, line 742:
```suggestion
        initialValue: false,
```

HazAT: Let's make it false please, I know this will cause more people to opt-out but it has to be explicit

HazAT (changes requested):

src/nextjs/templates.ts
```diff
@@ -28,10 +34,13 @@ export function getNextjsSentryBuildOptionsTemplate(): string {
     // Transpiles SDK to be compatible with IE11 (increases bundle size)
     transpileClientSDK: true,
 
-    // Routes browser requests to Sentry through a Next.js rewrite to circumvent ad-blockers. (increases server load)
+    // ${
+      tunnelRoute ? 'Route' : 'Uncomment to route'
+    } browser requests to Sentry through a Next.js rewrite to circumvent ad-blockers.
+    // This can increase your server load as well as your Vercel bill.
```
lforst, line 40: This can increase your server bill in general, not just on vercel!

Lms24:
Updated PR:
- selection defaults to "No"/`false`
- made texts more generic as this is not just a Vercel issue (TIL)

HazAT (approved):
🚢  and please prep a release for it

PR #566, opened 2024-04-24
fix(wizard): Handle missing auth token in wizard API endpoint response

### LINKED ISSUE #564: Handle missing auth token data more gracefully
(INC-727 Follow-up | ref #563)

The wizard expects a valid auth token when receiving the project selection response from the Sentry backend. INC-727 has shown that this isn't always the case so let's at least more gracefully handle this scenario. 

I see two possible ways:
1. Exit with a message that this is most likely a Sentry issue and should be reported via GH
2. Log a warning and continue without a valid auth token. Instead we set something like `authToken: "_ADD_YOUR_TOKEN_"`.

I think 2 is potentially possible but if we run into a lot of troubles with it, 1 is a good enough fallback.

Lms24:
This PR adjusts the `SentryProjectData` type to reflect that the wizard might not always receive an auth token (as we learned in INC-727). Consequently, we now handle not receiving an auth token and proceed with a dummy token and a warning for users instead:

image

closes #564

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- fix(wizard): Handle missing auth token in wizard API endpoint response
+- fix(wizard): Handle missing auth token in wizard API endpoint response (#566)
 
 ## 3.22.0
 


src/utils/clack-utils.ts
```diff
@@ -726,6 +726,7 @@ export function isUsingTypeScript() {
   }
 }
 
+const DUMMY_AUTH_TOKEN = '_YOUR_AUTH_TOKEN_';
```
AbhiPrasad, line 729:
```suggestion
const DUMMY_AUTH_TOKEN = '_YOUR_SENTRY_AUTH_TOKEN_';
```

AbhiPrasad (approved):

src/utils/clack-utils.ts
Lms24: good idea!

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- fix(wizard): Handle missing auth token in wizard API endpoint response
```
```suggestion
- fix(wizard): Handle missing auth token in wizard API endpoint response (#566)
```

PR #592, opened 2024-06-18
fix(nextjs): Fix Types of GlobalError

s1gr1d:
Adding a TS snippet for `GlobalError`. 
fixes https://github.com/getsentry/sentry-javascript/issues/12520

### DIFF
--- a/src/nextjs/templates.ts
+++ b/src/nextjs/templates.ts
@@ -393,7 +393,10 @@ export default function GlobalError({ error }: { error: Error & { digest?: strin
   return (
     <html>
       <body>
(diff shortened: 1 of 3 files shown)


src/nextjs/templates.ts
```diff
@@ -392,7 +416,7 @@ export default function GlobalError({ error }) {
   return (
     <html>
       <body>
-        <Error />
+        <NextError statusCode={error.props?.statusCode || 0} />
```
lforst, line 419: We should also update this.

lforst (approved):

PR #599, opened 2024-06-19
deps: Bump `glob` to `8.1.0`

### LINKED ISSUE #595: Bump `glob`
We get npm warnings when running the wizard because of the old `glob` version we are using.

lforst:
The changes look safe to me for our usage: https://github.com/isaacs/node-glob/blob/main/changelog.md

Fixes #595


lforst:
@lucas-zimerman, would be great if you could try this change out, since capacitor et al. are the only wizards using `glob`.

andreiborza (approved):

PR #604, opened 2024-06-21
feat(remix): Update `start` script for built-in Remix servers.

onurtemizkan:
Updates `start` script in `package.json` with `NODE_OPTIONS="--import ..."` when the server instrumentation file is not imported inside server file (or server file does not exist).

### DIFF
--- a/src/remix/sdk-setup.ts
+++ b/src/remix/sdk-setup.ts
@@ -362,6 +362,30 @@ export async function updateStartScript(instrumentationFile: string) {
     );
   }
 
+  if (packageJson.scripts.start.includes('NODE_OPTIONS')) {
+    clack.log.warn(
+      `Found existing NODE_OPTIONS in ${chalk.cyan(
+        'start',
+      )} script. Skipping adding Sentry initialization.`,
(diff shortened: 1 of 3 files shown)


src/remix/sdk-setup.ts
```diff
@@ -376,8 +353,32 @@ export async function initializeSentryOnEntryClient(
...
+
+  if (!packageJson.scripts || !packageJson.scripts.start) {
+    throw new Error(
+      "Couldn't find a `start` script in your package.json. Please add one manually.",
+    );
+  }
+
+  const startCommand = packageJson.scripts.start;
+
+  packageJson.scripts.start = `NODE_OPTIONS='--import ./${instrumentationFile}' ${startCommand}`;
```
AbhiPrasad, line 367: Should we try to detect for usage of `node xxx` or `remix-serve xxx` before replacing the `start` command?

onurtemizkan: Updated 👍

AbhiPrasad (approved):

PR #605, opened 2024-06-23
feat(nextjs): check for '[lang]' folder structure when creating example page

HaraldNordgren:
Currently, for my project where pages are placed in `src/app/[lang]`, it will place the example page in the wrong place and it has to be manually moved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - feat(remix): Switch to OTEL setup (#593)
 - feat(remix): Update `start` script for built-in Remix servers (#604)
 - deps: Bump glob to `10.4.2` (#599)
+- feat(nextjs): Check for '[lang]' folder structure when creating example page ([#605](https://github.com/getsentry/sentry-wizard/pull/605))
 
 ## 3.23.3
 
--- a/lib/Helper/File.ts
+++ b/lib/Helper/File.ts
@@ -49,6 +49,10 @@ export function exists(globPattern: string): boolean {
   }, true);
 }
 
+export function directoryExists(dirPath: string): boolean {
+  return exists(dirPath) && fs.lstatSync(dirPath).isDirectory();
+}
+
 export function matchesContent(
   globPattern: string,
   contentPattern: RegExp,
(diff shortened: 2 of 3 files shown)


HaraldNordgren:
Ping @AbhiPrasad

AbhiPrasad:
Thank you for the PR @HaraldNordgren! I put this on my todo to review

lforst:
Hi, thanks for the PR but we will likely not go through with it because `[lang]` is not a convention.

Is there a specific reason why the page doesn't work for you directly in `app/src`?

HaraldNordgren:
Hi @lforst 

Are you sure it's not a convention? It's mentioned here as the way to do internationalization: https://nextjs.org/docs/app/building-your-application/routing/internationalization#routing-overview

lforst:
The way I read the docs you shared, it is not a convention: https://nextjs.org/docs/app/building-your-application/routing/internationalization#routing-overview:~:text=Changing%20displayed%20content%20based%20on%20the%20user%E2%80%99s%20preferred%20locale%2C%20or%20localization%2C%20is%20not%20something%20specific%20to%20Next.js

Next.js has no built-in way to do this. Everything you do with regard to localization is userland code.

This question still stands 🤔:

Is there a specific reason why the page doesn't work for you directly in app/src?

HaraldNordgren:
@ifrost

It fails like this if I don't move the example page into the `[lang]` folder:

```
$ npm run build

> hava-frontend@0.9.0 build
> next build

- info Loaded env from /Users/Harald/dd/hava-web/.env.production
- warn You have enabled experimental features (instrumentationHook, outputFileTracingExcludes) in next.config.js.
- warn Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use at your own risk.

- error sentry-example-page/page.tsx doesn't have a root layout. To fix this error, make sure every page has a root layout.
- info Creating an optimized production build .
```

lforst:
Okay, now I understand. Thanks! This we need to solve differently and more generically than this PR. I created an issue to track this: https://github.com/getsentry/sentry-wizard/issues/607

Thanks for helping out!

HaraldNordgren:
Fair enough @lforst , I put up my refactoring as its own PR if it's of any interest: https://github.com/getsentry/sentry-wizard/pull/609.

PR #609, opened 2024-07-06
feat(nextjs): Refactor logic

HaraldNordgren:
Creating helper `directoryExists` to simplify logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
+- feat(nextjs): Refactor Refactor logic (#607)
+
 ## 3.25.2
 
 - ref: Improve intro and wizard selection (#625)
--- a/src/nextjs/nextjs-wizard.ts
+++ b/src/nextjs/nextjs-wizard.ts
@@ -44,7 +44,7 @@ import {
 } from './templates';
 import { traceStep, withTelemetry } from '../telemetry';
 import { getPackageVersion, hasPackageInstalled } from '../utils/package-json';
-import { getNextJsVersionBucket } from './utils';
+import { directoryExists, getNextJsVersionBucket } from './utils';
 import { configureCI } from '../sourcemaps/sourcemaps-wizard';
 
 export function runNextjsWizard(options: WizardOptions) {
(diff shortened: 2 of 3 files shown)


lib/Helper/File.ts
```diff
@@ -49,6 +49,10 @@ export function exists(globPattern: string): boolean {
   }, true);
 }
 
+export function directoryExists(dirPath: string): boolean {
```
Lms24, line 52: Can you move this to `src/nextjs/utils.ts` instead? We're trying to phase out the `lib` directory as by now it contains only code from some older wizard flows which haven't been updated in a while.

Lms24 (commented):
Hey @HaraldNordgren thanks for opening this PR! 

The change looks reasonable to me. Just wondering if there's a concrete motivation for it? Anything that's not working as you'd expect? 

Had a minor comment about file location but otherwise, we can merge this.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - feat(remix): Switch to OTEL setup (#593)
 - feat(remix): Update `start` script for built-in Remix servers (#604)
 - deps: Bump glob to `10.4.2` (#599)
+- feat(nextjs): Refactor Refactor logic ([#607](https://github.com/getsentry/sentry-wizard/pull/607))
```
Lms24, line 12: Can't suggest this in the GH UI but:

Can you add the changelog entry to a new `## Unreleased` section? Since you opened this PR we already released a new version, so the entry would be in the wrong version.

Lms24:
Oh sorry, disregard my comment about the motivation. I just found https://github.com/getsentry/sentry-wizard/pull/605. Was a bit out of the loop since I was on PTO the past couple of weeks.

lib/Helper/File.ts
HaraldNordgren: I will definitely do this, but I am currently traveling and I don’t have access to my computer for the coming weeks. I will do it as soon as I am back! 🤗

HaraldNordgren:
@Lms24 Updated the code! 🚀

HaraldNordgren:
It happens, don’t worry 🤗

For me, my other PR is more important because it’s annoying when re-running the wizard: https://github.com/getsentry/sentry-wizard/pull/610

PR #610, opened 2024-07-07
fix(nextjs): Don't add '.env.sentry-build-plugin' to .gitignore if it's already there

HaraldNordgren:


src/utils/clack-utils.ts
```diff
@@ -614,7 +616,16 @@ SENTRY_AUTH_TOKEN=${authToken}
 }
 
 async function addCliConfigFileToGitIgnore(filename: string): Promise<void> {
-  //TODO: Add a check to see if the file is already ignored in .gitignore
+  const regex = new RegExp(`^\\s*${filename}\\s*=`, 'g');
```
Lms24, line 619: Two questions:
1. why do we need the final `=`?
2. should this have `gm` for flags? 

Couldn't get the regex to match on a simple input without changing 1 and 2 (https://regex101.com/r/0I5uLE/1).

Lms24 (commented):
Hey @HaraldNordgren thanks for opening this PR! The intention looks good to me (this has been a low-prio long standing todo in my mind anyway). Just had a question about the regex.

src/utils/clack-utils.ts
HaraldNordgren: Thanks for the help @Lms24, much appreciated! I didn’t test this regex before committing and that is definitely my bad!

It was copied wholesale from the regex for SENTRY_AUTH_TOKEN in src/utils/clack-utils.ts

Lms24: No worries! The auth token itself isn't added to `.gitignore` but to an env file, hence the `=` is in there. No rush, feel free to update whenever you have time :)

HaraldNordgren: @Lms24 Updated now! 😅

lforst (approved):
Thanks!

HaraldNordgren:
@lforst Thank you!

This wizard is awesome, keep up the great work 🙌

PR #625, opened 2024-07-12
ref: Improve intro and wizard selection

### LINKED ISSUE #596: Wizard prints multiple initial logs
Some of them are confusing and duplicate.

Screenshot 2024-06-19 at 10 58 11

Lms24:
This PR refactors the wizard's startup logic, intro and integration selection.

* The general intro section now uses `clack` like most wizard flows. This improves the UI and removes a couple of redundant or unnecessary console logs (#596)
* Refactored the args parsing logic and streamlined + made it more explicit for the clack-based wizards
* Removed the old step shim classes for the clack-based wizard flows
* Bumped `yargs` to 16. Would like to bump it to 17 (latest) but this entails dropping Node 14 support. Maybe that's fine for the wizard but I'd rather do it in a separate PR and cut a major for it. 
* Old (legacy/step-based) wizards are still called in the original way. We just pipe the untouched `argv` to the old `run` function which takes care of everything.

Tested with various option, environment variable and integration combinations - seems to work fine.

This PR should bring us a big step closer to eventually get rid of the step-based wizards. 

Wizard selection:

image

After selection:

image

closes #596


bin.ts
```diff
@@ -14,11 +17,12 @@ if (!satisfies(process.version, NODE_VERSION_RANGE)) {
 }
 
 import { Integration, Platform } from './lib/Constants';
-import { run } from './lib/Setup';
```
lforst, line 17: l/m: I wonder if we should add a quick deprecation notice in lib/Setup.ts or smth like a hint not to use that file anymore.

lforst (approved):
beautiful

PR #628, opened 2024-07-23
ci: dependency review action

mdtro:
Add dependency review. Any new dependencies added with known high or critical vulnerabilities will fail CI.


mdtro:
Going to deploy this via global workflows.

Issue #627, opened 2024-07-15
[Next.js] Support Typescript `next.config.ts`


lforst:
The Next.js team recently added support for TypeScript `next.config.ts` files. The wizard currently doesn't consider TS Next.js configs. We should consider all of the possible extensions: .ts, .cts, .mts, .cjs, .mjs, .js

Ref: https://github.com/vercel/next.js/discussions/35969#discussioncomment-10005034
Ref: https://github.com/vercel/next.js/pull/63051

PR #630, opened 2024-07-26
feat(nextjs): Support all `next.config` file types.

onurtemizkan:
Resolves: https://github.com/getsentry/sentry-wizard/issues/627

Added support for `cjs`, `ts`, `cts` and `mts` file extensions.


lforst (approved):
Nice work!

PR #632, opened 2024-07-28
fix: Add missing `Error.getInitialProps` calls in Next.js error page snippets

Pikachews:
Without this call, the following error occurs when the Error page is invoked:
> TypeError: Cannot set properties of undefined (setting '_sentryTraceData')

as described in https://github.com/getsentry/sentry-javascript/issues/5906


lforst (approved):
Thanks!

Issue #636, opened 2024-08-01
Wizard seems to stall before exiting


lforst:
Currently when running the wizard, I tried with Next.js, the wizard appears to not exit when it's complete.

nikolovlazar:
+1 on this. I've also just experienced it.

Lms24:
I'll look into this soon. Likely a regression introduced in  #625 😬

PR #638, opened 2024-08-02
fix: Properly close open handles

lforst:
I used some dark magic debugging techniques and found that at the time the wizard was completing there were 2 open non-TTY handles:
- The telemetry flushing socket to Sentry (this one likely didn't prevent the process from exiting, at least until the flushing completed)
- A child process `'open -W https://sentry.io/account/settings/wizard/asdfadsf/'`

This PR should fix https://github.com/getsentry/sentry-wizard/issues/636 by adding the `wait: false` option to `opn`. I have no idea why this became a problem just now. Maybe with the recent bump of yargs in #625 it wasn't delegated to a non-blocking child process anymore or something (idk).

### DIFF
--- a/lib/Steps/OpenSentry.ts
+++ b/lib/Steps/OpenSentry.ts
@@ -48,7 +48,7 @@ export class OpenSentry extends BaseStep {
 
       const urlToOpen = urlObj.toString();
 
-      opn(urlToOpen).catch(() => {
+      opn(urlToOpen, { wait: false }).catch(() => {
         // opn throws in environments that don't have a browser (e.g. remote shells) so we just noop here
       });
 
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -8,6 +8,7 @@ import {
   runWithAsyncContext,
   setTag,
   startSpan,
+  flush,
 } from '@sentry/node';
 import packageJson from '../package.json';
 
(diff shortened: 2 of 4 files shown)


lib/Steps/OpenSentry.ts
```diff
@@ -48,7 +48,7 @@ export class OpenSentry extends BaseStep {
 
       const urlToOpen = urlObj.toString();
 
-      opn(urlToOpen).catch(() => {
+      opn(urlToOpen, { wait: false }).catch(() => {
```
Lms24, line 51: good catch! I never would have thought that this is `true` by default (looks like they flipped [from `true`](https://www.npmjs.com/package/opn/v/5.4.0#wait) [to `false`](https://www.npmjs.com/package/opn#wait) in opn 6.0.0 but we still use 5.4.0)

src/telemetry.ts
```diff
@@ -49,7 +50,12 @@ export async function withTelemetry<F>(
     throw e;
   } finally {
     sentryHub.endSession();
-    await sentryClient.flush(3000);
+    await sentryClient.flush(3000).then(null, () => {
+      // If telemetry flushing fails we generally don't care
+    });
+    await flush(3000).then(null, () => {
+      // If telemetry flushing fails we generally don't care
+    });
```
Lms24, line 58: do we need both calls here?

Lms24 (approved):
Yeah also not sure why this is becoming more of an issue now but it also appeared in the past from time to time. Maybe it was a dep bump...

Anyway, thanks for fixing! Also would love to learn more about the dark debugging magic :D

lforst:
Because it may actually be useful to readers, the dark magic debugging technique I used was `console.log(process._getActiveHandles());`.

src/telemetry.ts
Lms24: yeah no worries, let's leave it as-is

Issue #521, opened 2023-12-23
Sentry Wizard should not be opinionated 


kenn:
### Environment

Sentry Wizard for Remix

### Steps to Reproduce

1. `npx @sentry/wizard@latest -i remix`

### Expected Result

It should make minimal changes to the original source, only add wrapper, unmodified anywhere else.

<img width="1184" alt="Screenshot 2023-12-23 at 11 19 20" src="https://github.com/getsentry/sentry-wizard/assets/10350/0b8dd72b-443d-4da6-b1c3-ef7780f97e0f">

### Actual Result

It also changes how new lines should be removed at each file, or add semicolon. Respect our preference. :)

<img width="1182" alt="Screenshot 2023-12-23 at 11 19 08" src="https://github.com/getsentry/sentry-wizard/assets/10350/41f6fae5-2c64-41f4-9ad2-dcec4810e6bb">

Lms24:
Hi @kenn, thanks for writing in! 

I generally agree with you that our wizard shouldn't format things differently to your preferred style. 

However, we use `recast` which already has quite a bit of logic built in to detect the format of the file it parses and apply the same format when writing the modfied AST back to the file. 

It might be worth checking our recast settings to be sure we do our best here but other than that I believe it's unlikely that we'll ever be able to always apply the exact formatting style.

Worth noting that also the Astro CLI (which for me is the gold standard for code mod wizards) a lot of the time gets formatting wrong. It's a tricky problem to solve.

If you have concrete ideas to improve things, I'm always happy to review PRs. Leaving this open for the time being but as I said it's unlikely that we can do a lot here.

kenn:
Got it. [Prettier](https://prettier.io/) fixes it on save, so it's not a big deal. I just wanted to report in case it's missed, but feel free to close this issue!

Lms24:
We briefly discussed trying to detect if a project uses prettier and, if yes, running prettier from within the wizard. I think we didn't investigate this in greater detail due to time constraints but we could also revisit this idea.

kenn:
That sounds like a good idea! Thanks @Lms24

PR #644, opened 2024-08-05
feat: Try running Prettier on exit

onurtemizkan:
Resolves: https://github.com/getsentry/sentry-wizard/issues/521

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -1,6 +1,6 @@
 // @ts-ignore - clack is ESM and TS complains about that. It works though
 import * as clack from '@clack/prompts';
-import { abortIfCancelled } from './utils/clack-utils';
+import { abortIfCancelled, runPrettierIfInstalled } from './utils/clack-utils';
 import { runReactNativeWizard } from './react-native/react-native-wizard';
 
 import { run as legacyRun } from '../lib/Setup';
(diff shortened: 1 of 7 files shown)


Lms24:
Oof I totally missed this PR, Sorry! I'll take a look soon!

src/run.ts
```diff
@@ -127,6 +127,8 @@ export async function run(argv: Args) {
     default:
       clack.log.error(`No setup wizard selected!`);
   }
+
+  await runPrettierIfInstalled();
```
Lms24, line 131: h: let's move this call inside the individual run*wizard functions. If we leave it here, the prompt would be rendered after the outro messages which looks a bit off.

onurtemizkan: @Lms24, moved the prettier runs inside `nextjs`, `remix`, `svelte` and `react-native` wizards.

Lms24 (approved):
Thanks Onur! Tested with SvelteKit and it worked well. I think there's a limitation we might need to address depending on user feedback which is that we might need to limit applying prettier to the changed files. Should probably be doable since we know which files we generally modify. We could also use git commands to find new/changed files. 

I'll merge this in for now but feel free to work on this if you have free cycles.

PR #645, opened 2024-08-06
fix: Fix issue stream URL for self-hosted instances

maximal:
Otherwise, the wizard would print something like
```
https://org-slug.my-sentry-instance.domain.org/issues/?project=666
```
intead of the correct one
```
https://my-sentry-instance.domain.org/organizations/org-slug/issues/?project=666
```


lforst (approved):
Thanks! I think the changelog needs an entry but otherwise lgtm.

Issue #622, opened 2024-07-11
SDK feature Selection for JS Frameworks


Lms24:
Tracking feature selection in wizard for JS frameworks. For more information on the task, see parent issue

```[tasklist]
### Frameworks
- [x] NextJS - https://github.com/getsentry/sentry-wizard/pull/631
- [x] SvelteKit - https://github.com/getsentry/sentry-wizard/pull/648
- [x] Remix - https://github.com/getsentry/sentry-wizard/pull/646
```

PR #648, opened 2024-08-12
feat(sveltekit): Add feature selection

onurtemizkan:
Ref: #622

### DIFF
--- a/src/sveltekit/sdk-setup.ts
+++ b/src/sveltekit/sdk-setup.ts
@@ -66,7 +66,7 @@ export async function createOrMergeSvelteKitFiles(
       id: 'replay',
       prompt: `Do you want to enable ${chalk.bold(
         'Sentry Session Replay',
-      )} to get reproduction of frontend errors via user sessions?`,
+      )} to get a video-like reproduction of errors during a user session?`,
       enabledHint: 'recommended, but increases bundle size',
     },
   ] as const);
--- a/src/sveltekit/sdk-setup.ts
+++ b/src/sveltekit/sdk-setup.ts
@@ -96,7 +96,7 @@ export async function createOrMergeSvelteKitFiles(
       selectedFeatures,
     );
   } else {
-    await mergeHooksFile(originalClientHooksFile, 'client', dsn);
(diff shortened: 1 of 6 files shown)


src/sveltekit/sdk-setup.ts
```diff
@@ -50,6 +54,23 @@ export async function createOrMergeSvelteKitFiles(
...
+      prompt: `Do you want to enable ${chalk.bold(
+        'Tracing',
+      )} to track the performance of your application?`,
+      enabledHint: 'recommended',
+    },
+    {
+      id: 'replay',
+      prompt: `Do you want to enable ${chalk.bold(
+        'Sentry Session Replay',
+      )} to get reproduction of frontend errors via user sessions?`,
```
Lms24, line 69: The sentence reads funky to me and theoretically, replay can also surface backend errors. What about something like this?

```suggestion
      )} to get a video-like reproduction of errors during a user session?`,
```

I think we should also adjust this in the other wizards where we use `featureSelectionPrompt`.

Lms24 (commented):
Sorry for the delayed review @onurtemizkan! I gave the wizard a try with the various permutations and found some things:

- if I select "No" for tracing but already have server or client hooks files prior to running the wizard, `tracesSampleRate` is still added.
- Likewise, replay is added if I select "No" if a client hooks file exists prior to running the wizard

If I didn't have hooks files prior to running the wizard, things worked correctly 👍

Lms24:
heads-up: I'm just assigning myself to the PR to track internally that I'm reviewing it. Please continue as usual :)

onurtemizkan:
@Lms24, thanks for the review! Updated the PR.

Lms24 (commented):
Thanks for updating! Can confirm that the init calls now look correct for previously existing client and server hooks files. 

~One last thing: We should also not add the `replayIntegration` import if replay is not selected.~

Lms24 (approved):
Sorry, ignore my last review comment! I think the import was just there because I didn't reset the client hooks file correctly. LGTM!

PR #649, opened 2024-08-20
test: Add Remix e2e tests with test utils.

onurtemizkan:
Adds utilities for E2E testing wizard CLI

### Non-testing updates
- This PR adds new CLI arguments to provide pre-selected project data. This is required to run tests with dummy Auth Tokens / DSNs without opening browser / signing in. Looking at the comments, I got the impression that, this feature was already planned. Please let me know if that's not the case, and I'll try to figure out another way.

### Testing utilities
The flow of the initial e2e test structure is:

- Create a new git repo in the local test application
- Run the wizard on it with pre-provided auth token / dsn.
  - Optionally check for the `stdout` for expected prompts / messages, provide conditional `stdin`
- Check for the file contents / file existences in the application after the wizard is finished. 
- Check if the project builds successfully
- Check if the project runs on production mode successfully
- Check if the project runs in development mode successfully

In the future, we can also add new utilities to check if the events/transaction from the created example page are sent successfully to Sentry


package.json
```diff
@@ -39,10 +39,11 @@
     "semver": "^7.5.3",
     "xcode": "3.0.1",
     "xml-js": "^1.6.11",
-    "yargs": "^16.2.0"
+    "yargs": "^17.7.2"
```
lforst, line 42: Yargs 17 dropped support for Node 14 in https://github.com/yargs/yargs/pull/2384 which is why we didn't bump it earlier. I'd say though that bumping it is probably fine, I would just do it in a separate PR and cut a major release for the wizard (no big deal imo). Does that make sense?

e2e-tests/utils/index.ts
```diff
@@ -0,0 +1,292 @@
...
+  }
+}
+
+/**
+ * Run the wizard with the given integration and project directory
+ *
+ * @param integration
+ * @param projectDir
+ */
+export async function runWizard(integration: Integration, projectDir: string) {
```
lforst, line 145: What if, instead of having all of this functionality in this function, we call this function something like `startWizardInstance` and it returns the `wizardTestEnv` so you can call stuff on it inside the jest tests? That way, we're a little more dynamic in what we can test and the various permutations we can test for more complicated wizards.

package.json
onurtemizkan: I had to update it when I was experimenting but it seems we can stay at the current version with what I ended up. So just reverted the version update 👍

e2e-tests/utils/index.ts
onurtemizkan: Updated and moved to Jest runner 👍

src/run.ts
```diff
@@ -36,13 +36,52 @@ type Args = {
...
+    authToken: string;
+    selfHosted: boolean;
+    dsn: string;
+    projectId: string;
+    projectSlug: string;
+    projectName: string;
+    orgId: string;
+    orgName: string;
+    orgSlug: string;
+  };
```
onurtemizkan, line 49: Note: This overlaps with https://github.com/getsentry/sentry-wizard/pull/671. So if `preSelectedProject` is provided, the newly added `--orgSlug` and `--projectSlug` will not have any effect on the wizard behaviour.

e2e-tests/tests/remix.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    const replayOptionPrompted = await wizardInstance.waitForOutput(
+      'Do you want to enable Sentry Session Replay',
+      240_000,
+    );
+
+    if (replayOptionPrompted) {
+      wizardInstance.sendStdin(KEYS.ENTER);
+    }
+
+    const examplePagePrompted = await wizardInstance.waitForOutput(
```
lforst, line 60: I don't fully understand why we need the `optional` configuration... Would you mind explaining it to me? I feel like I'm missing something. 😄

lforst (approved):
The tests run suspicously fast 😂 but otherwise lgtm. Would you mind going over the changes I made one more time and make sure that I didn't do anything dumb? Then we can ship this! Sorry for the insane delay - totally on me.

e2e-tests/tests/remix.test.ts
onurtemizkan: It's primarily for local development where the test flows can be fragile, for example when the package manager is detected somehow and it's not prompted, or the leftovers from the example page exist and it's not prompted.

I think we can improve the utils in follow-up PRs to make them more debuggable, but this makes writing / debugging tests slightly easier in the meantime.

lforst: Seems good! 👍

onurtemizkan:
@lforst, the updates look good to me, I just increased the timeout after `yarn install`. That was timing out for me locally. Thanks for the review!

PR #652, opened 2024-08-26
feat: Detect Yarn v2+

krystofwoldrich:
- fixes: https://github.com/getsentry/sentry-react-native/issues/4043

This PR adds support for Yarn v2+

Newer version of Yarn package manager have different behaviour and different commands, it might be better to treat them as two separate package managers. 

To correctly handle Yarn v2+ is important for React Native since the default template now uses Yarn v3.

### DIFF
--- a/src/utils/package-manager.ts
+++ b/src/utils/package-manager.ts
@@ -27,9 +27,9 @@ export const BUN: PackageManager = {
   flags: '',
   detect: () => fs.existsSync(path.join(process.cwd(), BUN.lockFile)),
 };
-export const YARN_CLASSIC: PackageManager = {
+export const YARN_V1: PackageManager = {
   name: 'yarn',
-  label: 'Yarn Classic',
+  label: 'Yarn V1',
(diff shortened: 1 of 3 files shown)


src/utils/package-manager.ts
```diff
@@ -24,6 +25,26 @@ export const BUN: PackageManager = {
   buildCommand: 'bun run build',
   runScriptCommand: 'bun run',
   flags: '',
+  detect: () => fs.existsSync(path.join(process.cwd(), BUN.lockFile)),
+};
+export const YARN_CLASSIC: PackageManager = {
+  name: 'yarn',
```
lforst, line 31: Let's not call this "yarn classic" as nobody knows what the hell that is 😂 Instead I would say something like "yarn v1" and "yarn v2/v3/v4" or "yarn (>= version 2)"

krystofwoldrich: Yep, make sense.

lforst (approved):

PR #665, opened 2024-09-06
ref(nextjs): Adjust dev server command in verification message

Lms24:
Reported by @vgrozdanic 

The `next dev` command is usually not directly used by people but rather the wrapping package.json script. So this PR now shows `<pacMan run> dev` instead of `next dev`.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


vgrozdanic (approved):
Thanks a lot for fixing this ❤️

lforst (approved):

Issue #664, opened 2024-09-06
Remix Wizard creates `.sentryclirc` and `.env.sentry-build-plugin`


Lms24:
Looks like we create both `.sentryclirc` (old) and `.env.sentry-build-plugin` files, both containing the auth token to upload source maps. 

Logs are duplicated as well for both ones. We only need one of them and I'd vote for the `.env-sentry-build-plugin` file as it's newer. 

![Image](https://github.com/user-attachments/assets/f1bc9d0d-37a6-4b5d-938e-51e55d49781a)

Lms24:
@onurtemizkan before we work on this, do you remember if there's a specific reason that we add both files for remix?

PR #667, opened 2024-09-09
fix(remix): Don't create `.sentrclirc` if project uses Vite.

onurtemizkan:
Fixes: https://github.com/getsentry/sentry-wizard/issues/664

We need `.sentryclirc` file for the projects that do not use Vite, where we update the `build` script in `package.json` to use `sentry-upload-sourcemaps`. 

It is currently creating both for projects using Vite, so I moved it to where we update the build script instead.


Lms24 (approved):
Thanks for fixing!

PR #670, opened 2024-09-11
feat: Only format changed files with Prettier

onurtemizkan:
Following up on #644 

Formatting changed/untracked files only with Prettier.

### DIFF
--- a/src/utils/clack-utils.ts
+++ b/src/utils/clack-utils.ts
@@ -856,7 +856,7 @@ export async function getOrAskForProjectData(
   }
   const { url: sentryUrl, selfHosted } = await traceStep(
     'ask-self-hosted',
-    () => askForSelfHosted(options.url),
+    () => askForSelfHosted(options.url, options.saas),
   );
 
   const { projects, apiKeys } = await traceStep('login', () =>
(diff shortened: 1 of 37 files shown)


src/utils/clack-utils.ts
```diff
@@ -696,13 +706,16 @@ export async function runPrettierIfInstalled(): Promise<void> {
...
       await new Promise<void>((resolve, reject) => {
-        childProcess.exec('npx prettier --write .', (err) => {
-          if (err) {
-            reject(err);
-          } else {
-            resolve();
-          }
-        });
+        childProcess.exec(
+          `npx prettier --write ${prettierSupportsCache ? '--cache' : ''} .`,
```
Lms24, line 710: Does this work if users themselves didn't use `--cache` before? Reading the [documentation](https://prettier.io/docs/en/cli.html#--cache), I'm not sure because it sounds like a cache already needs to be populated so that a file is skipped. 

I'm not really worried about the performance aspect but rather that we format files that users (for whatever reason) haven't formatted before or don't want formatted. 

We already have some logic around detecting new and changed files via git at the beginning of the wizard. Maybe we could re-use that?

Lms24 (approved):
Tested it with a SvelteKit project. Worked like a charm. Thanks Onur (and sorry for the wait, I was ooo last week).

PR #671, opened 2024-09-12
feat: allow passing or and project slug parameters

obostjancic:
Part of https://github.com/getsentry/sentry/issues/77381

Adds `--org` and `--project` parameters that narrow down options in project selection step.
If there is only one project matching the provided slugs it gets automatically selected.
If no projects match the passed slugs wizard falls back to the current behaviour

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- allow passing or and project slug parameters
+- feat: Allow passing or and project slug parameters
   ([#671](https://github.com/getsentry/sentry-wizard/pull/671))
 
 ## 3.29.0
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- feat: Allow passing or and project slug parameters (#671)
```
```suggestion
- feat: Allow passing org and project slug parameters (#671)
```

Lms24 (approved):
LGTM! Thanks for adding this!

Side-note: We _could_ also make adjustments to our endpoint/view to take the potentially passed params into account to reduce the loading time. Right now, this change would still load a lot of unnecessary data. I'll add a note to https://github.com/getsentry/sentry-wizard/issues/357, which - despite its name - is all about reworking the project selection flow.

Issue #674, opened 2024-09-20
Wizards installing bundler plugins v2 should not emit `.sentryclirc`


Lms24:
Since v2 of the bundler plugins we no longer permit reading configuration for sentry CLI (primarily auth token) from `.sentryclirc`.

However, some(?) of our wizards still happily create this file for our users 😬 we should switch to generating the `.env.sentry-build-plugin` file instead which is picked up automatically by the plugins.

```[tasklist]
### Confirmed affected wizards
- [ ] https://github.com/getsentry/sentry-wizard/pull/675
```

```[tasklist]
### To be checked
- [x] Remix
- [x] NextJS
- [ ] Sourcemaps
- [ ] React Native
```

Lms24:
Going to close this issue since we de-prioritized all feature work related to the wizard in favour of the AI-based `sentry init` SDK setup.

PR #675, opened 2024-09-20
fix(sveltekit): Create bundler plugin env file instead of sentryclirc

lforst:
Ref https://github.com/getsentry/sentry-wizard/issues/674


shellmayr (approved):
Nice one 🙏

Lms24 (approved):
Thanks!

PR #678, opened 2024-09-26
feat: Add `--saas` CLI arg to skip self-hosted or SaaS selection step

Lms24:
Currently, our in-product onboarding shows the `--url` option because we want to skip the "Are you using Sentry.io or self-hosted" question (in-product, we already know the URL). 

From the wizard perspective, we default to the SaaS url if users don't pass a custom`--url`. However, if users pass a custom URL that is the SaaS URL we unnecessarily inject `url` params into code where it's not strictly necessary. 

The new `--saas` flag introduced in this PR will skip over the self-hosted question and if it's set to `true` assume that users are using SaaS. We can display `--saas` in the UI (for SaaS) instead of `--url`, making the snippet a bit shorter and avoiding to unnecessarily inject the `url` params into user code.

### DIFF
--- a/bin.ts
+++ b/bin.ts
@@ -60,6 +60,11 @@ const argv = yargs(hideBin(process.argv))
     alias: 'url',
     describe: 'The url to your Sentry installation\nenv: SENTRY_WIZARD_URL',
   })
+  .option('saas', {
+    default: false,
+    describe: 'If set, skip the self-hosted or SaaS URL selection process',
+    type: 'boolean',
+  })
   .option('s', {
     alias: 'signup',
     default: false,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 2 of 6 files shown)


bin.ts
```diff
@@ -60,6 +60,11 @@ const argv = yargs(hideBin(process.argv))
     alias: 'url',
     describe: 'The url to your Sentry installation\nenv: SENTRY_WIZARD_URL',
   })
+  .option('saas', {
```
Lms24, line 63: Realized we didn't add `--project` and `--org` to this list. @obostjancic was this on purpose? If not, I'll quickly add it. This ensures that `--help` lists the two options in the help message.

matejminar: I also noticed that it's missing from the readme file

Lms24: Good point. I think the readme was generally a bit outdated. I updated it, along with adding `--project` and `--org` to the readme and `--help` output in a follow-up PR: https://github.com/getsentry/sentry-wizard/pull/679

obostjancic (approved):

PR #684, opened 2024-10-09
test: Add Next.js E2E tests

onurtemizkan:
Adds e2e tests using a plain project from `create-next-app`.


e2e-tests/.env.example
```diff
@@ -0,0 +1,11 @@
...
+SENTRY_TEST_AUTH_TOKEN="TEST_AUTH_TOKEN"
+
+### SENTRY TEST DSN ###
+SENTRY_TEST_DSN="TEST_DSN"
+
+### SENTRY TEST ORG ###
+SENTRY_TEST_ORG="TEST_ORG"
+
+### SENTRY TEST PROJECT ###
+SENTRY_TEST_PROJECT="TEST_PROJECT"
```
onurtemizkan, line 11: @Lms24, @lforst -- Could we create a test project and set these env variables on the CI?

lforst: Anecdotally we should avoid depending on any production systems for our tests. We recently stopped doing so in the main js repo because it was blocking us from cutting releases and merging a tad to often. Do you think it would be a hard requirement?

onurtemizkan: This seems required for testing if the test project builds and runs in the production mode. When there's no valid `AUTH_TOKEN`, the build fails as the webpack plugin is already set on the `next.config` file. So, given there's no option to skip configuring `next.config` in the wizard at the moment, we can instead do either:

- Try reverting the local changes on `next.config` file before building the test project.
- Don't test the `build` and `start`. Only test if the project runs on `dev` mode which is working without the need for these.

lforst: Right, I think we would want to test build and start so I am on board with pointing to a prod project. We should just make sure that it is not blocking 1) Releases of the wizard 2) PRs 3) Local test runs.

We can point the tests to the `sentry-javascript-sdks` org and the `sentry-wizard-e2e-tests` project. I think this can simply be via the github action yaml. Do we need any secrets? If so I can ask for them to be added.

onurtemizkan: Cool, so adding the `AUTH_TOKEN` would be enough for now, I'll set the rest on the GA yaml.

e2e-tests/test-applications/nextjs-test-app/src/app/globals.css
```diff
@@ -0,0 +1,42 @@
+:root {
```
lforst, line 1: I would probably just remove the CSS files. They shouldn't add anything to the tests.

e2e-tests/tests/nextjs.test.ts
```diff
@@ -0,0 +1,170 @@
...
+      wizardInstance.sendStdin(KEYS.DOWN);
+      wizardInstance.sendStdin(KEYS.ENTER);
+    }
+
+    const routeThroughNextJsPrompted = await wizardInstance.waitForOutput(
+      'Do you want to route Sentry requests in the browser through your Next.js server',
+      {
+        timeout: 240_000,
+      },
+    );
```
lforst, line 44: Is this order of operations prone to race conditions? Isn't it possible that the wizard emits "Do you want to ..." before we were able to attach the .waitForOutput() call.

I know it's a pain in the butt and makes the tests super unreadable, but we should probably err on the side of caution and attach the .waitForOutput before we send the `ENTER` key, and then afterwards await it. Does that make sense? We have the same pattern in the JS repo E2E tests quite often.

lforst (commented):
Love it. 2 small comments then I would ship it. Thanks a lot!

e2e-tests/tests/nextjs.test.ts
onurtemizkan: Updated 👍 https://github.com/getsentry/sentry-wizard/pull/684/commits/4a6c56aa261943e850e62ce297e5b81ebe85a5ac

lforst (approved):
nice, thank you

PR #687, opened 2024-10-15
ref: Handle edge cases in formatting step more gracefully

Lms24:
Just noticed while running a repro sveltekit app in a new directory where I didn't yet initialize git that the auto-format step wasn't behaving correctly. For some reason, the child process promise when running `npx prettier ...` never resolved and I had to quit the wizard.

This PR makes a couple of changes to the prettier formatting step:
- bail if we're not in a git repo
- bail if no unchanged/untracked files are detected (could happen if users commit the changes while the wizard is runningI)
- only then prompt users if they want to run prettier

### DIFF
--- a/src/utils/clack-utils.ts
+++ b/src/utils/clack-utils.ts
@@ -233,7 +233,7 @@ The wizard will create and update files.`,
   });
 }
 
-function isInGitRepo() {
+export function isInGitRepo() {
   try {
     childProcess.execSync('git rev-parse --is-inside-work-tree', {
       stdio: 'ignore',
@@ -244,7 +244,7 @@ function isInGitRepo() {
   }
 }
 
-function getUncommittedOrUntrackedFiles(): string[] {
+export function getUncommittedOrUntrackedFiles(): string[] {
   try {
     const gitStatus = childProcess
       .execSync('git status --porcelain=v1')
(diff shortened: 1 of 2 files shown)


src/utils/clack-utils.ts
```diff
@@ -697,6 +684,32 @@ export async function runPrettierIfInstalled(): Promise<void> {
...
+      clack.confirm({
+        message:
+          'Looks like you have Prettier in your project. Do you want to run it on your files?',
+      }),
+    );
+
+    if (!shouldRunPrettier) {
+      return;
+    }
+
```
Lms24, line 712: No behavior changes here, just made a small refactoring to avoid the double-nested if block

src/utils/clack-utils.ts
```diff
@@ -715,7 +728,7 @@ export async function runPrettierIfInstalled(): Promise<void> {
       });
     } catch (e) {
       prettierSpinner.stop('Prettier failed to run.');
-      clack.log.error(
+      clack.log.warn(
```
Lms24, line 731: I think it's ok to downgrade this to a warning because it won't (shouldn't) cause the wizard to error. It'll just continue with the next step.

onurtemizkan (approved):
Looks good to me!

PR #692, opened 2024-10-17
Fix typo in a couple of steps

nikolovlazar:
Added a missing word "you" in a couple of the wizard steps because it read like a typo: `Did add the code to your app/global-error.tsx file as described above?`

#skip-changelog


Lms24 (approved):
thx

Lms24:
I just added a #skip-changelog comment to your PR description so we don't have to worry about the changelog entry here. Usually, we have automation in place that checks for a changelog entry but for this PR I think we don't need one. Thanks for contributing!

PR #694, opened 2024-10-22
Fix(Cordova)  enhanced node path detection, debug symbols now run in foreground.

lucas-zimerman:
This is a refactor on how the Cordova Integration runs, the main changes on the PR are:

- More path options for sentry.properties.
- Use node path from .xcode.env when node is not found.
- Skip the integration if node isn't found, avoiding a build failure when running Sentry-CLI.
- Use a more modern approach to find where Sentry-CLI is located.


lucas-zimerman:
formatted code:
```bash
#!/bin/bash
# print commands before executing them and stop on first error
set -x -e

echo "warning: uploading debug symbols - set SENTRY_SKIP_DSYM_UPLOAD=true to skip this"

[ -z "$SENTRY_FORCE_FOREGROUND"] && SENTRY_FORCE_FOREGROUND=true
[[ "$SENTRY_FORCE_FOREGROUND" == true ]] && SENTRY_FORCE_FOREGROUND_FLAG="--force-foreground"

# SETUP NODE BINARY PATH
get_node_command() {
  # First, check if NODE_BINARY is set and executable
  if [ -x "$(command -v ${NODE_BINARY})" ]; then
    echo "${NODE_BINARY}"
    return 0
  fi

  # Second, fallback to `which node`
  if [ -x "$(which node)" ]; then
    echo "node"
    return 0
  fi

  # Third, check for the highest version in the NVM directory
  NVM_NODE_VERSIONS_DIR="$HOME/.nvm/versions/node"
  if [ -d "$NVM_NODE_VERSIONS_DIR" ] && [ "$(ls -A $NVM_NODE_VERSIONS_DIR)" ]; then
    HIGHEST_VERSION=$(ls -v "$NVM_NODE_VERSIONS_DIR" | tail -n 1)
    NODE_BINARY="$NVM_NODE_VERSIONS_DIR/$HIGHEST_VERSION/bin/node"
    if [ -x "$NODE_BINARY" ]; then
      echo "$NODE_BINARY"
      return 0
    fi
  fi

  # Return empty if no node binary was found
  echo ""
  return 0
}

LOCAL_NODE_BINARY=$(get_node_command)

if [ -z "$LOCAL_NODE_BINARY" ]; then
  echo "warning: SENTRY: Node.js binary not found! Skipping symbol upload."
  exit 0
else
  echo "Using Node.js from ${LOCAL_NODE_BINARY}"
fi

# SETUP SENTRY_PROPERTIES
if [ -z "$SENTRY_PROPERTIES" ]; then
  # Check if the script is running in the root directory
  if [ -f "./sentry.properties" ]; then
    export SENTRY_PROPERTIES=sentry.properties
  elif [ -f "../../sentry.properties" ]; then
    export SENTRY_PROPERTIES=../../sentry.properties
  else
    echo "warning: SENTRY: sentry.properties file not found! Skipping symbol upload."
    exit 0
  fi
fi

echo "sentry properties found at : $(readlink -f ${SENTRY_PROPERTIES})"

# SETUP SENTRY CLI
[ -z "$SENTRY_CLI_EXECUTABLE" ] && SENTRY_CLI_PACKAGE_PATH=$("$LOCAL_NODE_BINARY" --print "require('path').dirname(require.resolve('@sentry/cli/package.json'))")
[ -z "$SENTRY_CLI_EXECUTABLE" ] && SENTRY_CLI_EXECUTABLE="${SENTRY_CLI_PACKAGE_PATH}/bin/sentry-cli"

SENTRY_COMMAND="\"$SENTRY_CLI_EXECUTABLE\" upload-dsym $SENTRY_FORCE_FOREGROUND_FLAG"

# UPLOAD DEBUG SYMBOLS
if [ "$SENTRY_SKIP_DSYM_UPLOAD" != true ]; then
  # 'warning:' triggers a warning in Xcode, 'error:' triggers an error
  set +x +e # disable printing commands otherwise we might print `error:` by accident and allow continuing on error
  SENTRY_XCODE_COMMAND_OUTPUT=$(/bin/sh -c "$LOCAL_NODE_BINARY  $SENTRY_COMMAND"  2>&1)
  if [ $? -eq 0 ]; then
    echo "$SENTRY_XCODE_COMMAND_OUTPUT"
    echo "$SENTRY_XCODE_COMMAND_OUTPUT" | awk '{print "output: sentry-cli - " $0}'
  else
    echo "error: sentry-cli - To disable debug symbols upload, set SENTRY_SKIP_DSYM_UPLOAD=true in your environment variables. Or to allow failing upload, set SENTRY_ALLOW_FAILURE=true"
    echo "error: sentry-cli - $SENTRY_XCODE_COMMAND_OUTPUT"
  fi
  set -x -e # re-enable
else
  echo "SENTRY_SKIP_DSYM_UPLOAD=true, skipping debug symbols upload"
fi
```

lib/Steps/Integrations/Cordova.ts
```diff
@@ -190,31 +190,68 @@ export class Cordova extends BaseIntegration {
         shellPath: '/bin/sh',
         shellScript:
           // eslint-disable-next-line prefer-template
```
krystofwoldrich, line 192: Can we move this to a shell file that will ship with wizard and then copy the content to create the build phase?

For example `lib/Steps/Integrations/Cordova/SentryUploadDebugFiles.sh`?

lib/Steps/Integrations/Cordova.ts
```diff
@@ -190,31 +190,68 @@ export class Cordova extends BaseIntegration {
...
-          '\\n' +
-          'function getProperty {\\n' +
-          '    PROP_KEY=$1\\n' +
-          '    PROP_VALUE=`cat $SENTRY_PROPERTIES | grep "$PROP_KEY" | cut -d\'=\' -f2`\\n' +
-          '    if [ -z "$PROP_VALUE" ]; then\\n' +
-          '        echo "plugins/sentry-cordova/node_modules/@sentry/cli/bin/sentry-cli"\\n' +
-          '    else\\n' +
-          '        echo $PROP_VALUE\\n' +
+          '[ -z "$SENTRY_FORCE_FOREGROUND"] && SENTRY_FORCE_FOREGROUND=true\\n' +
+          '[[ "$SENTRY_FORCE_FOREGROUND" == true ]] && SENTRY_FORCE_FOREGROUND_FLAG="--force-foreground"\\n' +
```
krystofwoldrich, line 198: Since https://github.com/getsentry/sentry-cli/releases/tag/2.37.0 the upload is always in foreground.

lib/Steps/Integrations/Cordova.ts
```diff
@@ -190,31 +190,68 @@ export class Cordova extends BaseIntegration {
...
+          '    HIGHEST_VERSION=$(ls -v "$NVM_NODE_VERSIONS_DIR" | tail -n 1)\\n' +
+          '    NODE_BINARY="$NVM_NODE_VERSIONS_DIR/$HIGHEST_VERSION/bin/node"\\n' +
+          '    if [ -x "$NODE_BINARY" ]; then\\n' +
+          '      echo "$NODE_BINARY"\\n' +
+          '      return 0\\n' +
           '    fi\\n' +
+          '  fi\\n' +
+          '  echo ""\\n' +
+          '  return 0\\n' +
           '}\\n' +
```
krystofwoldrich, line 219: Are we getting complains about the node not being set?
I think we should just allow to use `NODE_BINARY` if global node is not set.
We can make sure that the `NODE_BINARY` by the wizard.

For example RN [deprecated](https://github.com/facebook/react-native/blob/c684f9fc62e865a5ecbd8ade0f5628f016eebcb2/packages/react-native/scripts/xcode/with-environment.sh#L41) the dynamic node lookup

lucas-zimerman: I experienced it where node wasn't found by Xcode while it was found by the terminal

lucas-zimerman: The idea of .xcode.env seems good to me!

lib/Steps/Integrations/Cordova.ts
```diff
@@ -222,6 +223,7 @@ export class Cordova extends BaseIntegration {
         return;
       }
     }
+
```
lucas-zimerman, line 226:
```suggestion
```

lib/Steps/Integrations/Cordova.ts
```diff
@@ -168,47 +186,30 @@ export class Cordova extends BaseIntegration {
...
       if (script.shellScript.match(/SENTRY_PROPERTIES/)) {
         return;
       }
     }
-    const cwd = path.join(process.cwd(), 'sentry.properties');
+
+    const script = fs
+      .readFileSync(xcodeSymbolScriptPath, 'utf8')
+      .replace(/\\/g, '\\\\')
+      .replace(/\n/g, '\\n');
```
lucas-zimerman, line 203: image
Diff check from the shell script file compared to the same script copy/pasted on the xcode project shows the logic above matches the xcode output.

krystofwoldrich: I my comment, I thought we would add the file to the wizard, and then copy it to the build phase.

Since we are adding the file in the Cordova SDK. https://github.com/getsentry/sentry-cordova/pull/372/files

Can we simplify the code to just executing the file https://github.com/getsentry/sentry-wizard/blob/fe849febbedb56dc803f79c79d2347b75900c032/src/react-native/xcode.ts#L189

krystofwoldrich (approved):
LGTM 🚀

krystofwoldrich (approved):
LGTM 🚀

PR #695, opened 2024-10-24
test: Add Sveltekit E2E tests

onurtemizkan:
Adds E2E tests on a plain Sveltekit app.

### DIFF
--- a/e2e-tests/tests/sveltekit.test.ts
+++ b/e2e-tests/tests/sveltekit.test.ts
@@ -1,0 +1,121 @@
+import { Integration } from '../../lib/Constants';
+import {
+  checkEnvBuildPlugin,
+  checkFileContents,
+  checkFileExists,
+  checkIfBuilds,
+  checkIfRunsOnDevMode,
+  checkIfRunsOnProdMode,
+  checkPackageJson,
+  cleanupGit,
+  KEYS,
+  revertLocalChanges,
+  startWizardInstance,
+} from '../utils';
+import * as path from 'path';
+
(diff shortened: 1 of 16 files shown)


e2e-tests/tests/sveltekit.test.ts
Lms24, line 1: Can we add or adjust a test that checks for `replayIntegration` being added and `tracesSampleRate: 1.0`?

Lms24 (commented):
Thanks for adding these tests! Looks good to me. One small request but once that's addressed, we can merge the PR.

A couple of follow-up items outside the scope of this PR:

- I have a hard time getting the tests running locally because volta pins the Node version of the repo to Node 14 (which the wizard is still compatible with). SvelteKit requires Node 18 or higher so the installation fails if I simply run `yarn test:e2e:run Sveltekit`. Is there a way we could set the Node version in the e2e test apps/adjust the volta version for e2e tests to something like Node 20/22? Also, I think it'd be great to add a small README how to run the tests locally.

- If you have some time, it'd be great to also test apps where there already exist certain files that Sentry would create freshly. For example, in SvelteKit, users could already have created their own `hooks.client|server` files, so we could check that the wizard inserts Sentry code correctly. Let's maybe start with SvelteKit and add others later as we see fit.

e2e-tests/tests/sveltekit.test.ts
onurtemizkan: Sure, added 👍

Lms24 (approved):
Thanks for the quick change!

PR #697, opened 2024-10-29
test: Set node version for e2e tests.

onurtemizkan:
Sets node version to `20.18.0` in E2E tests

Note: I'll add a README after the updates on utilities.


Lms24 (approved):
The "runs in prod" test still times out for me locally but I'm probably missing a test setup step. I checked and the node version problems are definitely fixed with this PR. Let's merge this in and I'll follow the readme once it's ready to find out what I'm missing 😅

PR #711, opened 2024-11-14
feat(nuxt): Add Nuxt menu option and install SDK

andreiborza:
Tested locally with a Nuxt app. Will add tests and e2e tests https://github.com/getsentry/sentry-wizard/issues/710.

#skip-changelog

Closes: #706


lforst (dismissed):
In this state, this PR does more bad than good. Let's only merge once the wizard actually does a functioning SDK setup.

Lms24 (approved):
Nice work so far! I ran the wizard and things look good. I think it's okay to merge this into the epic branch.

One nit (feel free to do in this PR or in a separate one): I think this is not on you but can we make the filename in red be coloured in cyan? Just for consistency...
image

andreiborza:
@Lms24 updated to cyan

PR #712, opened 2024-11-15
feat: Pin JS SDK versions to v8

lforst:
This PR pins the installed SDK versions to v8 so we can safely release v9 without running into any compatibility issues.

We are adding a label option to the installPackage util to prevent any ugly log messages.

### DIFF
--- a/src/nextjs/nextjs-wizard.ts
+++ b/src/nextjs/nextjs-wizard.ts
@@ -94,6 +94,7 @@ export async function runNextjsWizardWithTelemetry(
   const { packageManager: packageManagerFromInstallStep } =
     await installPackage({
       packageName: '@sentry/nextjs@^8',
+      packageNameDisplayLabel: '@sentry/nextjs',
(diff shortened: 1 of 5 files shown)


lforst:
@andreiborza Reminder to maybe also do this for Nuxt/Solid when this is merged.

.prettierrc
```diff
@@ -1,6 +1,14 @@
...
+  "trailingComma": "all",
+  "overrides": [
+    {
+      "files": "CHANGELOG.md",
+      "options": {
+        "proseWrap": "preserve"
+      }
+    }
+  ]
+}
```
Lms24, line 14: just to confirm: Is this on purpose? No objections btw

Lms24 (approved):

.prettierrc
lforst: Yeah, I got annoyed by prettier fully re-formatting the changelog when formatting-on-save. This causes there to not be any changes anymore!

PR #713, opened 2024-11-15
feat(nuxt): Add sdk module to nuxt config and create client and server config files

andreiborza:
Tested locally on a nuxt app with various configs. The nuxt default config is taken directly from the latest nuxt version when creating a basic project using their wizard.

#skip-changelog

Closes: #707 and #708

### DIFF
--- a/src/nuxt/nuxt-wizard.ts
+++ b/src/nuxt/nuxt-wizard.ts
@@ -3,7 +3,7 @@ import * as clack from '@clack/prompts';
 import * as Sentry from '@sentry/node';
 import { lt, minVersion } from 'semver';
 import type { WizardOptions } from '../utils/types';
-import { withTelemetry } from '../telemetry';
+import { traceStep, withTelemetry } from '../telemetry';
 import {
   abort,
   abortIfCancelled,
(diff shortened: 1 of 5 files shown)


src/nuxt/nuxt-wizard.ts
```diff
@@ -85,4 +84,16 @@ export async function runNuxtWizardWithTelemetry(
...
+
+  await traceStep('configure-sdk', async () => {
+    await addSDKModule(nuxtConfig, {
+      org: selectedProject.organization.slug,
+      project: selectedProject.slug,
+      url: selfHosted ? sentryUrl : undefined,
+    });
+
+    await createConfigFiles(selectedProject.keys[0].dsn.public);
+  });
```
andreiborza, line 98: q: I handle error handling within the respective helpers. Would it be better to throw and wrap `traceStep` in a try catch?

src/nuxt/sdk-setup.ts
```diff
@@ -0,0 +1,175 @@
...
+          }),
+        );
+        Sentry.setTag(
+          `overwrite-${configVariant}-config`,
+          overwriteExistingConfigs,
+        );
+
+        shouldWriteFile = overwriteExistingConfigs;
+
+        if (overwriteExistingConfigs) {
```
lforst, line 148: If the user selects "no" when asked to overwrite, we should still show them what they would need to put in those files.

lforst (approved):

src/nuxt/sdk-setup.ts
andreiborza: 👍 updated

Issue #705, opened 2024-11-14
Support Nuxt SDK Setup


andreiborza:
The Sentry Wizard should support setting up the Nuxt SDK.

PR #719, opened 2024-11-25
feat(nuxt): Add nuxt wizard

andreiborza:
Closes: #705


src/nuxt/sdk-setup.ts
```diff
@@ -0,0 +1,209 @@
...
+    // so we fall back to showing how to configure the nuxt config
+    // manually.
+    if (e instanceof MagicastError) {
+      clack.log.warn(
+        `Automatic configuration of ${chalk.cyan(
+          path.basename(config),
+        )} failed, please add the following settings:`,
+      );
+      // eslint-disable-next-line no-console
+      console.log(`\n\n${getNuxtModuleFallbackTemplate(options)}\n\n`);
```
Lms24, line 93: l (feel free to disregard if N/A): We have a diff-like helper function for displaying code snippets: `makeCodeSnippet` in `clack-utils.ts`. Might be useful here but I'll let you make the call if it's necessary.

src/nuxt/sdk-setup.ts
```diff
@@ -0,0 +1,209 @@
...
+      }
+
+      if (shouldWriteFile) {
+        await fs.promises.writeFile(
+          path.join(process.cwd(), typeScriptDetected ? tsConfig : jsConfig),
+          getSentryConfigContents(dsn, configVariant, selectedFeatures),
+          { encoding: 'utf8', flag: 'w' },
+        );
+        clack.log.success(
+          `Created fresh ${chalk.cyan(
```
Lms24, line 189: l/personal preference: 

```suggestion
          `Created new ${chalk.cyan(
```

Lms24 (approved):
I used the wizard in an OSS nuxt project and it worked well! The steps seem fine to me, wording is great! :shipit:

src/nuxt/sdk-setup.ts
andreiborza: Thanks, I'll disregard for now to get this out, but will revisit.

e2e-tests/test-applications/nuxt-4-test-app
onurtemizkan, line 1: @andreiborza, this test app is added here as a submodule. I think that's why the Nuxt tests are broken on master.

andreiborza: Thanks, nice catch @onurtemizkan, I opened up a pr to fix it: https://github.com/getsentry/sentry-wizard/pull/742

PR #720, opened 2024-11-26
chore(Android): Removes profiling from android manifest template

kahest:
Removes Profiling from Android Manifest template. Related: https://github.com/getsentry/sentry-docs/pull/11905


bruno-garcia (approved):
Good catch!

PR #725, opened 2024-11-28
feat(nuxt): Add downgrade path to nitro 2.9.7

andreiborza:
Screenshot 2024-11-29 at 10 39 22@2x

See: https://github.com/getsentry/sentry-javascript/issues/14514


AbhiPrasad (approved):

src/utils/clack-utils.ts
```diff
@@ -1469,3 +1491,18 @@ export async function featureSelectionPrompt<F extends ReadonlyArray<Feature>>(
     return selectedFeatures as { [key in F[number]['id']]: boolean };
   });
 }
+
+export async function askShouldAddNuxtOverride(
```
Lms24, line 1495: l: given this was added to clack-utils, let's try to get rid of the nuxt-specific part.

```suggestion
export async function askShouldAddPackageOverride(
```

src/utils/clack-utils.ts
```diff
@@ -1469,3 +1491,18 @@ export async function featureSelectionPrompt<F extends ReadonlyArray<Feature>>(
...
+  pkgName: string,
+  pkgVersion: string,
+): Promise<boolean> {
+  return traceStep(`ask-add-nuxt-overrides`, () =>
```
Lms24, line 1499:
```suggestion
  return traceStep(`ask-add-override`, () =>
```

Lms24 (approved):

src/utils/clack-utils.ts
andreiborza: Right, updated!

src/utils/clack-utils.ts
andreiborza: Named it `ask-add-package-override`

Issue #726, opened 2024-11-29
Install iitm when using pnpm


andreiborza:
Users might run into https://docs.sentry.io/platforms/javascript/guides/nuxt/troubleshooting/#pnpm-resolving-import-in-the-middle-external-package-errors when using pnpm.

Detect and install iitm when the package manager is pnpm.

PR #727, opened 2024-12-02
feat(nuxt): Add import-in-the-middle install step when using pnpm

andreiborza:
Waiting on: https://github.com/getsentry/sentry-docs/pull/12002 to update the trouble shooting url.

Closes: #726

### DIFF
--- a/src/nuxt/sdk-setup.ts
+++ b/src/nuxt/sdk-setup.ts
@@ -18,12 +18,15 @@ import {
   abort,
   abortIfCancelled,
   askShouldAddPackageOverride,
+  askShouldInstallPackage,
   featureSelectionPrompt,
+  installPackage,
   isUsingTypeScript,
 } from '../utils/clack-utils';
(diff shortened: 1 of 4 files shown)


src/nuxt/sdk-setup.ts
```diff
@@ -245,4 +259,26 @@ export async function addNuxtOverrides(
       await packageManager.addOverride(pkgName, pkgVersion);
     }
   }
+
+  if (PNPM.detect()) {
+    // For pnpm, we probably want to install iitm
```
Lms24, line 264: l: probably? 😅 
```suggestion
    // For pnpm, we probably want to install iitm
```

Lms24 (approved):

src/nuxt/sdk-setup.ts
andreiborza: We can probably remove that :)

Issue #672, opened 2024-09-12
Add Wizard for Angular


mydea:
We want to add a wizard flow for Angular.

```[tasklist]
### Tasks
- [x] Add menu option and install SDK - #765
- [x] Add Sentry setup in `main.ts` - https://github.com/getsentry/sentry-wizard/pull/768
- [x] Add Sentry setup in App Config - https://github.com/getsentry/sentry-wizard/pull/769
- [x] configure source map upload - https://github.com/getsentry/sentry-wizard/pull/770
- [x] Add e2e test app(s) - https://github.com/getsentry/sentry-wizard/pull/780
- [ ] https://github.com/getsentry/sentry-wizard/pull/953
- [ ] https://github.com/getsentry/sentry-wizard/pull/963
- [ ] https://github.com/getsentry/sentry-docs/pull/13472
```

PR #741, opened 2024-12-12
feat: Add Angular Wizard

onurtemizkan:
Resolves: https://github.com/getsentry/sentry-wizard/issues/672

Adds wizard for Angular projects

Notes:
- Does not support projects using NGModules, as they are not enabled by default anymore
- Uses the internal sourcemaps wizard for Sentry CLI configuration
- Does not create example pages, we can add that in a follow-up PR.

### DIFF
--- a/src/angular/angular-wizard.ts
+++ b/src/angular/angular-wizard.ts
@@ -7,6 +7,7 @@ import chalk from 'chalk';
 import type { WizardOptions } from '../utils/types';
 import { traceStep, withTelemetry } from '../telemetry';
 import {
+  abortIfCancelled,
   confirmContinueIfNoOrDirtyGitRepo,
   ensurePackageIsInstalled,
   featureSelectionPrompt,
@@ -14,12 +15,17 @@ import {
   getPackageDotJson,
   installPackage,
   printWelcome,
+  runPrettierIfInstalled,
 } from '../utils/clack-utils';
 import { getPackageVersion, hasPackageInstalled } from '../utils/package-json';
-import { gte, minVersion } from 'semver';
-import { initalizeSentryOnAppModule, updateAppConfig } from './sdk-setup';
+import { gte, minVersion, SemVer } from 'semver';
+import {
+  initalizeSentryOnApplicationEntry,
+  updateAppConfig,
+} from './sdk-setup';
 import { addSourcemapEntryToAngularJSON } from './codemods/sourcemaps';
 import { runSourcemapsWizard } from '../sourcemaps/sourcemaps-wizard';
+import * as Sentry from '@sentry/node';
 
 const MIN_SUPPORTED_ANGULAR_VERSION = '14.0.0';
 
@@ -38,7 +44,7 @@ async function runAngularWizardWithTelemetry(
   options: WizardOptions,
 ): Promise<void> {
   printWelcome({
-    wizardName: 'Sentry Remix Wizard',
+    wizardName: 'Sentry Angular Wizard',
     promoCode: options.promoCode,
     telemetryEnabled: options.telemetryEnabled,
   });
@@ -49,24 +55,28 @@ async function runAngularWizardWithTelemetry(
 
   await ensurePackageIsInstalled(packageJson, '@angular/core', 'Angular');
 
-  const installedAngularVersion = getPackageVersion(
-    '@angular/core',
-    packageJson,
-  );
+  let installedAngularVersion = getPackageVersion('@angular/core', packageJson);
 
   if (!installedAngularVersion) {
     clack.log.warn('Could not determine installed Angular version.');
 
-    return;
+    installedAngularVersion = await abortIfCancelled(
+      clack.text({
+        message: 'Please enter the installed Angular version:',
+        validate(value) {
+          if (!value) {
+            return 'Please enter the installed Angular version.';
+          }
+
+          if (!minVersion(value)) {
+            return `Invalid Angular version provided: ${value}`;
+          }
+        },
+      }),
+    );
   }
 
-  const installedMinVersion = minVersion(installedAngularVersion);
-
-  if (!installedMinVersion) {
-    clack.log.warn('Could not determine minimum Angular version.');
-
-    return;
-  }
+  const installedMinVersion = minVersion(installedAngularVersion) as SemVer;
 
   const isSupportedAngularVersion = gte(
     installedMinVersion,
@@ -77,6 +87,11 @@ async function runAngularWizardWithTelemetry(
     clack.log.warn(
       `Angular version ${MIN_SUPPORTED_ANGULAR_VERSION} or higher is required.`,
     );
+    clack.log.warn(
+      `Please refer to Sentry's version compatibility table for more information: ${chalk.underline(
+        'https://docs.sentry.io/platforms/javascript/guides/angular/#angular-version-compatibility',
+      )}`,
+    );
 
     return;
   }
@@ -84,10 +99,17 @@ async function runAngularWizardWithTelemetry(
   const { selectedProject, authToken, sentryUrl, selfHosted } =
     await getOrAskForProjectData(options, 'javascript-angular');
 
+  const sdkAlreadyInstalled = hasPackageInstalled(
+    '@sentry/angular',
+    packageJson,
+  );
+
+  Sentry.setTag('sdk-already-installed', sdkAlreadyInstalled);
+
   await installPackage({
     packageName: '@sentry/angular@^8',
     packageNameDisplayLabel: '@sentry/angular',
-    alreadyInstalled: hasPackageInstalled('@sentry/angular', packageJson),
+    alreadyInstalled: sdkAlreadyInstalled,
   });
 
   const dsn = selectedProject.keys[0].dsn.public;
@@ -109,16 +131,19 @@ async function runAngularWizardWithTelemetry(
     },
   ] as const);
 
-  await traceStep('Inject Sentry to Angular app config', async () => {
-    await initalizeSentryOnAppModule(dsn, selectedFeatures);
-  });
+  await traceStep(
+    'Initialize Sentry on Angular application entry point',
+    async () => {
+      await initalizeSentryOnApplicationEntry(dsn, selectedFeatures);
+    },
+  );
 
   await traceStep('Update Angular project configuration', async () => {
     await updateAppConfig(installedMinVersion, selectedFeatures.performance);
   });
 
   await traceStep('Setup for sourcemap uploads', async () => {
-    addSourcemapEntryToAngularJSON();
+    await addSourcemapEntryToAngularJSON();
 
     if (!options.preSelectedProject) {
       options.preSelectedProject = {
@@ -148,7 +173,13 @@ async function runAngularWizardWithTelemetry(
     await runSourcemapsWizard(options, 'angular');
   });
 
-  clack.log.success(
-    'Sentry has been successfully configured for your Angular project',
+  await traceStep('Run Prettier', async () => {
+    await runPrettierIfInstalled();
+  });
+
+  clack.outro(`
+    ${chalk.green(
+    'Sentry has been successfully configured for your Angular project.',
+  )}`
   );
 }
--- a/src/angular/angular-wizard.ts
+++ b/src/angular/angular-wizard.ts
@@ -62,13 +62,19 @@ async function runAngularWizardWithTelemetry(
 
     installedAngularVersion = await abortIfCancelled(
       clack.text({
-        message: 'Please enter the installed Angular version: ',
+        message: `Please enter your installed Angular major version (e.g. ${chalk.cyan(
+          '18',
+        )} for Angular 18)`,
         validate(value) {
           if (!value) {
-            return 'Please enter the installed Angular version.';
+            return 'Angular version is required';
           }
 
-          if (!minVersion(value)) {
+          try {
+            if (!minVersion(value)) {
+              return `Invalid Angular version provided: ${value}`;
+            }
+          } catch (error) {
             return `Invalid Angular version provided: ${value}`;
           }
         },
--- a/src/angular/angular-wizard.ts
+++ b/src/angular/angular-wizard.ts
@@ -1,0 +1,154 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+
+// @ts-expect-error - clack is ESM and TS complains about that. It works though
+import clack from '@clack/prompts';
+
+import chalk from 'chalk';
+import type { WizardOptions } from '../utils/types';
+import { traceStep, withTelemetry } from '../telemetry';
+import {
+  confirmContinueIfNoOrDirtyGitRepo,
+  ensurePackageIsInstalled,
+  featureSelectionPrompt,
+  getOrAskForProjectData,
+  getPackageDotJson,
+  installPackage,
+  printWelcome,
+} from '../utils/clack-utils';
+import { getPackageVersion, hasPackageInstalled } from '../utils/package-json';
+import { gte, minVersion } from 'semver';
+import { initalizeSentryOnAppModule, updateAppConfig } from './sdk-setup';
+import { addSourcemapEntryToAngularJSON } from './codemods/sourcemaps';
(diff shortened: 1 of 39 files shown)


src/angular/codemods/app-config.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  if (!hasProvideAppInitializer) {
+    imports.$add({
+      from: '@angular/core',
+      imported: 'provideAppInitializer',
+      local: 'provideAppInitializer',
+    });
+  }
+
+  const hasInject = imports.$items.some((item) => item.local === 'inject');
```
andreiborza, line 90: m: and here, can we narrow down?

src/angular/codemods/app-config.ts
```diff
@@ -0,0 +1,232 @@
...
+      local: 'inject',
+    });
+  }
+}
+
+function addAppInitializer(originalAppConfigMod: ProxifiedModule<any>): void {
+  const imports = originalAppConfigMod.imports;
+
+  const hasAppInitializer = imports.$items.some(
+    (item) => item.local === 'APP_INITIALIZER',
```
andreiborza, line 105: l: could also narrow here, although might be specific enough already.

src/angular/angular-wizard.ts
```diff
@@ -0,0 +1,154 @@
...
+  const installedAngularVersion = getPackageVersion(
+    '@angular/core',
+    packageJson,
+  );
+
+  if (!installedAngularVersion) {
+    clack.log.warn('Could not determine installed Angular version.');
+
+    return;
+  }
```
Lms24, line 61: m: Given that this might be our fault rather than a user-misconfiguration or similar, can we fall back to asking users for the version?

src/angular/angular-wizard.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  const installedMinVersion = minVersion(installedAngularVersion);
+
+  if (!installedMinVersion) {
+    clack.log.warn('Could not determine minimum Angular version.');
+
+    return;
+  }
```
Lms24, line 69: Given the above, we shouldn't fall into this case, so we might as well remove it

src/angular/sdk-setup.ts
```diff
@@ -0,0 +1,107 @@
...
+    return;
+  }
+
+  const updatedAppModuleMod = updateAppModuleMod(
+    originalAppModule,
+    dsn,
+    selectedFeatures,
+  );
+
+  await writeFile(updatedAppModuleMod.$ast, appModulePath);
```
andreiborza, line 58: m: I think we need a try/catch around this and gracefully abort the wizard. A single typo in user's config would crash this.

src/angular/angular-wizard.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  const isSupportedAngularVersion = gte(
+    installedMinVersion,
+    MIN_SUPPORTED_ANGULAR_VERSION,
+  );
+
+  if (!isSupportedAngularVersion) {
+    clack.log.warn(
+      `Angular version ${MIN_SUPPORTED_ANGULAR_VERSION} or higher is required.`,
+    );
```
Lms24, line 79: l: Let's print the link to our version compatibility table in the docs (https://docs.sentry.io/platforms/javascript/guides/angular/#angular-version-compatibility) and tell them to find the suitable SDK version instead.

src/angular/sdk-setup.ts
```diff
@@ -0,0 +1,107 @@
...
+      appConfig,
+      angularVersion,
+      isTracingEnabled,
+    );
+
+    await writeFile(updatedAppConfigMod.$ast, appConfigPath);
+  } catch (error) {
+    clack.log.error(
+      `Failed to update your app config ${chalk.cyan(appConfigFilename)}`,
+    );
```
andreiborza, line 97: l/m: In the nuxt wizard, I specifically checked for `MagicastError` here and logged out what users need to do to their config to make it work. Maybe something to consider here too if possible?

src/angular/angular-wizard.ts
```diff
@@ -0,0 +1,154 @@
...
+      prompt: `Do you want to enable ${chalk.bold(
+        'Sentry Session Replay',
+      )} to get a video-like reproduction of errors during a user session?`,
+      enabledHint: 'recommended, but increases bundle size',
+    },
+  ] as const);
+
+  await traceStep('Inject Sentry to Angular app config', async () => {
+    await initalizeSentryOnAppModule(dsn, selectedFeatures);
+  });
```
Lms24, line 114: I think what we're doing here is not related to the "app config" or "app module" at all, but we rather init Sentry in `main.ts` right? I'd rather call the the step and function something like app entry point or so. The "app module" makes this sound like it modifies `app.module.ts`.

src/angular/angular-wizard.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  await traceStep('Update Angular project configuration', async () => {
+    await updateAppConfig(installedMinVersion, selectedFeatures.performance);
+  });
+
+  await traceStep('Setup for sourcemap uploads', async () => {
+    addSourcemapEntryToAngularJSON();
```
Lms24, line 121: m: Let's log a message that we modified `angular.json`. I completely missed this when running the wizard (also related to my comment below).

src/angular/angular-wizard.ts
```diff
@@ -0,0 +1,154 @@
...
+              },
+            },
+          ],
+        },
+      };
+
+      options.url = sentryUrl;
+    }
+
+    await runSourcemapsWizard(options, 'angular');
```
Lms24, line 148: l: Generally, running the source maps wizard here seems fine to me. I found the integration quite seamless. However, in the source maps wizard, there's a step where I need to verify that source maps are created. I think, given that within the wizard run, already enabled source maps emission, we might as well just skip this step. This probably requires some kind of new option in the source maps wizard. Feel free to add that as you see fit.

src/angular/angular-wizard.ts
```diff
@@ -0,0 +1,154 @@
...
+  });
+
```
Lms24, line 150: m: Before we print the outro, let's call `runPrettierIfInstalled()` to optionally reformat the project files.

src/angular/codemods/app-config.ts
```diff
@@ -0,0 +1,232 @@
...
+    (item) => item.local === 'ErrorHandler',
+  );
+
+  if (!hasErrorHandler) {
+    imports.$add({
+      from: '@angular/core',
+      imported: 'ErrorHandler',
+      local: 'ErrorHandler',
+    });
+  }
```
Lms24, line 57: m: If they already have their own ErrorHandler, users need to combine theirs with ours. Let's print a link to our docs on how to do that. https://docs.sentry.io/platforms/javascript/guides/angular/features/error-handler/#multiple-error-handlers

src/angular/codemods/sourcemaps.ts
```diff
@@ -0,0 +1,45 @@
...
+export function addSourcemapEntryToAngularJSON(): void {
+  const angularJsonPath = path.join(process.cwd(), 'angular.json');
+
+  const angularJSONFile = fs.readFileSync(angularJsonPath, 'utf-8');
+
+  const angularJson = JSON.parse(angularJSONFile);
+
+  if (!angularJson) {
+    throw new Error('Could not find in angular.json in your project');
+  }
```
Lms24, line 15: l: If we end up here, let's provide users a fallback, which in this case is to manually enable source maps upload. 2 ideas: 

1. we link to docs
2. we show the snippet that we show in the source maps wizard if there's no pre-selected project

Lms24 (commented):
Thanks Onur! I left some comments that we should address. One thing we still need to handle is to at least abort gracefully if we detect an NGModule-based project. Right now, I don't think the wizard would handle this very well. What we can do is to heuristically search for an `app.module.ts` file. The file name is just a convention though (users could name this file however they like) but probably a good enough one.

My gut feeling tells me that a big part (if not the majority) of Angular users will still use NGModules so we should think about finding a way to support these as well. We can take care of this in a follow-up PR though. For now, let's focus on printing good fallback instructions. 

Also, let's add a if we encounter a module- or app-config-based setup so that we can gather some data if it's worth investing time into handling NGModules better.

The general path through the wizard should be to always show fallback instructions (link to docs, in-wizard code snippets) of how users can manually do what the wizard fails to do. 

So from my PoV let's do the following in this PR:
- address all review comments

For us to follow up in future PRs (who does it TBD):
- example page
- handle NGModules

src/angular/codemods/app-config.ts
```diff
@@ -0,0 +1,232 @@
...
+          b.objectProperty(
+            b.identifier('useValue'),
+            b.identifier('Sentry.createErrorHandler()'),
+          ),
+        ]);
+
+        providers.elements.push(
+          // @ts-expect-error - errorHandlerObject is an objectExpression
+          errorHandlerObject,
+        );
```
Lms24, line 167: m: Angular only permits one ErrorHandler per application. We need to check if users already declared one and if so, let's not add an additional one but link to our docs how to combine their handler with ours: https://docs.sentry.io/platforms/javascript/guides/angular/features/error-handler/

Lms24:
Just FYI: 
image

@andreiborza and I commented on the same line and discussed offline with which suggestion to go to. I just removed the comments to avoid confusion :)

src/angular/angular-wizard.ts
```diff
@@ -49,24 +55,28 @@ async function runAngularWizardWithTelemetry(
...
   if (!installedAngularVersion) {
     clack.log.warn('Could not determine installed Angular version.');
 
-    return;
+    installedAngularVersion = await abortIfCancelled(
+      clack.text({
+        message: 'Please enter the installed Angular version:',
+        validate(value) {
+          if (!value) {
+            return 'Please enter the installed Angular version.';
```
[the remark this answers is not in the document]
Lms24: The major is enough though, right? Finding the specific minor/patch might be tricky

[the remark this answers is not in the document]
andreiborza: Yea, if we're fine with major that's ok too but I think the point still stands that I'd want some guideline on what kind of answer you expect from me as a user. So we could just say (e.g. chalk.cyan(18) for version 18)? Something like that.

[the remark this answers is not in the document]
Lms24: yup, sounds good!

onurtemizkan:
Thanks for the reviews, @andreiborza, @Lms24!
All the points should be addressed.

src/angular/angular-wizard.ts
```diff
@@ -0,0 +1,190 @@
...
+  const installedMinVersion = minVersion(installedAngularVersion) as SemVer;
+
+  const isSupportedAngularVersion = gte(
+    installedMinVersion,
+    MIN_SUPPORTED_ANGULAR_VERSION,
+  );
+
+  if (!isSupportedAngularVersion) {
+    clack.log.warn(
+      `Angular version ${MIN_SUPPORTED_ANGULAR_VERSION} or higher is required.`,
```
andreiborza, line 94:
```suggestion
      `Angular version ${chalk.cyan(MIN_SUPPORTED_ANGULAR_VERSION)} or higher is required.`,
```

src/angular/codemods/app-config.ts
```diff
@@ -0,0 +1,259 @@
...
+                (prop.value as Identifier).name === 'ErrorHandler',
+            ),
+        );
+
+        // If there is already an ErrorHandler provider, we skip adding it and log a message
+        if (hasErrorHandlerProvider) {
+          clack.log
+            .warn(`ErrorHandler provider already exists in your app config.
+Please refer to the Sentry Angular SDK documentation to combine it manually with Sentry's ErrorHandler.
+https://docs.sentry.io/platforms/javascript/guides/angular/features/error-handler/
```
andreiborza, line 176: l: Let's underline the docs link like the other links.

src/angular/codemods/sourcemaps.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    projectConfig.architect.build.configurations.production = {
+      sourceMap: true,
+    };
+  }
+
+  try {
+    fs.writeFileSync(angularJsonPath, JSON.stringify(angularJson, null, 2));
+  } catch (error) {
+    clack.log.error(`Failed to write sourcemap configuration to angular.json`);
```
andreiborza, line 49: m: What should users do in this case? Can we show them a pasteable snippet of what they could do instead?

src/angular/sdk-setup.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  if (hasSentryContent(appEntryPath, originalAppEntry.$code)) {
+    return;
+  }
+
+  const updatedAppEntryMod = updateAppEntryMod(
+    originalAppEntry,
+    dsn,
+    selectedFeatures,
+  );
```
andreiborza, line 57: m: Could this not already fail? How about moving that down into the `try`?

src/angular/sdk-setup.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  try {
+    await writeFile(updatedAppEntryMod.$ast, appEntryPath);
+  } catch (error: unknown) {
+    if (error instanceof MagicastError) {
```
andreiborza, line 62: m: In the nuxt wizard I went back on specifically checking for MagicastErrors because really any error warrants instructing users how to possibly get around it. See https://github.com/getsentry/sentry-wizard/blob/master/src/nuxt/sdk-setup.ts#L122-L149.

src/angular/sdk-setup.ts
```diff
@@ -0,0 +1,150 @@
...
+        chalk.dim(
+          typeof error === 'object' && error != null && 'toString' in error
+            ? error.toString()
+            : typeof error === 'string'
+            ? error
+            : '',
+        ),
+      );
+
+      await abort('Exiting Wizard');
```
andreiborza, line 83: m: I think we don't want to abort here if we instruct users on what to do. cc @Lms24

src/angular/sdk-setup.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    await writeFile(updatedAppConfigMod.$ast, appConfigPath);
+  } catch (error: unknown) {
+    if (error instanceof MagicastError) {
+      clack.log.warn(
+        `Failed to update your app config ${chalk.cyan(
+          appConfigFilename,
+        )} automatically.
+Please refer to the documentation for manual setup
+https://docs.sentry.io/platforms/javascript/guides/angular/#configure`,
```
andreiborza, line 125: m: Same comment as above, I think we should refrain from checking the error type here and always log out alternatives.

src/angular/sdk-setup.ts
```diff
@@ -0,0 +1,150 @@
...
+        chalk.dim(
+          typeof error === 'object' && error != null && 'toString' in error
+            ? error.toString()
+            : typeof error === 'string'
+            ? error
+            : '',
+        ),
+      );
+
+      await abort('Exiting Wizard');
```
andreiborza, line 143: m: Not sure if we want to abort here either.

andreiborza:
Thanks @onurtemizkan, I added my new review. Mostly LGTM, just some clarifications.

src/angular/sdk-setup.ts
Lms24: i think we usually just continue, yes

src/sourcemaps/sourcemaps-wizard.ts
```diff
@@ -220,7 +229,11 @@ async function startToolSetupFlow(
     case 'angular':
       await configureSentryCLI(
         options,
-        configureAngularSourcemapGenerationFlow,
+        // Angular wizard handles the angular.json setup itself
+        !preSelectedTool || preSelectedTool !== 'angular'
+          ? configureAngularSourcemapGenerationFlow
+          : // Not leaving this as undefined to avoid default. This is expected to be a no-op.
+            async () => Promise.resolve(),
```
Lms24, line 236: I see there's a new paramter (`skipValidation`). Should we use it here instead of creating the noop function?

Issue #739, opened 2024-12-11
Add deployment platform inquiry and configure `autoInjectServerSentry` accordingly


andreiborza:
For platforms that cannot make use of `--import` (atm Vercel and Netlify), we need to configure `autoInjectServerSentry` to `top-level-import` and direct users to docs.

PR #747, opened 2024-12-18
feat(nuxt): Add deployment-platform flow with links to docs

andreiborza:
We now ask where users plan to deploy their app:
Screenshot 2024-12-18 at 17 01 15@2x

For Vercel and Netlify we configure top level import of the Sentry server config file and link to docs:
Screenshot 2024-12-18 at 17 01 27@2x

For other/none we instruct people at the end to `--import` their Sentry server config file and link to docs:
Screenshot 2024-12-18 at 17 02 50@2x

This also improves the handling when a user's nuxt config can not be read. Instead of exiting the wizard, we continue and instruct people how to manually add it (already existed, but we bailed immediately after instructing).

Closes: #739

### DIFF
--- a/src/nuxt/nuxt-wizard.ts
+++ b/src/nuxt/nuxt-wizard.ts
@@ -26,6 +26,7 @@ import {
   getNuxtConfig,
   createConfigFiles,
   addNuxtOverrides,
+  askDeploymentPlatform,
 } from './sdk-setup';
 import {
   createExampleComponent,
@@ -33,6 +34,7 @@ import {
   supportsExamplePage,
 } from './sdk-example';
 import { isNuxtV4 } from './utils';
+import { DeploymentPlatform } from './types';
 
 export function runNuxtWizard(options: WizardOptions) {
   return withTelemetry(
@@ -116,8 +118,10 @@ export async function runNuxtWizardWithTelemetry(
     selfHosted,
   };
 
+  const deploymentPlatform = await askDeploymentPlatform();
+
   await traceStep('configure-sdk', async () => {
-    await addSDKModule(nuxtConfig, projectData);
+    await addSDKModule(nuxtConfig, projectData, deploymentPlatform);
     await createConfigFiles(selectedProject.keys[0].dsn.public);
   });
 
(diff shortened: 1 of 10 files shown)


s1gr1d (commented):
I checked the general flow of the wizard and I think it's good 👍

src/nuxt/nuxt-wizard.ts
```diff
@@ -149,16 +153,32 @@ export async function runNuxtWizardWithTelemetry(
...
+
   let msg = chalk.green('\nSuccessfully installed the Sentry Nuxt SDK!');
 
+  if (canImportSentryServerConfigFile) {
+    msg += `\n\nAfter building your Nuxt app, you need to ${chalk.cyan(
+      '--import',
+    )} the Sentry server config file.\n\nFor more info see: ${chalk.cyan(
+      'https://docs.sentry.io/platforms/javascript/guides/nuxt/install/cli-import/#initializing-sentry-with---import',
+    )}`;
+  }
```
Lms24, line 180: l: given that this is a pretty crucial step, I was thinking that we could make this a step that users have to confirm instead of showing this in the outro message. 

So basically, print this message and make them confirm the step with an "Ok, I got it" or so. There's similar precedence for something like this in our [source map wizard](https://github.com/getsentry/sentry-wizard/blob/e6a1fb05cb35c8ddaff84ad29781c6acc1a72b79/src/sourcemaps/tools/angular.ts#L20-L43). 

I'll leave this up to you though, so feel free to go with whatever you prefer.

src/nuxt/nuxt-wizard.ts
```diff
@@ -149,16 +153,32 @@ export async function runNuxtWizardWithTelemetry(
...
 ): string {
+  const canImportSentryServerConfigFile =
+    deploymentPlatform !== 'vercel' && deploymentPlatform !== 'netlify';
+
   let msg = chalk.green('\nSuccessfully installed the Sentry Nuxt SDK!');
 
+  if (canImportSentryServerConfigFile) {
+    msg += `\n\nAfter building your Nuxt app, you need to ${chalk.cyan(
+      '--import',
+    )} the Sentry server config file.\n\nFor more info see: ${chalk.cyan(
```
Lms24, line 177: l: Feel free to go with something else but I was a bit confused what I'd need to do at first. 

```suggestion
    )} the Sentry server config file when running your app.\n\nFor more info see: ${chalk.cyan(
```

Lms24 (approved):
Nice! I like the idea of having this deployment provider section. Maybe something we can think of for more frameworks but it's certainly most relevant for Nuxt (and SolidStart in the future)

src/nuxt/nuxt-wizard.ts
andreiborza: @Lms24 @s1gr1d  how about this?
Screenshot 2024-12-19 at 09 43 26@2x

Lms24: I like it! I think it's most important to have this one extra step so that people actually read what the wizard tells them, before the outro message is printed. Asking about opening docs achieves this well. Thanks!

PR #752, opened 2025-01-07
feat(react-native): Add minimum supported Sentry React Native SDK version detection (>=5.0.0)

krystofwoldrich:
Add warning when running on lower than supported Sentry RN SDK version.

Aborts when canceled or answered no.

This will be helpful when cutting min supported version in the future. Example min 6.X.X.

Example:

Screenshot 2025-01-07 at 15 53 53

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


krystofwoldrich:
@antonis

PR #754, opened 2025-01-09
feat: add support for bun.lock plain text lock file

### LINKED ISSUE #745: Add `bun.lock` to package manager detection
Bun now supports a plain text lockfile: `bun.lock` which we should also check when detecting bun as packagemanager.

See: https://bun.sh/blog/bun-lock-text-lockfile

a-hariti:
Augment package manager detection to support more than one type of lock file to accomodate for Bun's migration to a plain text file lock.

Closes #745

### DIFF
--- a/src/utils/package-manager.ts
+++ b/src/utils/package-manager.ts
@@ -6,10 +6,15 @@ import * as Sentry from '@sentry/node';
 import { traceStep } from '../telemetry';
 import { getPackageDotJson, updatePackageDotJson } from './clack-utils';
 
-export interface PackageManager {
+// some package managers like Bun support multiple lock files (bun.lockb, bun.lock)
+type OneLockFile = string;
+type MultipleLockFiles = string[];
+export type LockFile = OneLockFile | MultipleLockFiles;
+
+export interface PackageManager<Lock = LockFile> {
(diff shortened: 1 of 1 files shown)


src/utils/package-manager.ts
```diff
@@ -6,10 +6,15 @@ import * as Sentry from '@sentry/node';
 import { traceStep } from '../telemetry';
 import { getPackageDotJson, updatePackageDotJson } from './clack-utils';
 
-export interface PackageManager {
+// some package managers like Bun support multiple lock files (bun.lockb, bun.lock)
```
lforst, line 9: I know this wasn't defined in the issue we're resolving with this PR but can I ask you to go with a wild approach instead? 😂

We are not using `lockFile` anywhere so having it (and the generics and stuff) is overengineering. I would just remove the lockfile property and inline everything. Does that make sense?

a-hariti: I thought so as well, 

I was very anxious to going through all the `lockFile` references just find out it's not used anywhere

lforst (approved):
beautiful <3

a-hariti:
finally a net deletion diff 😂

PR #759, opened 2025-01-15
ref(telemetry): Set tunnelroute decision as attribute and tag

Lms24:
also pass the span to the `traceStep` callback so that we can more easily interact with the step span.
#skip-changelog


lforst (approved):

Lms24:
hmm something is up with our e2e tests. I don't think this change could impact Nuxt at all

PR #781, opened 2025-02-06
chore: add coverage upload to Codecov

philprime:
- Adds a Codecov configuration file with some default values. See the `codecov.yml` for comments on the select values.
- Adds the Codecov GitHub-Action to the `job_test` and `job_e2e_test`
- The coverage upload is configured not to fail the CI on error
- The two upload steps are marked as `unit-tests` and `e2e-tests` so they should show up individually in Codecov

#skip-changelog


Lms24 (commented):
Before we merge this, would you mind leaving a line or two in the PR description about what specifically we configured? thanks!

codecov.yml
Lms24, line 1: Will any of this configuration here block PRs or require a certain coverage level?

Lms24:
Are you planning adding the upload of coverage reports in a follow-up PR?

codecov.yml
philprime: I do not know what branch protections are configured for this repository, but if Codecov is not in this list it might show a failed check that coverage is not sufficient. At this point I do not believe it will block RPs

philprime:
@Lms24 thanks for pointing out that the upload of coverage is missing. I'll add it and update the PR description accordingly

codecov.yml
Lms24: sounds good enough for me :D 

> it might show a failed check that coverage is not sufficient

If it comes to this (and it blocks us from merging), I'd say we try to disable this but we'll see. All good for now from my end!

philprime:
@Lms24 git is not doing what I want today, it committed the generated assets from the e2e tests.

Lms24 (approved):
Thanks for making the changes!

PR #785, opened 2025-02-07
feat(apple): add extended whitespace support to AppDelegate detection; add tests for code-tools

philprime:
- Adds tests for the code tools used to insert the setup snippet
- Adds more support for white-spacing in AppDelegate method detection

### DIFF
--- a/src/apple/code-tools.ts
+++ b/src/apple/code-tools.ts
@@ -1,15 +1,15 @@
+import * as Sentry from '@sentry/node';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as templates from './templates';
-import * as Sentry from '@sentry/node';
 // @ts-ignore - clack is ESM and TS complains about that. It works though
 import clack from '@clack/prompts';
 
 const swiftAppLaunchRegex =
-  /(func\s+application\s*\(_\sapplication:[^,]+,\s*didFinishLaunchingWithOptions[^,]+:[^)]+\)\s+->\s+Bool\s+{)|func\s+applicationDidFinishLaunching\(_\s+aNotification:\s+Notification\)\s+{/im;
+  /(func\s+application\s*\(\s*_\s+application:\s*[^,]+,\s*didFinishLaunchingWithOptions[^,]+:\s*[^)]+\s*\)\s+->\s+Bool\s+{)|func\s+applicationDidFinishLaunching\s*\(\s*_\s+aNotification:\s+Notification\s*\)\s*{/im;
 const objcAppLaunchRegex =
-  /-\s*\(BOOL\)\s*application:\s*\(UIApplication\s*\*\)\s*application\s+didFinishLaunchingWithOptions:\s*\(NSDictionary\s*\*\)\s*launchOptions\s*{/im;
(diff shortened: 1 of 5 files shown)


src/apple/code-tools.ts
```diff
@@ -97,6 +97,7 @@ export function addCodeSnippetToProject(
   if (!match) {
     const swiftUIMatch = swiftUIRegex.exec(fileContent);
     if (!swiftUIMatch) {
+      // This branch is not reached, because we already checked for SwiftUI in isAppDelegateFile
```
philipphofmann, line 100: `m`: Why do we need the check then?

test/apple/code-tools.test.ts
```diff
@@ -1,20 +1,1038 @@
...
+
+@implementation AppDelegate
+
+- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+    // Override point for customization after application launch.
+    return YES;
+}
+
+@end`;
+const validAppDelegateObjcWithSentry = `@import Sentry;
```
philipphofmann, line 75: `l`
```suggestion
const validAppDelegateObjCWithSentry = `@import Sentry;
```

test/apple/code-tools.test.ts
```diff
@@ -1,20 +1,1038 @@
...
+                // -- Act --
+                const result = exportForTesting.isAppDelegateFile(filePath);
+
+                // -- Assert --
+                expect(result).toBeTruthy();
+              });
+            });
+          }
+
+          describe('invalid cases', () => {
```
philipphofmann, line 246: `m`: What about adding some cases for garbage content in the files, such as `adsfasdf` or what if there is a missing `}` in the app delegate, so the code doesn't compile.

philipphofmann (commented):
Thanks for the extensive test suite. I added some suggestions to cover a few more edge cases.

src/apple/code-tools.ts
philprime: The compiler requires the null check, I can probably get rid of it, but only by refactoring the implementation. This PR focuses on adding initial test coverage

test/apple/code-tools.test.ts
philprime: Good feedback, it seems I made a copy-paste error here (so many tests 😅). I'll add a garbage example example but we actually don't care about compiling here, because we are not interpreting code, only character matching

philipphofmann: I also don't care too much about compiling here, but I care about the wizard not crashing when somebody has garbage in their AppDelegate 😄

philprime: I agree, but there is no difference in the test equivalence classes, so it would just result in multiple tests testing for the same thing.

I added something for completeness sake, but just wanted to point this out.

philipphofmann: I disagree, the `asdasdf garbage` test is an extra equivalence class when you don't know the code. When you know the code it's not, though. I think it adds value. Thanks.

philipphofmann (approved):
LGTM

PR #786, opened 2025-02-07
test(apple): add unit tests for Fastfile injection

philprime:
- Adds unit tests to inject sentry cli into Fastfile
- Changed the import of `@clack/prompts` to fix mocking

### DIFF
--- a/test/apple/fastfile.test.ts
+++ b/test/apple/fastfile.test.ts
@@ -54,18 +54,35 @@ describe('fastlane', () => {
       }[] = [
         {
           name: 'no platform',
-          content: 'lane :test do\nend',
+          content: `lane :test do
+end`,
           expected: { index: 0, length: 17 },
         },
         {
           name: 'platform is ios',
-          content: 'platform: ios\nlane :test do\nend',
+          content: `platform: ios
+lane :test do
+end`,
           expected: { index: 0, length: 31 },
         },
         {
           name: 'platform is ios and other platform',
-          content: 'platform: ios\nend\nplatform: android\nlane :test do\nend',
-          expected: { index: 0, length: 53 },
+          content: `platform: ios
+end
+platform: android
+  lane :test do
+  end`,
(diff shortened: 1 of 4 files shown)


test/apple/fastfile.test.ts
```diff
@@ -46,4 +44,490 @@ describe('fastlane', () => {
...
+  describe('#findIOSPlatform', () => {
+    describe('platform block detection', () => {
+      const variations: {
+        name: string;
+        content: string;
+        expected: { index: number; length: number };
+      }[] = [
+        {
+          name: 'no platform',
+          content: 'lane :test do\nend',
```
philipphofmann, line 57: `l`: This is a bit hard to read with the inline `\n`. It would be great if we could split this up over multiple lines instead.

test/apple/fastfile.test.ts
```diff
@@ -46,4 +44,490 @@ describe('fastlane', () => {
...
+          expected: { index: 0, length: 17 },
+        },
+        {
+          name: 'platform is ios',
+          content: 'platform: ios\nlane :test do\nend',
+          expected: { index: 0, length: 31 },
+        },
+        {
+          name: 'platform is ios and other platform',
+          content: 'platform: ios\nend\nplatform: android\nlane :test do\nend',
```
philipphofmann, line 67: `m`: I think it would be great to add test cases when the platform uses tabs and spaces before it. So basically a test for something like
```
content: '   platform: ios\nend\nplatform: android\nlane :test do\nend',
```

Then we know the regex also works when somebody has their code formatted with indentation.

philipphofmann (approved):
LGTM, with two minor comments. Thanks for adding all the tests.

test/apple/fastfile.test.ts
philprime: Do we support the case when the indentation is formatted?
Do we in general have any specs of cases we support?

Issue #788, opened 2025-02-12
Installation Issues (NextJS 15)


MadsZBC:
### Environment

I'm using latest sentry SDK and Nextjs 15
### Steps to Reproduce

 `npx @sentry/wizard@latest -i nextjs --saas --org mydsicc-22 --project javascript-nextjs`

### Expected Result

Install Sentry and set it up as it should do, it can be fixed with a --force when doing npm install but i cannot seem to pass that variable to the command. 

### Actual Result
I have attached some logs

[sentry-wizard-installation-error-1739394183600.log](https://github.com/user-attachments/files/18774002/sentry-wizard-installation-error-1739394183600.log)
```
■  Encountered the following error during installation:
│
│  Error: Command failed: npm add @sentry/nextjs@^8
│  npm error code ERESOLVE
│  npm error ERESOLVE could not resolve
│  npm error
│  npm error While resolving: swagger-ui-react@5.18.3
│  npm error Found: react@19.0.0
│  npm error node_modules/react
│  npm error   peer react@">=16.8.0" from @dnd-kit/accessibility@3.1.1
│  npm error   node_modules/@dnd-kit/accessibility
│  npm error     @dnd-kit/accessibility@"^3.1.1" from @dnd-kit/core@6.3.1
│  npm error     node_modules/@dnd-kit/core
│  npm error       peer @dnd-kit/core@"^6.3.0" from @dnd-kit/modifiers@9.0.0
│  npm error       node_modules/@dnd-kit/modifiers
│  npm error         @dnd-kit/modifiers@"^9.0.0" from the root project
│  npm error       2 more (@dnd-kit/sortable, the root project)
│  npm error   peer react@">=16.8.0" from @dnd-kit/core@6.3.1
│  npm error   node_modules/@dnd-kit/core
│  npm error     peer @dnd-kit/core@"^6.3.0" from @dnd-kit/modifiers@9.0.0
│  npm error     node_modules/@dnd-kit/modifiers
│  npm error       @dnd-kit/modifiers@"^9.0.0" from the root project
│  npm error     peer @dnd-kit/core@"^6.3.0" from @dnd-kit/sortable@10.0.0
│  npm error     node_modules/@dnd-kit/sortable
│  npm error       @dnd-kit/sortable@"^10.0.0" from the root project
│  npm error     1 more (the root project)
│  npm error   94 more (@dnd-kit/modifiers, @dnd-kit/sortable, ...)
│  npm error
[...]

PR #791, opened 2025-02-14
feat: Add `forceInstall` option to NPM-based wizards

Lms24:
Adds a CLI argument to make node package managers `--force`-install a package when calling `installPackage`. This is of course only optional and defaults to `false`/`undefined`. 

Added this option to all SDK setup wizards using NPM. Didn't add it to the source maps wizard but we can do so at any time. If reviewers have objections to merging this, I'm also happy to close it in favour of not permitting force installs at all.

closes https://github.com/getsentry/sentry-wizard/issues/788


lforst (approved):
I thought about doing this by default but default but an option is a great escape hatch.

Issue #758, opened 2025-01-15
Provide self-contained binaries for the wizard


BYK:
Currently we tell people to use `brew` or `npx` to run the wizard. A typical Android dev (and many others) don't have either on their system. We should provide self-contained binaries for ease of use.

Spotlight does this: https://github.com/getsentry/spotlight/releases/tag/%40spotlightjs%2Fspotlight%402.9.0

It shouldn't be too hard to port the logic here and I'm working on making this packaging system a separate tool itself anyway.

BYK:
Okay, [fossilize](https://github.com/BYK/fossilize) is ready: https://github.com/getsentry/spotlight/pull/703

Will implement this in this week.

PR #793, opened 2025-02-14
ref!: Upgrade main node version to v18

BYK:
Prerequisite for #758.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- ref: Bump main Node.js version to the earliest LTS v18 ([#793](https://github.com/getsentry/sentry-wizard/pull/793))
```
Lms24, line 5: l: Feel free to ignore but we started marking breaking changes with a `!` in the JS SDK repo recently and I think that's a pretty nice pattern. Just wanna avoid that this gets released as something other than a major release :D 

```suggestion
- ref!: Bump main Node.js version to the earliest LTS v18 ([#793](https://github.com/getsentry/sentry-wizard/pull/793))
```

Lms24 (commented):
If we drop support for Node < 18.20.6, we can also remove 14 and 16 from the test matrix in the build.yml worflow

CHANGELOG.md
BYK: Happy to do that. If we're okay with a major release though, I'd rather jump to Node 20. WDYT?

Lms24 (approved):

PR #796, opened 2025-02-18
ref: Remove lodash, fix type errors, modernize

BYK:
The added forced resolution for `simple-plist` is to fix bundling issues
with this package that is solved in the latest version.

Context: wollardj/simple-plist#58

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
       "json"
     ],
     "modulePathIgnorePatterns": [
-      "<rootDir>/dist/__mocks__"
+      "<rootDir>/dist/"
     ],
     "testPathIgnorePatterns": [
       "/dist/",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "chalk": "^2.4.1",
     "glob": "9.3.5",
     "inquirer": "^6.2.0",
-    "lodash": "^4.17.15",
     "magicast": "^0.2.10",
     "opn": "^5.4.0",
     "r2": "^2.0.1",
@@ -49,7 +48,8 @@
     "@types/inquirer": "^0.0.43",
     "@types/jest": "^29.5.0",
     "@types/lodash": "^4.14.144",
-    "@types/node": "^10.11.0",
+    "@types/node": "^18.19.76",
+    "@types/opn": "5.1.0",
     "@types/rimraf": "^3.0.2",
     "@types/semver": "^7.3.7",
     "@types/yargs": "^16.0.9",
@@ -66,6 +66,9 @@
     "ts-node": "^10.9.1",
     "tsx": "^3.14.0",
     "typescript": "^5.0.4"
+  },
+  "resolutions": {
+    "simple-plist": "1.4.0-0"
   },
   "engines": {
     "node": ">=18.20.6",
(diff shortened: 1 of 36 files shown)


BYK:
I happened to got into this craziness when trying to debug something when trying to get #758 done.

If you think it is not worth it, I'll resign. Otherwise I think it would be good to modernize these things a bit.

package.json
```diff
@@ -67,6 +67,9 @@
     "tsx": "^3.14.0",
     "typescript": "^5.0.4"
   },
+  "resolutions": {
+    "simple-plist": "1.4.0-0"
+  },
```
Lms24, line 72: Is this resolution only relevant for bundling on our end? Asking because it has no effect when users install the wizard and I just wanna make sure that this is not necessary.

tsconfig.build.json
```diff
@@ -5,15 +5,16 @@
     "noImplicitAny": true,
     "strictNullChecks": true,
     "noFallthroughCasesInSwitch": true,
-    "module": "commonjs",
-    "target": "es5",
-    "lib": ["es2015"],
+    "lib": ["ES2022"],
+    "module": "node16",
+    "target": "ES2022",
```
Lms24, line 10: Just to double-check: 
- `"module": "node16"` means that we still transpile to CJS, correct? If we were to transpile to ESM, this would be breaking and we also need to adjust the package.json entry points accordingly.
- `"target": "es2022"`: is es2022 supported by Node 18?

types/r2.d.ts
Lms24, line 1: did we vendor these types from anywhere? If so, we need to copy the license

Lms24 (commented):
Hey, sorry for the long review but I have a few questions/concerns I want to clarify before approving.

The most important question I have is: What's the motivation for these changes? Are all of them required for #758 (or something else you're working on)? I couldn't find anything about this in the PR description so I'd appreciate some more insights. 

- This is a large PR with multiple changes against a code base with poor test coverage. Not your fault obviously but it makes such refactors more risky. So I just wanna make sure these changes are necessary
- All code in `/lib` is only relevant for the legacy electron and cordova wizards. I don't think we should spend much time on refactoring or improving these but rather convert them to the new wizards in `src` instead. This should already solve stuff like depending on lodash or some of the sketchy imports. Ultimately we'll be able to delete a lot of code :) 
- Are the tsconfig language level changes necessary? Besides being potentially breaking (see comment), it requires us to remove all `require` calls (as you did in the PR) and I'm wondering if it is worth the hassle? Not a complete deal breaker though, just wanna make sure this is aligned.

Is there a way we can split this up into more reviewable chunks? I see some high level changes:
- removing outdated dependencies (lodash)
- raising language level
- better types (r2)
So perhaps we can split these up?

package.json
BYK: This is actually bundled in from my fossilize work (#758). It causes a bundling issue and that should be the only change relevant to us. I'll double check the changes but the reports didn't mention anything notable.

tsconfig.build.json
BYK: Took these from https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-18

Don't think it affects CJS/ESM as I need working CJS bundles for #758.

types/r2.d.ts
BYK: I wrote them by hand 😅

BYK:
Closed as I'll be splitting this as suggested.

PR #797, opened 2025-02-19
ref!: Follow up to Node v18 changes

BYK:
Split off from #796

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,13 @@
 # Changelog
 
-
 ## Unreleased
 
 - ref!: Bump main Node.js version to the earliest LTS v18 ([#793](https://github.com/getsentry/sentry-wizard/pull/793))
+- ref!: Follow up to Node v18 changes ([#797](https://github.com/getsentry/sentry-wizard/pull/797))
 
 ## 3.42.0
 
 - feat: Update `nextjs`, `remix`, `sveltekit` and `nuxt` wizards to install v9 ([#794](https://github.com/getsentry/sentry-wizard/pull/794))
-
 
 ## 3.41.0
 
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     "@clack/prompts": "0.7.0",
     "@sentry/cli": "^1.77.3",
     "@sentry/node": "^7.119.2",
-    "@sentry/sveltekit": "^9.1.0",
     "axios": "1.7.4",
     "chalk": "^2.4.1",
     "glob": "9.3.5",
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "@clack/prompts": "0.7.0",
     "@sentry/cli": "^1.77.3",
     "@sentry/node": "^7.119.2",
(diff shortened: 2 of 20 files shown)


CHANGELOG.md
Lms24, line 1: heads-up, this likely needs a rebase from the hotfix I cut via #798

package.json
```diff
@@ -27,6 +27,7 @@
     "@clack/prompts": "0.7.0",
     "@sentry/cli": "^1.77.3",
     "@sentry/node": "^7.119.2",
+    "@sentry/sveltekit": "^9.1.0",
```
Lms24, line 30: any chance this is a leftover from an e2e test run?

src/utils/clack-utils.ts
```diff
@@ -1,32 +1,27 @@
...
+import { Feature, SentryProjectData, WizardOptions } from './types';
 
-export const opn = require('opn') as (
-  url: string,
-  options?: {
-    wait?: boolean;
-  },
-) => Promise<childProcess.ChildProcess>;
-
+export { opn };
```
Lms24, line 24: Since we don't add a type cast to `opn` anymore, can we get rid of this export and import from `'opn'` directly wherever we need it?

tsconfig.build.json
```diff
@@ -5,15 +5,16 @@
...
+    "module": "node16",
+    "target": "ES2022",
     "skipLibCheck": true,
     "types": [],
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
     "sourceMap": true,
     "inlineSources": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
```
Lms24, line 17: l: since we no longer `require` the `package.json` to read the CLI version, can we get rid of `resolveJsonModule` (IIRC it defaults to `false`). 

edit: Actually not sure if we require a json somewhere else, so feel free to ignore

tsconfig.build.json
```diff
@@ -5,15 +5,16 @@
     "noImplicitAny": true,
     "strictNullChecks": true,
     "noFallthroughCasesInSwitch": true,
-    "module": "commonjs",
-    "target": "es5",
-    "lib": ["es2015"],
+    "lib": ["ES2022"],
+    "module": "node16",
+    "target": "ES2022",
```
Lms24, line 10: thanks for pointing out [this link](https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-18) in the original PR! TIL 💡

Lms24 (commented):
Looks like we still need to pull in a couple of type declarations from the original PR to get the build to pass.

package.json
BYK: possible, will revert.

src/utils/clack-utils.ts
BYK: Good idea but I'd do that later?

tsconfig.build.json
BYK: Will do this in a follow up, already planned :)

Lms24 (approved):

PR #799, opened 2025-02-20
ref: Remove obsolete deps (r2, lodash)

BYK:


Lms24 (approved):
no objections other than that the legacy wizards will be replaced anyway long term. But changes LGTM

PR #800, opened 2025-02-20
feat: Update minimum supported Electron version

timfish:
The recently released v6.0.0 of the Electron SDK bumps the minimum supported version of Electron to v23. 

It looks like we didn't update this for the last major version!

#skip-changelog


AbhiPrasad (approved):

timfish:
Are the test failures here relevent?

Issue #561, opened 2024-04-17
Dynamically import `package.json` instead of `JSON.parse`ing it


Lms24:
Alternative issue title: There's no such thing as stability in JS-land

Looks like Bun is going to permit comments in `package.json` (https://github.com/oven-sh/bun/pull/10287). This will break our wizard which currently reads users' `package.json` files and parses them with `JSON.parse`. `JSON.parse` throws when it encounters comments. We catch this in the wizard but this obviously degrades the setup flow (iirc we even early return in some cases, e.g. sourcemaps wizard). 

If they go through with this change, we need to move to an alternative parsing method. I guess we could try to dynamically `import` the file instead but iirc we tried this before and ran into some issues. Opening this issue to track the problem.

Prio-wise, I'd argue this is low because:

1. It's still discouraged to add comments, bun just won't error on it. Still, this opens up a new class of problems
2. Users have to use bun because npm et al. would already error
3. Users have to actually put comments into their package.json

Until we get to this, we'll need to tell folks to _not_ add comments into their `package.json`.

lucas-zimerman:
It's kinda wild that solution was accepted, I'd rather make some plugin for vscode where it parses some `package.json.comment` file and it just applies the comments as a visual effect on package.json, allowing folks to see the comments and also not breaking the JSON standard.

Lms24:
For now I'll close this issue due to priorization. We can re-open when this becomes problematic IRL. If anyone experiences problems please feel free to tag me so that we can reopen this.

PR #801, opened 2025-02-21
ref: No more dynamic requires

BYK:
This patch changes all dynamic requires except for a few, 'safe' cases and 'sentry/cli' related ones as it will be removed in a follow up

### DIFF
--- a/lib/Steps/ChooseIntegration.ts
+++ b/lib/Steps/ChooseIntegration.ts
@@ -12,14 +12,19 @@ import { Cordova } from './Integrations/Cordova';
 import { Electron } from './Integrations/Electron';
 import { hasPackageInstalled } from '../../src/utils/package-json';
 import { dim } from '../Helper/Logging';
+import { readFileSync } from 'node:fs';
 
-let projectPackage: any = {};
+let projectPackage: Record<string, unknown> = {};
 
 try {
   // If we run directly in setup-wizard
-  projectPackage = require('../../package.json');
+  projectPackage = JSON.parse(
+    readFileSync('../../package.json', 'utf-8'),
+  ) as Record<string, unknown>;
 } catch {
-  projectPackage = require(`${process.cwd()}/package.json`);
+  projectPackage = JSON.parse(
+    readFileSync(`${process.cwd()}/package.json`, 'utf-8'),
+  ) as Record<string, unknown>;
 }
 
 type IntegrationPromptAnswer = {
--- a/lib/Steps/OpenSentry.ts
+++ b/lib/Steps/OpenSentry.ts
@@ -1,5 +1,5 @@
+import { URL } from 'node:url';
 import type { Answers } from 'inquirer';
-import { URL } from 'url';
 
 import { mapIntegrationToPlatform } from '../Constants';
 import { BottomBar } from '../Helper/BottomBar';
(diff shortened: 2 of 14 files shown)


lib/Steps/ChooseIntegration.ts
```diff
@@ -12,14 +12,19 @@ import { Cordova } from './Integrations/Cordova';
...
+import { readFileSync } from 'node:fs';
 
-let projectPackage: any = {};
+let projectPackage: Record<string, unknown> = {};
 
 try {
   // If we run directly in setup-wizard
-  projectPackage = require('../../package.json');
+  projectPackage = JSON.parse(
+    readFileSync('../../package.json', 'utf-8'),
```
Lms24, line 22: Funny story about these: `require` is more fault tolerant than `JSON.parse`. Bun decided to permit comments in `package.json` files. Whatever drove them to make this decision is beyond me but here we are. `JSON.parse` will crash on Json with comments. 

I opened #561 a while ago but closed it because it never came up in reality. I'm fine with using JSON.parse for the moment and revisit when people actually report crashes. 

Gotta love the JS ecosystem

lib/Steps/OpenSentry.ts
```diff
@@ -1,13 +1,13 @@
...
-import { URL } from 'url';
 
 import { mapIntegrationToPlatform } from '../Constants';
 import { BottomBar } from '../Helper/BottomBar';
 import { dim, green, l, nl, red } from '../Helper/Logging';
 import { getCurrentIntegration } from '../Helper/Wizard';
 import { BaseStep } from './BaseStep';
 
-const opn = require('opn');
+import opn from 'opn';
```
Lms24, line 10: just to be sure, you tested the old wizard flows (electron/cordova), right? I remember previously running into problems with importing `opn` which was why it was `require`'d before. Might have solved itself with the other language level changes though.  

Also asking because `opn` usage is not tested in our e2e tests as we pass in a pre-selected project to avoid going through the web-based project selection flow which is opened via `opn`

Lms24 (commented):
Had a question about `opn` but otherwise LGTM

lib/Steps/ChooseIntegration.ts
BYK: It is tricky to make dynamic requires work with the static binaries. It is not impossible but is a bit more explicit (you need to use `node:module` and do `makeRequire` with an explicit starting path).

If we get complaints, I can switch to that.

lib/Steps/OpenSentry.ts
BYK: I certainly did not as I have no idea how to test these. I'm just relying on our test suite so if this is something that needs to be tested, then we should test it explicitly. I can do it if you guide me.

BYK: Okay tested `opn` manually with this type of import, just works without issues. I think enabling `esModuleCompat` in `tsconfig` might have been the actual fix for the imports.

Lms24: `yarn try -i electron` should do the trick. If we check that the browser is actually opened when the wizard redirects you to log in, we know that `opn` did what it's supposed to.  

I'm aware that ideally this has an automated test. It is quite complicated to test this though, which is why we opted to work around this step in the e2e tests.

Lms24 (approved):

PR #802, opened 2025-02-21
ref: Remove @sentry/cli as a dependency

BYK:
We no longer require @sentry/cli to exist or to be installed so removing it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -853,7 +853,7 @@ npx @sentry/wizard -s -i reactNative
 
 ## v0.6.1
 
-- Fixed <https://github.com/getsentry/react-native-sentry/issues/304>
+- Fixed https://github.com/getsentry/react-native-sentry/issues/304
 
 ## v0.6.0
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -852,7 +853,7 @@ npx @sentry/wizard -s -i reactNative
 
 ## v0.6.1
 
-- Fixed https://github.com/getsentry/react-native-sentry/issues/304
+- Fixed <https://github.com/getsentry/react-native-sentry/issues/304>
```
Lms24, line 856:
```suggestion
```

auto format?

Lms24 (commented):
We somewhere print the Sentry CLI version when starting the wizard. I think we can get rid of this as well now that it's no longer a hard dependency.

CHANGELOG.md
Lms24, line 856:
```suggestion
- Fixed https://github.com/getsentry/react-native-sentry/issues/304
```

auto-format?

BYK: Yup, should I revert?

BYK: reverted

Lms24 (approved):
Thanks!

PR #806, opened 2025-02-21
feat: Add self-contained binary artifacts

### LINKED ISSUE #758: Provide self-contained binaries for the wizard

BYK:
The added forced resolution for `simple-plist` is to fix bundling issues
with this package that is solved in the latest version.

Context: wollardj/simple-plist#58

Fixes #758.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,8 +133,14 @@ jobs:
           flutter pub get
       - name: Install dependencies with yarn
         run: yarn install --frozen-lockfile
+      - name: Get test binaries
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ github.sha }}
+          path: |
+            ${{ github.workspace }}/dist-bin/
       - name: Run End-to-End Tests
-        run: yarn test:e2e ${{ matrix.wizard }}
+        run: yarn test:e2e:bin ${{ matrix.wizard }}
       - name: Push code coverage to codecov
         uses: codecov/codecov-action@13ce06bfc6bbe3ecf90edbbf1bc32fe5978ca1d3 # pin@v5.3.1
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,9 @@ jobs:
   job_build:
     name: Build
     runs-on: ubuntu-latest
+    env:
+      APPLE_CERT_PATH: /tmp/certs.p12
+      APPLE_API_KEY_PATH: /tmp/apple_key.json
     steps:
       - uses: actions/checkout@v4
       - name: Set up Node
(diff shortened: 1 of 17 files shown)


bin.ts
```diff
@@ -52,85 +52,90 @@ const PRESELECTED_PROJECT_OPTIONS = {
...
+    'force-install': {
+      default: false,
+      describe: 'Force install the SDK NPM package',
+      type: 'boolean',
+    },
+    ...PRESELECTED_PROJECT_OPTIONS,
+  })
+  // This prevents `yargs` from trying to read the local package.json
+  // as it's not available in the Node Single Executable Binary artifacts versions
+  .version(process.env.npm_package_version as string).argv;
```
Lms24, line 138: Is this env variable always set? If so TIL 😅

.github/workflows/build.yml
```diff
@@ -108,8 +133,12 @@ jobs:
           flutter pub get
       - name: Install dependencies with yarn
         run: yarn install --frozen-lockfile
+      - name: Download built binaries from build job
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ github.sha }}
       - name: Run End-to-End Tests
-        run: yarn test:e2e ${{ matrix.wizard }}
+        run: yarn test:e2e:bin ${{ matrix.wizard }}
```
Lms24, line 141: wait, does this mean we only test against the self-contained binaries but no longer with Node (i.e. simulate an `npx @sentry/wizard` run)? If so, let's please do both. Might be sense to do this in a matrix or parallelize it in some way (happy to leave details up to you)

.github/workflows/build.yml
```diff
@@ -21,6 +24,27 @@ jobs:
         run: yarn install --frozen-lockfile
       - name: Build
         run: yarn build
+      - name: Setup Apple Code Signing
```
Lms24, line 27: I see this adds quite a bit of runtime to the build step. Can we split this off into its own job so that other jobs that don't depend on it can already start? 

The nice thing about the wizard is that CI is relatively fast and I'd be happy if it stays fast :D

.github/workflows/build.yml
BYK: We never simulate `npx` properly anyway as the old way is running `node dist/bin.js` which is not entirely the same thing.

The argument here is: fossilized binaries are created from `dist/bin.js` so if fossilized works, `dist/bin.js` should definitely work. The other way around is not guaranteed as Node SEA brings in more restrictions (such as no `require`).

I can create a test matrix but I'd say it will be a waste of time and resources.

bin.ts
BYK: It is always when using `fossilize` or `npx`. Otherwise it may not be set.

.github/workflows/build.yml
BYK: This should not block unit tests but it _rightfully_ should block e2e tests as the tests are run on these binaries. I'll see if there's improvements I can make.

.github/workflows/build.yml
Lms24: got it. I'm fine with this for now, we can revisit when there's a need to test them separately. Thanks for explaining

bin.ts
Lms24: ok, since we fall back to other methods that's fine, thx!

Lms24 (approved):

PR #810, opened 2025-02-21
test(nextjs): Catch type errors in `next.config.ts`

Lms24:
Our NextJS e2e test used a `next.config.mjs` file which means when building the nextjs app, type errors in this file are not caught. This caused the bug fixed via #798 to slip through. So this PR makes the next config a typescript file which will throw a build error if a type mismatch (like adding an unsupported property to the Sentry build options) is detected. 

#skip-changelog

### DIFF
--- a/e2e-tests/test-applications/nextjs-test-app/next.config.mjs
+++ b/e2e-tests/test-applications/nextjs-test-app/next.config.mjs
@@ -1,4 +1,0 @@
-/** @type {import('next').NextConfig} */
-const nextConfig = {};
-
-export default nextConfig;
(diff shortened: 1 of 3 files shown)


AbhiPrasad (approved):

Lms24:
hmm crap, next config TS only works with Next 15. I don't think we want to stop testing against Next 14 though, so we probably need to create a second e2e test app with Next 15. I'll try to slot this in when I have a spare cycle but can't spend too much time on this right now.

Lms24:
superseded by https://github.com/getsentry/sentry-wizard/pull/852

PR #820, opened 2025-02-24
fix: Add dist-bin binaries dir to .npmignore

BYK:
Right now we pack the fossilized binaries to npm package on publish. This patch adds the `dist-bin` directory to `.npmignore` to make this impossible.

Follow up to #806.

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,6 +52,7 @@ jobs:
         with:
           name: ${{ github.sha }}
           if-no-files-found: error
+          compression-level: 3
           path: |
             ${{ github.workspace }}/*.tgz
             ${{ github.workspace }}/dist-bin/*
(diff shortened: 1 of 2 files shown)


Lms24 (approved):
Good catch, thx!

Note for future work: We can use the `files` property in `package.json` to specify what should be in the tarball instead. We switched to this in the JS SDK repo and it's way easier to grasp than an npmignore file.

BYK:
I agree and I was gonna go that path but decided against because:

1. It was a more drastic and potentially risky change for this urgent fix
2. I don't know the code base well enough to trust my intuition on which files to include. I'd just go with dist folder but not sure if that would be enough.

FYI with the single file sentry-wizard.cjs under dist-bin, we can actually remove all dependencies to dev dependencies and just ship that file. 

Downside would be the need for source maps and a way to get d.ts files distributed.

PR #825, opened 2025-02-25
fix: Remove bulk from npm

BYK:
This patch removes further bulk from the npm package which became 183 MB due to `.node-cache` getting bundled in. This is a follow up to #820.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,6 +2,9 @@ assets
 examples
 docs
 dist-bin
+e2e-tests
+test
+scripts
 .idea
 .vscode
 .travis
@@ -9,4 +12,5 @@ dist-bin
 .craft.yml
 .prettierrc
 .eslintrc.js
+.node-cache
 tsconfig*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- fix: Remove bulk from npm ([#825](https://github.com/getsentry/sentry-wizard/pull/825)])
+
 ## 4.0.0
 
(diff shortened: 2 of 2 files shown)


Lms24 (commented):
wdyt about using `files` in the `package.json` instead of `.npmignore` this would prevent stuff like this in the future

package.json
```diff
@@ -22,6 +22,10 @@
   "typescript": {
     "definition": "dist/index.d.ts"
   },
+  "files": [
+    "dist",
+    "COPYING"
```
Lms24, line 27: what's `COPYING`?

Lms24 (commented):
Just had one more question (see comment).

I think this will exclude a lot of files we previously published (see [3.4.0](https://unpkg.com/browse/@sentry/wizard@3.4.0/) as an example) but all of them are source files (+ not referenced from package.json) or stuff like the CHANGELOG which doesn't need to be included.

I also ran it locally and things seemed to work. Happy to 👍 once the comment is addressed

Lms24:
I also checked by running `npm pack` (which is almost the same as `yarn pack` which we use in CI` and it seems to include everything in the tarball 👍

package.json
BYK: No idea, it was in the package that was built so I assumed it was something in the repo. Happy to remove

Lms24:
```suggestion
```

yeah let's remove it then. I don't see this in my local repo and neither in the tarball of 3.x, so I don't think we need it.

package.json
```diff
@@ -22,6 +22,9 @@
   "typescript": {
     "definition": "dist/index.d.ts"
   },
+  "files": [
+    "dist",
```
BYK, line 26:
```suggestion
    "dist"
```

PR #827, opened 2025-02-26
ref: Set wizard version at build time

Lms24:
While checking for regressions captured after the 4.x releases in Sentry, I noticed that we seem to [pick up random, incorrect](https://sentry.sentry.io/releases/?project=4505425820712960&statsPeriod=7d) `release` values instead of the wizard version in our telemetry setup. My best guess is that the previous method of preferring `process.env.npm_package_version` over reading the `package.json` somehow must have given us wrong versions in a couple of instances ( have no proof though, it could have been the package.json lookup as well). 

This likely showed a wrong version number to wizard users a startup and it reported wrong versions to Sentry.

This PR refactors how we set the wizard version by reading the version  from a `version.ts` file at build time and updating this file at release time. Ideally, we could throw in Rollup or some other proper bundler and use a magic string based approach like in the JS SDK repo. For time reasons, I opted to write a small node script which replaces the version file contents with the new version on release. The script is executed by the `craft-pre-release.sh` script.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
+## Unreleased
+
+- ref: Set wizard version at build time ([#827](https://github.com/getsentry/sentry-wizard/pull/827))
+
 ## 4.0.1
 
-- fix: Remove bulk from npm ([#825](https://github.com/getsentry/sentry-wizard/pull/825)])
+- fix: Remove bulk from npm ([#825](https://github.com/getsentry/sentry-wizard/pull/825))
 
 ## 4.0.0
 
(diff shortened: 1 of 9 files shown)


philprime (approved):
LGTM, I also like this approach way more than the dynamic lookup - especially for a value which should never change after release

BYK:
The process.env.npm_package_version trick was supposed to be static for fossilized builds as fossilize uses esbuild to build it's bundle and bakes this value in and replaces it at build time.

If there's interest we can switch to shipping a single cjs blob with 0 dependencies in our npm package. This is what Craft is doing right now as an example.

PR #828, opened 2025-02-26
ci: Limit signing to release/* and master branches

BYK:
We don't get signing secrets for external contributeros (and we shouldn't!) and we also don't really need to sign the binaries built on PRs. This PR limits signing to master branch and release/** branches.

#skip-changelog


Lms24 (approved):
thanks for fixing!

PR #831, opened 2025-02-27
fix: Update turbopack notice

matejminar:
nit: Just went through the flow and my package.json contained `--turbopack`, not `--turbo`.


constantinius:
There's another mention of `--turbo` [here](https://github.com/getsentry/sentry-wizard/blob/matejminar-patch-1/src/nextjs/nextjs-wizard.ts#L1006). Is this the same option?

andreiborza (approved):
Thanks!

matejminar:
> There's another mention of `--turbo` [here](https://github.com/getsentry/sentry-wizard/blob/matejminar-patch-1/src/nextjs/nextjs-wizard.ts#L1006). Is this the same option?

@constantinius that's a check to determine if this notice needs to be displayed or not - that logic is still correct 👍

andreiborza:
> There's another mention of `--turbo` [here](https://github.com/getsentry/sentry-wizard/blob/matejminar-patch-1/src/nextjs/nextjs-wizard.ts#L1006). Is this the same option?

Yea, I'd say we should check for both `--turbo` and `--turbopack`

constantinius (approved):

PR #833, opened 2025-02-28
feat(nuxt): More granular error catching while modifying config

s1gr1d:
Adds more try/catches around each step of modifying the Nuxt config. Each step fails by setting appropriate tags and displays an error message in the output.

### DIFF
--- a/src/nuxt/sdk-setup.ts
+++ b/src/nuxt/sdk-setup.ts
@@ -123,9 +123,38 @@ export async function addSDKModule(
       `Added Sentry Nuxt Module to ${chalk.cyan(path.basename(config))}.`,
     );
   } catch (e: unknown) {
-    clack.log.error(
-      'Error while adding the Sentry Nuxt Module to the Nuxt config.',
-    );
+    let errorType = 'unknown';
+    let errorMessage = 'Unknown error occurred';
+
+    if (e instanceof Error) {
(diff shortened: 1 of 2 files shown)


src/nuxt/sdk-setup.ts
```diff
@@ -98,10 +100,34 @@ export async function addSDKModule(
...
+
   try {
-    const mod = await loadFile(config);
+    module = await loadFile(config);
+  } catch (e) {
+    if (e instanceof SyntaxError || e.message.includes('Unexpected token')) {
+      Sentry.setTag(failureTagKey, 'loadFile-failed-syntax-error');
+    } else if (
+      e.message.includes('ENOENT') ||
+      e.message.includes('no such file')
```
andreiborza, line 112: m: Maybe we should also guard that `e` is an `Error`. Don't want to crash in case `loadFile` throws something else - even if unlikely.

src/nuxt/sdk-setup.ts
```diff
@@ -98,10 +100,34 @@ export async function addSDKModule(
...
+    ) {
+      Sentry.setTag(failureTagKey, 'loadFile-failed-file-not-found');
+    } else if (e instanceof Error) {
+      Sentry.setTag(failureTagKey, 'loadFile-failed');
+    }
```
andreiborza, line 117: m: And then this can just be an `else`

andreiborza (approved):

PR #837, opened 2025-03-04
feat: add coming-from parameter

obostjancic:
- Part of https://github.com/getsentry/projects/issues/759


lforst (approved):
Code-wise lgtm. However "partner" sounds extremely weird from a user-perspective as a CLI arg. I would name it something like `--coming-from`.

Lms24 (approved):
Please also update the readme with the new arg, thx!

PR #840, opened 2025-03-06
fix(sourcemaps): Prevent addition of multiple `sentry:sourcemaps` commands

obostjancic:
- Fixes: https://github.com/getsentry/sentry/issues/77918

### DIFF
--- a/src/sourcemaps/tools/sentry-cli.ts
+++ b/src/sourcemaps/tools/sentry-cli.ts
@@ -256,6 +256,15 @@ Please add it manually to your prod build command.`,
     return;
   }
 
+  if (oldCommand.endsWith(SENTRY_NPM_SCRIPT_NAME)) {
+    clack.log.info(
+      `The ${chalk.cyan(
+        SENTRY_NPM_SCRIPT_NAME,
+      )} script is already part of your ${chalk.cyan(buildCommand)} command.`,
+    );
+    return;
+  }
+
   packageDotJson.scripts[
     buildCommand
   ] = `${oldCommand} && ${packageManager.runScriptCommand} ${SENTRY_NPM_SCRIPT_NAME}`;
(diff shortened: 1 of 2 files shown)


src/sourcemaps/tools/sentry-cli.ts
```diff
@@ -256,6 +256,15 @@ Please add it manually to your prod build command.`,
     return;
   }
 
+  if (oldCommand.endsWith(SENTRY_NPM_SCRIPT_NAME)) {
+    clack.log.info(
+      `The ${chalk.cyan(
+        SENTRY_NPM_SCRIPT_NAME,
+      )} script is already part of your ${chalk.cyan(buildCommand)} command.`,
+    );
```
lforst, line 264: Thought: What if this matches but the command is outdated for our needs? What if the command is faulty?

Can we add some more information and instructions to the log message so that the user can double-check their setup?

obostjancic: Just saw this 🙈 
I'd like to keep this PR small and focused just on the prevention of multiple appearances of `sentry:sourcemaps` within the build script 
That being said, i think this should be a bit more robust overall but
-  I am not sure how the command would be outdated for our needs if we reference the `SENTRY_NPM_SCRIPT_NAME` and we always keep that up to date. 
- I can not think of a good way to check if the command is faulty without running it and observing the result.

If you have any ideas I'm happy to add them to a followup PR :D

Lms24: I think for now we can go with this approach. I don't know of any concrete plans to update the command, so we can tackle this in a follow-up/later as well. 

Maybe as a quick-fix we could still show the command in the log message, so that users can compare their version to what we _would have_ injected?

Lms24 (approved):
thanks! Had one optional suggestion but happy to let you make the call. Feel free to merge whenever.

src/sourcemaps/tools/sentry-cli.ts
obostjancic: Nice idea, here is how that looks like 
image

PR #847, opened 2025-03-10
ref: Reword Replay feature selection

### LINKED ISSUE #846: [Nuxt]: Simplify Sentry Wizard
### Description

I was testing our Nuxt Sentry Wizard and came across the copy, "Do you want to enable Sentry Session Replay to get a video-like reproduction of errors during a user session?" I was wondering if we should just refer to it as 'Session Replay', similar to how we say 'Tracing' instead of 'Sentry Tracing'.

- Contributes to https://github.com/getsentry/sentry/issues/79408#issuecomment-2706156295

Lms24:
Renames the replay feature from "Sentry Session Replay" to "Session Replay"

closes #846

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 - feat: Add `coming-from` parameter ([#837](https://github.com/getsentry/sentry-wizard/pull/837))
+- ref: Reword Replay feature selection ([#847](https://github.com/getsentry/sentry-wizard/pull/847))
 
 ## 4.2.0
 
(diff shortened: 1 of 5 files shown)


priscilawebdev:
@Lms24 I just used the Sentry CLI in one of my Next.js projects, but I still saw the label 'Sentry Session Replay.' Wasn't it supposed to be released already, or am I missing something?

Screenshot 2025-03-12 at 07 38 39

Lms24:
It wasn't released yet. I'll cut a release today

PR #850, opened 2025-03-10
test(e2e): Fix nuxt e2e tests

Lms24:
Pinning Nuxt to 3.15.0 for now since 3.16.0 breaks the e2e tests. Root cause: https://github.com/getsentry/sentry-javascript/issues/15628

No strong opinions but once we or Nuxt fix the 3.16.0 issue, we can also revert this PR if we prefer testing with the latest version again.

#skip-changelog


andreiborza (approved):
thanks!

Issue #851, opened 2025-03-10
Wizard errors but doesn't tell me what went wrong.


richardasymmetric:
### Environment
- ubuntu 24.04.2
- @sentry/wizard@4.2.0
- yarn 4.7.0
-   npm: '10.9.2',
-  node: '22.14.0',

### Steps to Reproduce

1. npx @sentry/wizard@latest -i nextjs --saas --org orgname --project projectname

### Expected Result

It should install sentry

### Actual Result

```
┌   Sentry Next.js Wizard 
│
◇   ────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                   │
│  The Sentry Next.js Wizard will help you set up Sentry for your application.                      │
│  Thank you for using Sentry :)                                                                    │
│                                                                                                   │
│  Version: 4.2.0                                                                                   │
│                                                                                                   │
│  This wizard sends telemetry data and crash reports to Sentry. This helps us improve the Wizard.  │
│  You can turn this off at any time by running sentry-wizard --disable-telemetry.                  │
│                                                                                                   │
├───────────────────────────────────────────────────────────────────────────────────────────────────╯
│
●  If the browser window didn't open automatically, please open the following link to log into Sentry:
│  
│  https://sentry.io/account/settings/wizard/.....
│
◇  Login complete.
│
◇  Selected project orgname/projectname
│
◇  Installation failed.
│
■  Encountered the following error during installation:
│  
│  Error: Command failed: yarn add @sentry/nextjs@^9  
│  
│  
│  The wizard has created a `sentry-wizard-installation-error-*.log` file. If you think this issue is caused by the Sentry wizard, create an issue on GitHub and include the log file's content:
│  https://github.com/getsentry/sentry-wizard/issues
│
└  Wizard setup cancelled.
```

Logfile: ``` {"stdout":"","stderr":""} ```

I'm happy to do any debugging, just need to know how/where

Lms24:
Hey @richardasymmetric thanks for reporting this! I'll have a look if we can be more verbose regarding logging the error. I agree that this unfortunately doesn't tell us a lot. Sorry that you're experiencing this!

Shot in the dark: We haven't tested yarn > 1 a lot so maybe there's some parameter combination that doesn't work well. 

Some other things to somewhat scope this:
- are you in a monorepo/yarn workspace project? If so, did you run the wizard in the package root or in the monorepo root directory?
- Did this fail consistently or was it a one-off? 

[...]

PR #859, opened 2025-03-17
ref(clack-utils): Use `child_process` `spawn` instead of `exec` when installing package

Lms24:
This PR refactors our `installPackage` helper to use `spawn` instead of `exec`. While using spawn is a bit more complicated, it gives us finer control over the the child process lifecycle as well as runtime. This is primarily made because I want to attempt fixing #851.

Additionally, we now send an error to Sentry  when this happens to keep track how often this would occur. Previously, such errors weren't captured by Sentry since we try/caught the promise rejection. 

I'll open another follow-up PR to split up the `clack-utils` file. This thing has become huge and no longer is easy to maintain.

### DIFF
--- a/src/utils/clack-utils.ts
+++ b/src/utils/clack-utils.ts
@@ -418,11 +418,7 @@ export async function installPackage({
             { encoding: 'utf8' },
           );
 
-          const errorMessage = `Installation command ${chalk.cyan(
-            stringifiedInstallCmd,
-          )} exited with code ${code}.`;
-
-          Sentry.captureException(errorMessage, {
+          Sentry.captureException('Package Installation Error', {
             tags: {
               'install-command': stringifiedInstallCmd,
               'package-manager': pkgManager.name,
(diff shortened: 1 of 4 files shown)


src/utils/clack-utils.ts
```diff
@@ -383,31 +385,87 @@ export async function installPackage({
...
+              `sentry-wizard-installation-error-${Date.now()}.log`,
+            ),
+            JSON.stringify(
+              {
+                stdout,
+                stderr,
+              },
+              null,
+              2,
+            ),
```
lforst, line 417: Dunno if it's worth doing this change. The output is an object with two gargantuan unreadable strings anyhow 🤷‍♂️

lforst (approved):

src/utils/clack-utils.ts
Lms24: fair enough :) reverted for now

PR #861, opened 2025-03-18
feat(react-native): Improve error reporting for RN Wizard

antonis:
Fixes https://github.com/getsentry/sentry-wizard/issues/841

## Description

This PR enhances error handling and logging for the RN Wizard

### DIFF
--- a/src/react-native/expo-metro.ts
+++ b/src/react-native/expo-metro.ts
@@ -31,7 +31,18 @@ export async function addSentryToExpoMetroConfig() {
     return undefined;
   }
 
+  Sentry.setTag('expo-metro-config', 'exists');
+  clack.log.info(`Updating existing ${metroConfigPath}.`);
+
   const mod = await parseMetroConfig();
+  if (!mod) {
+    clack.log.error(
+      `Could read from file ${chalk.cyan(
+        metroConfigPath,
+      )}, please follow the manual steps.`,
+    );
+    return await showInstructions();
+  }
 
   let didPatch = false;
   try {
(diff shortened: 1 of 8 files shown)


src/react-native/expo-metro.ts
```diff
@@ -28,7 +28,18 @@ export async function addSentryToExpoMetroConfig() {
...
+  Sentry.setTag('expo-metro-config', 'exists');
+  clack.log.info(`Updating existing ${metroConfigPath}.`);
+
   const mod = await parseMetroConfig();
+  if (!mod) {
+    clack.log.error(
+      `Could read from file ${chalk.cyan(
+        metroConfigPath,
+      )}, please follow the manual steps.`,
+    );
```
krystofwoldrich, line 40: The `parseMetroConfig` already log the error, we can only show the instructions here.

krystofwoldrich:
The improvements look good and make sense; let's just make sure every time we fail to do something to show copy copy-pastable manual step and `Sentry.captureExption('Error desc without PII')`.

src/react-native/expo-metro.ts
antonis: Good catch Krystof 👍 
Removed the duplicate error log with d55407e023747ec31a80facf71f644edae1e4f5a

krystofwoldrich:
@antonis Let me know if this ready for another pass, I still see some pending comments, feel free to just add a comment to them or 👍/👎.

antonis:
Thank you for your feedback @krystofwoldrich 🙇 

> let's just make sure every time we fail to do something to show copy copy-pastable manual step

I didn't identify more places were manual steps are missing.

> and Sentry.captureExption('Error desc without PII').

Updated with 5fbf440559f341abc6ed62c262e73195bdab51fb

> Due to possible PII in the stack traces, we can't at the moment capture the errors.

I removed the stack traces introduced in the commit above with fefeb2920b576894544f735dcdba4fa69ee1a071

This should be ready for another pass 🙏

PR #863, opened 2025-03-18
fix(nextjs): Create root layout for example page if it doesn't exist

### LINKED ISSUE #607: [Next.js] Check for existence of root layout before creating test page
We need to check whether a root layout exists before creating the test page. If not, we need to create a layout next to the test page ourselves.

Ref: https://github.com/getsentry/sentry-wizard/pull/605

Lms24:
This PR adds a check for a NextJS [root layout](https://nextjs.org/docs/app/api-reference/file-conventions/layout#root-layouts) file. Some NextJS apps don't have a root layout though, for instance if they only serve API routes. If users still would like to create an example page via the wizard, we also need to add a root layout so that the page can be rendered.

I noticed that newer versions actually create a root layout file automatically in dev mode when first serving the page but this doesn't apply to older NextJS versions. So it's better to handle this explicitly in the wizard. The file we inject is almost identical to the auto-generated one, just with some Sentry-specific text. 

This is tested by removing the root layout in our NextJS 14 E2E test app and assuring that it's re-added by our wizard in the tests.

closes #607


lforst (approved):
Nice! Ty!

PR #865, opened 2025-03-18
ref(utils): Unify `getPackageManger` and `detectPackageManager`

Lms24:
In our utility functions, we used to have two functions to obtain a package manager: `getPackageManger` and `detectPackageManager` which are both used throughout the code base. 

Both had slightly different semantics, namely that `detectPackageManager` would only attempt auto-detection but has no possibility to fall back to asking users which manager they're using. 

This PR now unifies both functions by:
- going forward only using `getPackageManager` everywhere
  - which internally uses `detectPackageManager`
- by default falling back to asking users which package manager they're using
- providing an optional fallback package manager param which will be returned _instead of asking the user_.
  - this is useful when we only need the manager e.g. in an outro message or so where it's not a 100% crucial to get the correct manager and asking users would be too much of a UX burden.
- caching the auto-detected or user-selected package manager to avoid going through this process multiple times. This also means that the fallback value in outro message already most likely isn't necessary because we almost always install a package beforehand.
- Adding more tests for the function

### DIFF
--- a/src/sourcemaps/sourcemaps-wizard.ts
+++ b/src/sourcemaps/sourcemaps-wizard.ts
@@ -319,7 +319,6 @@ async function printOutro(
   projectId: string,
 ): Promise<void> {
   const packageManager = await getPackageManager(NPM);
-  const buildCommand = packageManager.buildCommand ?? 'npm run build';
 
   const issueStreamUrl = getIssueStreamUrl({ url, orgSlug, projectId });
 
@@ -330,7 +329,9 @@ async function printOutro(
    ${chalk.cyan(`Test and validate your setup locally with the following Steps:
 
    1. Build your application in ${chalk.bold('production mode')}.
-      ${chalk.gray(`${arrow} For example, run ${chalk.bold(buildCommand)}.`)}
+      ${chalk.gray(
+        `${arrow} For example, run ${chalk.bold(packageManager.buildCommand)}.`,
+      )}
       ${chalk.gray(
         `${arrow} You should see source map upload logs in your console.`,
       )}


src/sourcemaps/sourcemaps-wizard.ts
```diff
@@ -312,8 +313,12 @@ SENTRY_AUTH_TOKEN=${authToken}
...
 
-function printOutro(url: string, orgSlug: string, projectId: string) {
-  const packageManager = detectPackageManger();
+async function printOutro(
+  url: string,
+  orgSlug: string,
+  projectId: string,
+): Promise<void> {
+  const packageManager = await getPackageManager(NPM);
   const buildCommand = packageManager?.buildCommand ?? 'npm run build';
```
andreiborza, line 322: l: we could remove this fallback here right?

src/utils/package-manager.ts
```diff
@@ -157,6 +157,11 @@ export const NPM: PackageManager = {
 
 export const packageManagers = [BUN, YARN_V1, YARN_V2, PNPM, NPM];
 
+/**
+ * Exported only for testing.
+ * DO NOT call this function directly!
+ * Use `getPackageManger` instead.
+ */
 export function detectPackageManger(
```
andreiborza, line 165: ultra-l: maybe we could prefix it with an underscore

andreiborza (approved):
Nice!

src/sourcemaps/sourcemaps-wizard.ts
Lms24: I think realistically, every source maps wizard flow installs at least one package so the fallback should never be used as the cache will already be populated. That being said, having the fallback doesn't hurt us either, for example in case we'd add some kind of minimalistic source maps tool setup flow where we don't install stuff. So if you're fine with it, I'd leave it in.

My rationale is that I don't think we should ever ask users which package manager they're using, just to display the last outro message. Does this sound reasonable to you?

src/utils/package-manager.ts
Lms24: good idea! I was contemplating moving it into `getPackageManager` entirely but this makes it way harder to test. Adding the leading underscore sounds like a good compromise :D

Lms24: [29039f4](https://github.com/getsentry/sentry-wizard/pull/865/commits/29039f4688dc004fe75c988081190a68b85cea2f)

src/sourcemaps/sourcemaps-wizard.ts
andreiborza: No, that totally makes sense but I actually mean the `buildCommand` line, we don't need the `npm run build` fallback because you already set `NPM` as the fallback manager so there is always a package manager available right?

Lms24: ah sorry, I misunderstood. You're right! [1594b1c](https://github.com/getsentry/sentry-wizard/pull/865/commits/1594b1c329503479161df6af150d699d900b8b06)

andreiborza: 👍 I realise it's not always clear which line is meant when a preview is rendered, I should have been more specific and avoid using words like "this", especially when the comment is for a line that hasn't been changed by the PR :).

Issue #572, opened 2024-05-25
Sentry Wizard fails to modify `Bundle React Native and images`


thomasttvo:
### Environment

Due to the older versions of Sentry not working with Hermes, we needed to modify `Bundle React Native and images` to support hermes, at which time we removed `$REACT_NATIVE_XCODE` from the build phase. The wizard for the latest version just performs a string replacement without checking whether the `$REACT_NATIVE_XCODE` is there, which reports false positive.

### Steps to Reproduce

Run `npx @sentry/wizard@latest -s -i reactNative`

### Expected Result

The wizard alerts us that it's unable to fix this build phase. It does that for metro config but not for `Bundle React Native and images`

### Actual Result

The wizard reported that it was able to fix the build phase just fine ✅ , which is wrong

Lms24:
Hey @thomasttvo thanks for writing in! Sounds like a reasonable expectation to me at first glance but I'm not deep enough into RN. Gonna tag @krystofwoldrich - would you mind taking a look please?

kahest:
@krystofwoldrich please give the [linked PR](https://github.com/getsentry/sentry-wizard/pull/866) another round of review so we can close this 🙏

PR #866, opened 2025-03-18
fix(react-native): Handle xcode build phase patching failure

antonis:
Fixes https://github.com/getsentry/sentry-wizard/issues/572

Handled `Bundle React Native and images` build phase patching gracefully and show a code snippet for manual configuration.

Screenshot 2025-03-19 at 2 42 32 PM


src/react-native/xcode.ts
```diff
@@ -148,10 +164,38 @@ export function addSentryWithBundledScriptsToBundleShellScript(
     );
   }
 
+  if (!isLikelyPlainReactNativeScript && !isLikelyExpoScript) {
```
krystofwoldrich, line 167: I think, we should always show the copy-paste snippet when we fail to patch the script.

Even when the script is likely expo or likely plain RN the replace might fail.

src/react-native/xcode.ts
```diff
@@ -45,7 +53,15 @@ export function patchBundlePhase(
   }
 
   const script: string = JSON.parse(bundlePhase.shellScript);
-  bundlePhase.shellScript = JSON.stringify(patch(script));
+  const patchedScript = patch(script);
+  if (patchedScript instanceof ErrorPatchSnippet) {
+    await showCopyPasteInstructions(
```
krystofwoldrich, line 58: I think we should clearly state they need to update the existing code. 

To avoid the user just copy pasting the shown code at the end of the build script.

The snippet method already allows to color the new/removed/existing code.

src/react-native/xcode.ts
antonis: Good idea Krystof 👍 
Updated with 0b075ee1aa8dbad1d2cab747cca0fa1fc648720e to show the snippet when patching fails for expo and react native.

src/react-native/xcode.ts
antonis: Good point @krystofwoldrich 👍 
Updated with 94f44edf675d63abab39771da3f6db22adfebaeb

src/react-native/xcode.ts
```diff
@@ -142,17 +157,73 @@ export function addSentryWithBundledScriptsToBundleShellScript(
...
+        'Bundle React Native code and images',
+      )} build phase.`,
+    );
+    if (isLikelyExpoScript) {
+      return new ErrorPatchSnippet(
+        makeCodeSnippet(true, (unchanged, plus, _minus) => {
+          return unchanged(
+            `${plus(
+              `/bin/sh \`"$NODE_BINARY" --print "require('path').dirname(require.resolve('@sentry/react-native/package.json')) + '/scripts/sentry-xcode.sh'"\``,
+            )}\`"$NODE_BINARY" --print "require('path').dirname(require.resolve('react-native/package.json')) + '/scripts/react-native-xcode.sh'"\``,
```
krystofwoldrich, line 180: L: It doesn't break any functionality, but it looks better when printed out with the space in between.

````suggestion
            )} \`"$NODE_BINARY" --print "require('path').dirname(require.resolve('react-native/package.json')) + '/scripts/react-native-xcode.sh'"\``,
````

File: src/react-native/xcode.ts [hunk 4]
```diff
@@ -142,17 +157,73 @@ export function addSentryWithBundledScriptsToBundleShellScript(
...
+      makeCodeSnippet(true, (unchanged, plus, _minus) => {
+        return unchanged(`${plus(`export SENTRY_PROPERTIES=sentry.properties
+export EXTRA_PACKAGER_ARGS="--sourcemap-output $DERIVED_FILE_DIR/main.jsbundle.map"
+`)}
+/bin/sh -c "$WITH_ENVIRONMENT ${plus(
+          `\\"../node_modules/@sentry/cli/bin/sentry-cli react-native xcode`,
+        )} $REACT_NATIVE_XCODE${plus(`\\"`)}"
+${plus(
+  `/bin/sh -c "$WITH_ENVIRONMENT ../node_modules/@sentry/react-native/scripts/collect-modules.sh`,
+)}"
```
krystofwoldrich, line 222: The last " was not marked as added.

Screenshot 2025-04-11 at 18 56 59

````suggestion
  `/bin/sh -c "$WITH_ENVIRONMENT ../node_modules/@sentry/react-native/scripts/collect-modules.sh"`,
)}
````

File: src/react-native/xcode.ts [hunk 5]
```diff
@@ -46,7 +51,16 @@ export function patchBundlePhase(
   }
 
   const script: string = JSON.parse(bundlePhase.shellScript);
-  bundlePhase.shellScript = JSON.stringify(patch(script));
+  const patchedScript = patch(script);
+  if (patchedScript instanceof ErrorPatchSnippet) {
+    await showCopyPasteInstructions(
```
krystofwoldrich, line 56: This function expects a file name, which causes the `'Bundle React Native code and images' build phase in your Xcode project` part to be colored.

I would just copy paste the impl here and adjust it for the Xcode copy paste, so only `Bundle React Native code and images` is colored. And since the text is quite long, we could add the note to a new line.

Screenshot 2025-04-11 at 19 09 23

krystofwoldrich:
I've noticed two small visual details; otherwise, it looks great!

src/react-native/xcode.ts
antonis: Good catch @krystofwoldrich 🙇 

> This function expects a file name, which causes the 'Bundle React Native code and images' build phase in your Xcode project part to be colored.

I tried to reuse the function and added the extra text `...in your Xcode project` to make sense of [the “…file” added at the end of the string](https://github.com/getsentry/sentry-wizard/blob/9b082e9828b6858e0117b12bc742ae5212409321/src/utils/clack/index.ts#L1463).

I tried an alternative shorter approach with 4c4c52bd6a922736257a2a69144798b3f71e8fda
Screenshot 2025-04-14 at 12 46 57 PM

I'll be happy to extend the `showCopyPasteInstructions` to be more flexible in a separate PR and iterate on this or just skip using the function on this case.

krystofwoldrich: I have not thought about this approach, but it works great. Thank you.

PR #867, opened 2025-03-19
ref(eslint): Fix eslint warnings (part 1)

Lms24:
We have far too many eslint warnings in the wizard repo. Most of these are easily fixable, so let's fix them. 

Goal: Treat eslint warnings as errors so that we don't get into this state again.

This PR is a first part of this initiative, containing around 25 eslint warning fixes. We'll gradually fix all warnigns over time.

Note: Most of these PRs will contain codecov warnings about coverage. I do think we should increase coverage but this is out of scope for the lint refactors (though we can of course always throw in tests if appropriate). So I'll largely ignore codecov in these PRs. 

#skip-changelog

ref: #868

### DIFF
--- a/src/remix/codemods/handle-error.ts
+++ b/src/remix/codemods/handle-error.ts
@@ -17,6 +17,8 @@ import chalk from 'chalk';
 import { generateCode } from 'magicast';
 
 export function instrumentHandleError(
+  // MagicAst returns `ProxifiedModule<any>` so therefore we have to use `any` here
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   originalEntryServerMod: ProxifiedModule<any>,
   serverEntryFilename: string,
 ): boolean {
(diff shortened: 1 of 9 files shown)


src/remix/codemods/handle-error.ts
```diff
@@ -17,6 +17,8 @@ import chalk from 'chalk';
 import { generateCode } from 'magicast';
 
 export function instrumentHandleError(
+  // MagicAst returns `ProxifiedModule<any>` so therefore we have to use `any` here
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   originalEntryServerMod: ProxifiedModule<any>,
```
Lms24, line 22: This isn't great and actually a problem in many wizard flows. I played around with somehow typing this better but didn't get anywhere yet. While magicast is nice to use for small changes, it's quite painful with typing and its abstractions on top of recase. I think we should consider ditching it in favour of recast so I don't want to spend too much time on fixing these types. If anyone has ideas how to type the proxified modules properly, I'm all ears :D

andreiborza (approved):

PR #879, opened 2025-03-19
ref(eslint): Set `@typescript-eslint/ban-ts-comment` rule to `error` 

### LINKED ISSUE #869: [eslint] Set '@typescript-eslint/ban-ts-comment': 'off' to 'error'
While this rule is off today, enabling it provides a lot of value and makes sure we have no unnecessary ts-ignore comments anymore

Lms24:
This PR still needs a rebase after #867 is merged but otherwise it's ready for a review

Enables the previously disabled `@typescript-eslint/ban-ts-comment` rule to throw a lint error in case of a violation.

Most changes are a simple replacement of `@ts-ignore` -> `@ts-expect-error` because we have to ignore some `clack` and `magicast` related import type errors (we should look at these separately!). I left comments for all actual code changes/fixes. 

#skip-changelog
closes #869

### DIFF
--- a/lib/Helper/Wizard.ts
+++ b/lib/Helper/Wizard.ts
@@ -6,20 +6,16 @@ import type { BaseIntegration } from '../Steps/Integrations/BaseIntegration';
 import { BottomBar } from './BottomBar';
 import { debug, dim, nl, red } from './Logging';
 
-function sanitizeAndValidateArgs(argv: Args): void {
+function sanitizeAndValidateArgs(argv: Args & Record<string, unknown>): void {
   if (argv.quiet === undefined) {
     argv.quiet = true;
     dim('will activate quiet mode for you');
   }
-  // @ts-ignore skip-connect does not exist on args
   if (argv['skip-connect']) {
-    // @ts-ignore skip-connect does not exist on args
-    argv.skipConnect = argv['skip-connect'];
-    // @ts-ignore skip-connect does not exist on args
+    argv.skipConnect = argv['skip-connect'] as Args['skipConnect'];
     delete argv['skip-connect'];
   }
-  // @ts-ignore skip-connect does not exist on args
-  argv.promoCode = argv['promo-code'];
+  argv.promoCode = argv['promo-code'] as Args['promoCode'];
 }
 
 export function getCurrentIntegration(answers: Answers): BaseIntegration {
@@ -31,7 +27,7 @@ export async function startWizard<M extends IStep>(
   ...steps: Array<{ new (debug: Args): M }>
 ): Promise<Answers> {
   try {
-    sanitizeAndValidateArgs(argv);
+    sanitizeAndValidateArgs(argv as Args & Record<string, unknown>);
     if (argv.debug) {
       debug(argv);
     }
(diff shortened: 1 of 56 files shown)


lib/Helper/Wizard.ts
```diff
@@ -6,20 +6,16 @@ import type { BaseIntegration } from '../Steps/Integrations/BaseIntegration';
 import { BottomBar } from './BottomBar';
 import { debug, dim, nl, red } from './Logging';
 
-function sanitizeAndValidateArgs(argv: Args): void {
+function sanitizeAndValidateArgs(argv: Args & Record<string, unknown>): void {
```
Lms24, line 9: Widening the type here allows us to get rid of a couple of ts-ignores. Not the prettiest fix but this is the legacy wizard code which I con't care too much about.

lib/Steps/Integrations/BaseIntegration.ts
```diff
@@ -9,8 +9,7 @@ export abstract class BaseIntegration extends BaseStep {
 
   public constructor(protected _argv: Args) {
     super(_argv);
-    // @ts-ignore property construct does not exist on BaseIntegration
-    this.type = this.construct;
+    this.type = this.constructor.name;
```
Lms24, line 12: i think this was a straight up bug beforehand (?)

lib/__tests__/Env.ts
```diff
@@ -2,20 +2,14 @@ import { readEnvironment } from '../Helper/Env';
...
-    // @ts-ignore: true not assignable to string/undefined
-    process.env.SENTRY_WIZARD_SKIP_CONNECT = true;
-    // @ts-ignore: true not assignable to string/undefined
-    process.env.SENTRY_WIZARD_QUIET = true;
-    // @ts-ignore: true not assignable to string/undefined
-    process.env.SENTRY_WIZARD_INTEGRATION = ['reactNative', 'electron'];
-    // @ts-ignore: true not assignable to string/undefined
-    process.env.SENTRY_WIZARD_PLATFORM = ['ios', 'android'];
-    // @ts-ignore: true not assignable to string/undefined
+    process.env.SENTRY_WIZARD_DEBUG = 'true';
```
Lms24, line 5: TS complained rightfully here. Changing the values to string here is the right thing to do and the test still passes (meaning `read-env` did its job)

src/sveltekit/sdk-setup.ts
```diff
@@ -446,13 +446,16 @@ function wrapHandle(mod: ProxifiedModule<any>): void {
     } else if (declaration.type === 'VariableDeclaration') {
       const declarations = declaration.declarations;
       declarations.forEach((declaration) => {
-        // @ts-ignore - id should always have a name in this case
-        if (!declaration.id || declaration.id.name !== 'handle') {
+        if (
+          !declaration.id ||
+          declaration.id.type !== 'Identifier' ||
+          (declaration.id.name && declaration.id.name !== 'handle')
+        ) {
```
Lms24, line 453: I lied in this ts-ignore comment 😅  There's no guarantee for the declaration.id to always have a name. The new check is more defensive and makes no more assumptions.

andreiborza (approved):
These clack ignores are wild 🫠

lib/Steps/Integrations/BaseIntegration.ts
andreiborza: Yea, looks weird.

PR #881, opened 2025-03-19
ref(eslint): Set `@typescript-eslint/no-unsafe-call` to `error`

### LINKED ISSUE #870: [eslint] Set '@typescript-eslint/no-unsafe-call': 'warn' to 'error'

Lms24:
This surfaced some weird type issue in the cordova wizard (see comment)

#skip-changelog

closes #870

### DIFF
--- a/lib/Steps/Integrations/Cordova.ts
+++ b/lib/Steps/Integrations/Cordova.ts
@@ -188,7 +188,6 @@ export class Cordova extends BaseIntegration {
       [],
       'PBXShellScriptBuildPhase',
       'Upload Debug Symbols to Sentry',
-      // @ts-expect-error - xcode project type expects a string for the target param but we pass null (not sure why but this wasn't changed)
       null,
       {
         shellPath: '/bin/sh',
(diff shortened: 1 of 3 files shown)


lib/Steps/Integrations/Cordova.ts
```diff
@@ -179,6 +188,7 @@ export class Cordova extends BaseIntegration {
       [],
       'PBXShellScriptBuildPhase',
       'Upload Debug Symbols to Sentry',
+      // @ts-expect-error - xcode project type expects a string for the target param but we pass null (not sure why but this wasn't changed)
```
Lms24, line 191: @philprime I think you added types for `xcode` recently right? Any idea if this is an actual bug/misuse of the library on our end or a typing issue?

philprime: Yes I did add the types but they were hand-made so there is a high probability that you should change the type to be nullable instead of the @ts-expect-error

Lms24: Thanks! I adjusted the `Project` type declaration ([c938286](https://github.com/getsentry/sentry-wizard/pull/881/commits/c93828679623946f210447e4d6521494c502d3da)). Turns out, the `@ts-expect-error` also muted another type error underneath which also required me to make a couple of the `optionsOrFolderType` props optional.

philprime (approved):
LGTM, thanks for replacing `any` with types

Issue #809, opened 2025-02-21
[flutter-wizard] Confirmation Instructions


smeubank:
The wizard is fast and works well, at the end we should be more instructive on what they should do to test errors, tracing and dsym uploads.

it should tell me what to run to test, be very direct with developer run `flutter run`

or better when there is dsym support, run xyz to upload source maps, THEN run abc to test everything out

ideally the error we inject we tell them about it, and it will have a transaction, like page load, directly associated with it.

![Image](https://github.com/user-attachments/assets/dcd41c1f-8793-4e70-9be1-9071e02f8b0d)

PR #886, opened 2025-03-20
Improve Next Steps for Flutter Outro

smeubank:
Adding something more concrete to the end of the flutter wizard to tell users

1. how to confirm/test setup
2. what more they can do

https://github.com/getsentry/sentry-wizard/issues/809

### DIFF
--- a/src/flutter/flutter-wizard.ts
+++ b/src/flutter/flutter-wizard.ts
@@ -157,12 +157,14 @@ Set the ${chalk.cyan(
   clack.outro(`
     ${chalk.greenBright('Successfully installed the Sentry Flutter SDK!')}
     
-    ${chalk.cyan(
-      `You can validate your setup by launching your application and checking Sentry issues page afterwards
-    ${issuesPageLink}`,
-    )}
+    ${chalk.cyan('Next steps:')}
+    1. Run ${chalk.bold('flutter run')} to test the setup - we've added a test error that will trigger on app start
+    2. For production builds, run ${chalk.bold('flutter build ios --obfuscate --split-debug-info=build/debug-info')} (or android/macos) then ${chalk.bold('flutter pub run sentry_dart_plugin')} to upload debug symbols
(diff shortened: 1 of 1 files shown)


src/flutter/flutter-wizard.ts
```diff
@@ -157,12 +157,14 @@ Set the ${chalk.cyan(
   clack.outro(`
     ${chalk.greenBright('Successfully installed the Sentry Flutter SDK!')}
     
-    ${chalk.cyan(
-      `You can validate your setup by launching your application and checking Sentry issues page afterwards
-    ${issuesPageLink}`,
-    )}
+    ${chalk.cyan('Next steps:')}
+    1. Run ${chalk.bold('flutter run')} to test the setup - we've added a test error that will trigger on app start
+    2. For production builds, run ${chalk.bold('flutter build ios --obfuscate --split-debug-info=build/debug-info')} (or android/macos) then ${chalk.bold('flutter pub run sentry_dart_plugin')} to upload debug symbols
```
buenaflor, line 162:
```suggestion
    2. For production builds, run ${chalk.bold('flutter build apk --obfuscate --split-debug-info=build/debug-info')} (or ios/macos) then ${chalk.bold('flutter pub run sentry_dart_plugin')} to upload debug symbols
```

buenaflor: qq: do we need to make it clear that (apk | ios | macos) are not the only builds that we support?

for example on android it's possible to do `flutter build appbundle` or `flutter build apk` and on ios/macos there are a bunch of other ways as well. 

essentially it doesn't matter what type of production build they do, what's important is that they add these flags `--obfuscate --split-debug-info=build/debug-info'`

smeubank: i updated it as you suggested, generally users can see the docs. I want to create a ticket to better handle dSYM for android and iOS. For now they are at least given some instructions and links to docs

buenaflor (approved):

PR #884, opened 2025-03-20
test(react-native): Add E2E Tests for React Native and Expo projects

antonis:
Fixes https://github.com/getsentry/sentry-wizard/issues/856

## Description
This PR adds happy path E2E tests for RN and Expo:
- [Build & Test / Expo E2E Tests (pull_request)](https://github.com/getsentry/sentry-wizard/actions/runs/13976534329/job/39131701164?pr=884)
    ✓ package.json is updated correctly
    ✓ _layout.tsx is updated correctly
    ✓ app.json is updated correctly
    ✓ metro.config.js is added
    ✓ .gitignore is updated correctly
- [Build & Test / ReactNative E2E Tests (pull_request)](https://github.com/getsentry/sentry-wizard/actions/runs/13976534329/job/39131704302?pr=884)
    ✓ package.json is updated correctly
    ✓ metro.config.js is updated correctly
    ✓ App.tsx is updated correctly
    ✓ android/sentry.properties is added
    ✓ build.gradle is updated correctly
   - ⚠️  The following ios checks were disabled on this PR with  https://github.com/getsentry/sentry-wizard/pull/884/commits/80ccdcdc9dd5f8d5f33b29dd2a259f99ec8d608b due to an issue running [`xcodeProject.addBuildPhase`](https://github.com/getsentry/sentry-wizard/blob/b25760001ebc0a2282c9716f60cb11e7be963e2d/src/react-native/xcode.ts#L183) on CI. This will be handled separately with https://github.com/getsentry/sentry-wizard/pull/889 🤞 
     ✓ ios/sentry.properties is added
     ✓ 'xcode project is updated correctly
- More test projects and test verifying other cases can be added separately

### Test applications
- A new React Native `0.78.1` app
- A new Expo `52` app

### To test locally run:
- `yarn test:e2e ReactNative`
- `yarn test:e2e Expo`

Issue #856, opened 2025-03-13
Add E2E Tests for React Native and Expo projects


krystofwoldrich:

antonis:
Closing with:
- https://github.com/getsentry/sentry-wizard/pull/884
-  https://github.com/getsentry/sentry-wizard/pull/889

The other tasks will tracked with [the issues mentioned above](https://github.com/getsentry/sentry-wizard/issues/856#issuecomment-2830349259).

PR #889, opened 2025-03-20
test(react-native): Add E2E iOS Tests for React Native

antonis:
**Based on: https://github.com/getsentry/sentry-wizard/pull/884**

Part of https://github.com/getsentry/sentry-wizard/issues/856

## Description
This PR aims to test the iOS React Native wizard modification:

✅  xcode project is updated correctly
   - ⚠️ The [`addBuildPhase`](https://github.com/getsentry/sentry-wizard/blob/master/src/react-native/xcode.ts#L183)([library implementation](https://github.com/apache/cordova-node-xcode/blob/master/lib/pbxProject.js#L872)) execution [fails (stucks) on CI](https://github.com/getsentry/sentry-wizard/actions/runs/13988452682/job/39166888988?pr=889). Note that all tests complete successfully locally.
      - Patching [the `xcode` library](https://github.com/apache/cordova-node-xcode) with  1d404ecfb54bb6a203e4e1d58ac244fefbe38de9 using [patch-package](https://github.com/ds300/patch-package) resolved the issues described in https://github.com/getsentry/sentry-wizard/pull/889#issuecomment-2828082568
 
✅ `ios/sentry.properties` is added

### To test locally run:
- `yarn test:e2e React-Native`

#skip-changelog


antonis:
> I'm still investigating why this happens and how to work around it.

[Adding a couple of missing declarations in the fork](https://github.com/apache/cordova-node-xcode/compare/master...antonis:cordova-node-xcode:add-missing-declarations) made [the tests to run successfully on CI](https://github.com/getsentry/sentry-wizard/actions/runs/14645580363/job/41098912121?pr=889) in strict mode 🎉 

The next step would be to come up with a solution that doesn't require forking https://github.com/apache/cordova-node-xcode

.github/workflows/build.yml
```diff
@@ -148,6 +148,11 @@ jobs:
           - ubuntu-latest
           # We need to use macos-15 to test with Xcode 16
           - macos-15
+        exclude:
+          - wizard: React-Native
+        include:
+          - wizard: React-Native
+            os: macos-15
```
antonis, line 155: Running on `ubuntu-latest` will make the `pod-install` step to be skipped and the test to fail.

patches/xcode+3.0.1.patch
antonis, line 1: Pathing was needed to fix the CI failure as described in https://github.com/getsentry/sentry-wizard/pull/889#issuecomment-2828082568

Probably this also fixes failures in environments similar to our CI setup too.

.github/workflows/build.yml
krystofwoldrich: Can we run the test with `pod-install -> no` on ubuntu?

patches/xcode+3.0.1.patch
krystofwoldrich: Since we were unable to reproduce this locally, it's okay to patch the lib for the E2E tests, but if we get any reports of this or we will be able to reproduce this in the future we need to fix the Xcode lib or migrate to a new supported lib (for example bacons/xcode).

package.json
```diff
@@ -62,6 +62,8 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-vitest": "^0.5.4",
     "fossilize": "^0.3.1",
+    "patch-package": "^8.0.0",
+    "postinstall-postinstall": "^2.1.0",
```
krystofwoldrich, line 66: Do we need the `postinstall-postinstall` package?

patches/xcode+3.0.1.patch
antonis: Makes sense 👍

package.json
```diff
@@ -80,6 +82,7 @@
   },
   "scripts": {
     "clean": "rm -rf ./dist",
+    "postinstall": "patch-package",
```
krystofwoldrich, line 85: I think we should run this manually in CI instead of the `postinstall` script. Since the post install would be also executed when users install the CLI.

krystofwoldrich:
The test looks good, I just have a two questions about the post install script and one about the run on linux.

package.json
antonis: I think it is needed according to [the `patch-package` instructions](https://github.com/ds300/patch-package?tab=readme-ov-file#yarn-v1) since [the CI is using yarn 1](https://github.com/getsentry/sentry-wizard/actions/runs/14663735607/job/41153718839?pr=889).

krystofwoldrich: Thank you I didn't know that.

package.json
krystofwoldrich: I just wan't to make sure this runs only for the development/tests, but not in prod when users install the wizard.

.github/workflows/build.yml
antonis: I reverted the `build.yml` change and [optionally skipped the `pod-install` check and iOS tests](https://github.com/getsentry/sentry-wizard/pull/889/files#diff-6280c9ecf15f057015e5983b97b9f4b840ac3dd3bb11ef5a3b8bd6f4d670a573).

package.json
antonis: Thank you for all the feedback @krystofwoldrich 🙇 
I was [initially thinking](https://github.com/getsentry/sentry-wizard/pull/889#discussion_r2060085493) of applying in all builds but since we have no reports of this issue it makes sense to keep it only for ci testing. [I reverted the dependencies additions (of both `patch-package` and. `postinstall`)](https://github.com/getsentry/sentry-wizard/pull/889/commits/152dee5f6d407a22e930a923ee209a89f434c101) and [just apply the patch before building on CI](https://github.com/getsentry/sentry-wizard/pull/889/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721).

krystofwoldrich (approved):
🚀 Looks good, thank you for adding the E2E tests.

PR #891, opened 2025-03-21
test(flutter): Fix failing snapshot tests

Lms24:
I'm not sure why these tests weren't failing in the respective PR (#887) but these are currently blocking other PRs. Updated the inline snapshots to reflect the changes. Reviewers please check if the snapshots are as expected, thanks :) 

#skip-changelog

### DIFF
--- a/test/flutter/templates.test.ts
+++ b/test/flutter/templates.test.ts
@@ -9,13 +9,13 @@ describe('Flutter code templates', () => {
     it('generates pubspec with project and org', () => {
       const template = pubspecOptions('fixture-project', 'fixture-org');
       expect(template).toMatchInlineSnapshot(`
-      "sentry:
-        upload_debug_symbols: true
-        upload_source_maps: true
-        project: fixture-project
-        org: fixture-org
-      "
-      `);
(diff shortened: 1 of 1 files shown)


test/flutter/templates.test.ts
```diff
@@ -9,13 +9,13 @@ describe('Flutter code templates', () => {
...
       const template = pubspecOptions('fixture-project', 'fixture-org');
       expect(template).toMatchInlineSnapshot(`
-      "sentry:
-        upload_debug_symbols: true
-        upload_source_maps: true
-        project: fixture-project
-        org: fixture-org
-      "
-      `);
+              "sentry:
```
buenaflor, line 12: are these new spacings from auto formatting?

buenaflor (approved):
👍

test/flutter/templates.test.ts
Lms24: honestly not sure. These are generated automatically when running `jest -u` to update the inline snapshots. I think it's fine though because the actual indentation is relative to the first `"`.

PR #897, opened 2025-03-24
test: refactor apple fixture projects to root directory

philprime:
I moved the fixtures/projects used by the apple tests to the root directory, because I can not use `yarn try -i ios` in the Xcode projects directory if it is nested in the e2e app (which has its own package.json defined).

I also added another rather complex project which will be used in another upcoming PR.


Lms24 (approved):
just to confirm: The fixtures that were previously in the e2e test folder are not actual e2e test apps, right? If so, sure, let's move them out of the e2e test folder. 

Have you tried nesting them within the apple test folder? Just wondering if it's not better to keep them close to the test file. But happy to let you make the call on this, so feel free to merge as-is!

philprime:
They were previously in the e2e test folder because all test apps are in there. But there are (currently) no e2e tests for the apple wizard.

As soon as we add e2e tests they will also use these fixture projects, that's why I added them to root. *But* when building tests we should not run them directly in the fixtures directory anyways, and instead copy the project to a temporary path and run tests there. this makes sure that tests are isolated and do not leave a dirty state

Lms24:
sounds good to me, thanks!

Issue #843, opened 2025-03-10
Improve Sentry Example page


matejminar:
### Problems
See below. Basically, it looks trash, and doesn't help users with active ad-blockers. 
![Image](https://github.com/user-attachments/assets/2f360b0b-af88-4fef-a7f7-066d9e0b2409)

### Acceptance Criteria
- A user can navigate to the issue stream specific to the project which will receive the sample errors
- A user can navigate to the docs specific to the SDK they're setting up.
- A user receives a confirmation message, explaining that Sentry successful saw the events.
- A user can understand perhaps why there not seeing the sample errors (Adblocker)
- A user has fun pressing a button

### Designs
https://www.figma.com/design/g3SUGaxJrUVZHXhJHIFS2Y/Specs%3A-Wizard-Example-Page?node-id=0-1&p=f&t=mcsONGca9Vg4t4MA-11

### Links
https://www.joshwcomeau.com/animation/3d-button/

### Docs
We need to document this design change here: https://develop.sentry.dev/sdk/expected-features/setup-wizards/
So that future platforms respect the new design.

PR #899, opened 2025-03-24
feat(next-js): New NextJS example page

ArthurKnaus:
Redesign the example page for NextJS.

Screenshot 2025-03-24 at 12 58 33

Deployed a preview here:
https://next-js-sentry-xi.vercel.app/sentry-example-page

Note:
Once https://github.com/getsentry/sentry-javascript/issues/15780 is done, I will revisit this and show the adblocker warning only when needed.

- part of https://github.com/getsentry/sentry-wizard/issues/843

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- feat(nextjs): New NextJS example page ([#899](https://github.com/getsentry/sentry-wizard/pull/899))
 - feat(telemetry): Add `is_binary` tag to distinguish fossilized binaries ([#857](https://github.com/getsentry/sentry-wizard/pull/857))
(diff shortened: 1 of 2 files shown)


lforst (approved):
👌

lforst:
The black-on-green text for the notification is maybe a bit stark. Do we have a color theme here we could apply?

lforst:
Honestly, now that I see this page, I am not sure whether we really need something like `isProbablyUsingAdBlock()`. The warning is really not intrusive and visible enough for people to see.

ArthurKnaus:
> Honestly, now that I see this page, I am not sure whether we really need something like `isProbablyUsingAdBlock()`. The warning is really not intrusive and visible enough for people to see.

@lforst we think it would be better UX to show an error message and omit the success message in case the SDK is blocked somehow. Showing success + not working is simply confusing (even with disclaimer).

PR #900, opened 2025-03-24
fix: add additional types to xcode.d.ts

philprime:
Added additional types based on analysis of Xcode 16 projects


Lms24 (approved):
Thanks!

PR #901, opened 2025-03-24
feat: add option to define project directory

philprime:
Adds the option `--xcode-project-dir` for the apple wizard to choose a target project directory containing the Xcode project.
This allows to run the wizard from a different directory than the project directory.

I added this option during development of the wizard, therefore it might not be necessary to have this available as a public option, but I do see it as a valuable extension.

Due to the impact of many scripts being relative to `process.cwd()`, I decided to only add it to apple-wizard for now, but we could add adoption for others too.

EDIT:
This option also fixes running the `yarn try` in subdirectories like `fixtures/test-applications/apple/no-target`, because even if you `cd` into the directory, the `process.cwd` will still be in root.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - fix(nextjs): Create root layout for example page if it doesn't exist([#863](https://github.com/getsentry/sentry-wizard/pull/863))
 - ref(utils): Unify `getPackageManger` and `detectPackageManager` ([#865](https://github.com/getsentry/sentry-wizard/pull/865))
 - feat: add option to ignore git changes ([#898](https://github.com/getsentry/sentry-wizard/pull/898))
+- fix(apple): Add additional types to `xcode.d.ts` ([#900](https://github.com/getsentry/sentry-wizard/pull/900))
 - feat: add option to define project directory ([#901](https://github.com/getsentry/sentry-wizard/pull/901))
 
 ## 4.5.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - fix(nextjs): Create root layout for example page if it doesn't exist([#863](https://github.com/getsentry/sentry-wizard/pull/863))
 - ref(utils): Unify `getPackageManger` and `detectPackageManager` ([#865](https://github.com/getsentry/sentry-wizard/pull/865))
 - feat: add option to ignore git changes ([#898](https://github.com/getsentry/sentry-wizard/pull/898))
+- feat: add option to define project directory ([#901](https://github.com/getsentry/sentry-wizard/pull/901))
 
 ## 4.5.0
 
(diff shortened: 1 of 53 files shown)


philprime:
I forgot to mention that I chose the type `cwd: string | undefined` instead of `cwd?: string` so that the compiler will point out which parts of the code need to be changed, also leaving an indicator for further development

bin.ts
```diff
@@ -137,6 +137,12 @@ const argv = yargs(hideBin(process.argv), process.cwd())
       describe: 'Ignore git changes in the project',
       type: 'boolean',
     },
+    'project-dir': {
+      default: undefined,
+      describe:
+        'The directory of your project. Only applies to the Apple wizard.',
+      type: 'string',
+    },
```
philipphofmann, line 145: `m`: I'm unsure adding an option only relevant to the Apple wizard for all platforms makes sense. I would either (1) choose an Apple specific parameter name, such as `xcode-project-dir` or (2) choose a parameter name that works for all platforms, such as `working-directory`, `work-dir`, that specifies the working directory.

philprime: I might have to point out that the `--uninstall` is also a generic name, while it is only used by react-native

philipphofmann: OK, but justifying something by saying we already have it sounds not like the best argument to me. Having this `--uninstall` only for RN was maybe not a great decision. I don't have enough context to make that call. @Lms24 what do you think?

Lms24: I agree that we should try to avoid situations like the `uninstall` flag (which I think we should remove eventually). I don't want to block this change though because if it makes sense for apple (or frankly any platform) I'm happy to support it. 

Here's my suggestion:

> I added this option during development of the wizard, therefore it might not be necessary to have this available as a public option, but I do see it as a valuable extension.

* If you think there's customer-facing value (maybe users asked for it, complained about the lack of this option, etc), let's add it publicly but scope it to something like `xcode-project-dir`
* If there's not much value outside of dev, let's not make it public. We can add `hidden: true` in the `yargs` call and we don't document it anywhere. 

These are the simple solutions. If we want to go more generally, an alternative would be to publicly declare the `working-directory` option but throw a usage error for any wizard where it's used but not supported (which will be the majority of the wizard flows). We can do this of course but I think we should at least know that we solve an actual user pain point :) 

As for the JS-based wizard flows: This rarely came up as a request (maybe once or twice a while ago) and so far we explicitly tell users to run the wizard in the root of their project package. So in a monorepo setup, users would run the wizard _within_ the specific package they want to set up Sentry in. Otherwise the wizard might detect wrong package (manager) information, package managers take the wrong package.json, `node_module` hoisting gets weird, etc.  So it's unlikely that we'll support this globally in all wizard flows.

philprime: Alright, I'll rename it to `--xcode-project-dir` (directory with a subdirectory named `<PROJECT>.xcodeproj`, not mistaken with the path to the Xcode project itself)and I'll hide it using `yargs`.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - feat: add option to ignore git changes ([#898](https://github.com/getsentry/sentry-wizard/pull/898))
 - fix(apple): Add additional types to `xcode.d.ts` ([#900](https://github.com/getsentry/sentry-wizard/pull/900))
 - fix: enable debug logs for option `--debug` ([#902](https://github.com/getsentry/sentry-wizard/pull/902))
+- feat: add option to define project directory ([#901](https://github.com/getsentry/sentry-wizard/pull/901))
```
philipphofmann, line 13: The feature is now internal. Maybe it makes sense to remove the changelog entry or to change the entry to reflect that the flag is internal.

philipphofmann (approved):
LGTM

PR #904, opened 2025-03-24
fix(apple): add support for synchronized Xcode folders

### LINKED ISSUE #818: Add support for folder-based Xcode projects
The Xcode project file format changed and the detection of the file to inject the code snippet fails.

This is most likely due to the `xcodes` dependency not being actively maintained. We should test this and implement proper parsing to fix the snippet injection.

# Steps to reproduce

- Create a sample Xcode project named something like `this-is-an-app`
- Run the sentry-wizard to setup the project
- Note this log message:

```
Added the Sentry dependency to your project but could not add the Sentry code snippet.
```

philprime:
Adds support for injecting code into Xcode projects using synchronized folders

Closes #818

## Summary how Xcode Projects are structured:
- Every Xcode Project has exactly one main group of type `PBXGroup`
- The main group contains a list of children identifiers
- Each child can be a `PBXGroup`, a `PBXFileReference` or a `PBXFileSystemSynchronizedRootGroup`
- Each `PBXGroup` has a list of children identifiers which again can be `PBXGroup`, `PBXFileReference` or `PBXFileSystemSynchronizedRootGroup`
- The target defines the list of `fileSystemSynchronizedGroups` which are `PBXFileSystemSynchronizedRootGroup` to be included in the build phase
- The `PBXFileSystemSynchronizedRootGroup` has a list of `membershipExceptions` which are files to be excluded from the build
- The Xcode project has a build phase `PBXSourcesBuildPhase` which has a list of `files` which are `PBXBuildFile`
- A file which is not part of a `PBXFileSystemSynchronizedRootGroup` must be added to the `files` list of the `PBXSourcesBuildPhase` build phase
- Nested subfolders in `fileSystemSynchronizedGroups` are not declared but recursively included

Based on the findings above the files included in the build phase are:
- All files in the `files` of the `PBXSourcesBuildPhase` build phase `Sources` of the target
- All files in directories of the `fileSystemSynchronizedGroups` of the target
- Excluding all files in the `membershipExceptions` of the `fileSystemSynchronizedGroups` of the target


sentry-wizard.code-workspace
```diff
@@ -0,0 +1,10 @@
+{
```
philprime, line 1: @Lms24 I am using a VSCode/Cursor workspace locally, so I can run E2E tests from the UI. Should we keep this in the PR or scrap it?

philprime: Thanks for the feedback!

> We usually don't add too much IDE configuration to the repo since not everyone uses vscode.

Makes sense to me, I'll remove it from the PR.

src/apple/xcode-manager.ts
```diff
@@ -6,11 +6,13 @@
 import clack from '@clack/prompts';
 import * as fs from 'node:fs';
 import * as path from 'node:path';
+import { debug } from '../utils/debug';
 import type { SentryProjectData } from '../utils/types';
 import * as templates from './templates';
 
 import {
   project as createXcodeProject,
+  PBXFileSystemSynchronizedRootGroup,
```
krystofwoldrich, line 15:
```suggestion
  type PBXFileSystemSynchronizedRootGroup,
```

src/apple/xcode-manager.ts
```diff
@@ -305,100 +310,296 @@ export class XcodeProject {
...
+    // - The target defines the list of `fileSystemSynchronizedGroups` which are `PBXFileSystemSynchronizedRootGroup` to be included in the build phase
+    // - The `PBXFileSystemSynchronizedRootGroup` has a list of `membershipExceptions` which are files to be excluded from the build
+    // - The Xcode project has a build phase `PBXSourcesBuildPhase` which has a list of `files` which are `PBXBuildFile`
+    // - A file which is not part of a `PBXFileSystemSynchronizedRootGroup` must be added to the `files` list of the `PBXSourcesBuildPhase` build phase
+    // - Nested subfolders in `fileSystemSynchronizedGroups` are not declared but recursively included
+    //
+    // Based on the findings above the files included in the build phase are:
+    // - All files in the `files` of the `PBXSourcesBuildPhase` build phase `Sources` of the target
+    // - All files in directories of the `fileSystemSynchronizedGroups` of the target
+    // - Excluding all files in the `membershipExceptions` of the `fileSystemSynchronizedGroups` of the target
```
krystofwoldrich, line 335: I've not seen any code related to excluding the files is that on purpose?

krystofwoldrich:
Besides the one comment, it looks good.

Thanks for the extensive tests.

philipphofmann (commented):
Looking good. I added a few comments. Almost LGTM.

src/apple/xcode-manager.ts
```diff
@@ -305,100 +310,296 @@ export class XcodeProject {
...
+      if (fs.statSync(filePath).isDirectory()) {
+        result.push(...this.getFilesInDirectoryTree(filePath));
+        continue;
+      }
+      // Ignore hidden files
+      if (file.startsWith('.')) {
+        continue;
+      }
+      // If the file is a file, add it to the result
+      result.push({
```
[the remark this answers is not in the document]
philprime: What sort of file could be neither directory nor a file? I believe file references are also considered files

src/apple/macos-system-helper.ts
```diff
@@ -0,0 +1,28 @@
+import { execSync } from 'child_process';
+import { debug } from '../utils/debug';
+
+export class MacOSSystemHelpers {
```
philprime, line 4: @Lms24 this utility class is used to resolve paths in Xcode projects. The commands `xcrun` and `xcode-select` come with the installation of the Xcode Command Line Utilities, therefore for Apple platforms developers they will be available.

How should we handle the usage in unit tests when they run on ubuntu in CI?
Should we create fakes for the CI workflows so that `execSync` is actually run, or should we just mock these functions in the javascript tests?

philprime: Additional option could be skipping tests that are supposed to run on macOS and create another job with `macos-latest` as the `runs-on`

src/apple/xcode-manager.ts
```diff
@@ -305,100 +310,296 @@ export class XcodeProject {
...
+          path: path.join(groupPath, name),
         });
-      } else if (groupReference) {
-        if (typeof groupReference === 'string') {
-          continue;
+        continue;
+      }
+
+      const group = this.objects.PBXGroup?.[child.value];
+      if (group && typeof group !== 'string') {
```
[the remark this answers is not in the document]
philprime: There are no types, the group is either an object with key-value pairs or it's a string for the comment. This is a weird side-effect of the `xcode` parser dependency

src/apple/xcode-manager.ts
```diff
@@ -305,100 +310,296 @@ export class XcodeProject {
...
+      const filePath = path.join(dirPath, file);
+      // If the file is a directory, recursively get the files in the directory
+      if (fs.statSync(filePath).isDirectory()) {
+        result.push(...this.getFilesInDirectoryTree(filePath));
+        continue;
+      }
+      // Ignore hidden files
+      if (file.startsWith('.')) {
+        continue;
+      }
```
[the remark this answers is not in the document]
philprime: That is a good point, we probably would want that. I'll move the statement up.

src/apple/xcode-manager.ts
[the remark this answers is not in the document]
philprime: Nevermind, looking at the available methods I can see there are multiple other cases:

```
isFile(): boolean;
isDirectory(): boolean;
isBlockDevice(): boolean;
isCharacterDevice(): boolean;
isSymbolicLink(): boolean;
isFIFO(): boolean;
isSocket(): boolean;
```

I'll add a check

src/apple/macos-system-helper.ts
philipphofmann: You could use fakes in the unit tests for Ubuntu and write some extra integration tests running on MacOS runners. Whatever we do, we should add some tests running on MacOS to ensure this code works correctly.

src/apple/xcode-manager.ts
```diff
@@ -302,103 +331,971 @@ export class XcodeProject {
...
+   *
+   * @param targetName - The name of the target to find
+   * @returns The native target, or undefined if the target is not found
+   */
+  private findNativeTargetByName(
+    targetName: string,
+  ): XcodeProjectObjectWithId<PBXNativeTarget> | undefined {
+    debug('Finding native target by name: ' + targetName);
+
+    const nativeTargets = Object.entries(this.objects.PBXNativeTarget ?? {});
```
philipphofmann, line 395: `l`: We could add an early return if the `nativeTargets` are empty and include a log a debug message.

src/apple/xcode-manager.ts
```diff
@@ -302,103 +331,971 @@ export class XcodeProject {
...
+
+      const buildFileRefId = buildFileObj.fileRef;
+      if (!buildFileRefId) {
+        debug(`File reference not found for file: ${file.value}`);
+        continue;
+      }
+      debug(`Build file reference found for file: ${file.value}`);
+
+      // Find the related file reference object
+      const buildFile = this.objects.PBXFileReference?.[buildFileRefId];
```
philipphofmann, line 508: `l`: When the `this.objects.PBXFileReference` is `undefined`, we iterate over the buildPhaseFiles with no reason and print log messages. Instead, we could add a check before the loop and an early return.

src/apple/xcode-manager.ts
```diff
@@ -302,103 +331,971 @@ export class XcodeProject {
...
+  private resolveAbsoluteFilePathRelativeToSdkRoot(
+    buildFile: XcodeProjectObjectWithId<PBXFileReference>,
+  ): string | undefined {
+    debug(
+      `Resolving absolute file path relative to SDK root for file reference: ${
+        buildFile.id
+      } with path: ${buildFile.obj.path ?? ''}`,
+    );
+
+    const sdkRoot = MacOSSystemHelpers.findSDKRootDirectoryPath();
```
philipphofmann, line 678: `m`: We might need to inject the `MacOSSystemHelpers` as a dependency so we can run the unit tests on Ubuntu.

src/apple/xcode-manager.ts
```diff
@@ -302,103 +331,971 @@ export class XcodeProject {
...
+        settings['CONFIGURATION_BUILD_DIR'] ??
+        settings['TARGET_BUILD_DIR'] ??
+        settings['BUILD_DIR'];
+
+      return this.buildProductsDir;
+    } catch (error) {
+      debug(`Failed to find build products directory path: ${error as string}`);
+      return undefined;
+    }
+  }
```
philipphofmann, line 1300: `m`: This looks similar to the code in `macos-system-helper.ts`. Shouldn't we also move it there?

src/apple/xcode-manager.ts
```diff
@@ -255,18 +264,38 @@ function addUploadSymbolsScript(
 }
 
 export class XcodeProject {
```
philipphofmann, line 266: `m`: This class is huge. Can we maybe break it up into smaller chunks?

philipphofmann (commented):
Almost LGTM

src/apple/macos-system-helper.ts
Lms24: I was just thinking that it might make sense to run the tests in question just on a macos runner then? We can also exclude the unit tests in the general unit test job and executed the specific ones in another job running on macos.

In regards to mocking/if we don't go down the macos runner route: Feel free to do whatever is necessary. No strong opinions. We usually make use of `vi.mock`/`vi.spyOn` for stuff like this, so it's not unheard of 😆

philprime: To combine your thoughts I'll do this:

- Create a second GHA job to run test suite on macOS too.
- For Linux I'll use fakes so that the entire test suite passes, no need to manage multiple sets of test suites.
- For macOS I'll run the entire test suite, also to ensure we are cross-platform-compatible
- Instead of creating mock CLIs for Linux, I'll use `vi.mock` / `vi.spyOn` to mock `execSync` or similar.
- I'll evaluate making the macOS system helper class an injectable dependency.

src/apple/xcode-manager.ts
philprime: I'll explore this, because due to the nature of JavaScript we can also mock/spy patch them

src/apple/xcode-manager.ts
philprime: Yes it makes sense to move the `xcodebuild` execution there too. Thanks for pointing this out.

src/apple/xcode-manager.ts
philprime: I agree it is massive, I might be able to split it up into reasonable parser/reader utility classes, but in general it is just a wrapper around the `XcodeProject` type

philipphofmann: I leave this up to you. No strong opinion.

philprime: I'll refactor it in another PR

src/apple/macos-system-helper.ts
philipphofmann: @philprime, do you plan to do this in this PR, or is there a way to make this happen in follow-up PRs? I would really appreciate getting this PR merged as it's already huge.

src/apple/macos-system-helper.ts
```diff
@@ -0,0 +1,64 @@
...
+        encoding: 'utf8',
+      }).trim();
+      return developerPath;
+    } catch (error) {
+      debug(`Failed to find developer directory path: ${error as string}`);
+      return undefined;
+    }
+  }
+
+  static readXcodeBuildSettings(
```
philipphofmann, line 29: Thanks for moving this here.

philipphofmann (commented):
@philprime, how can we get this PR merged quickly? It's enormous and I would like to get it off my plate.

src/apple/macos-system-helper.ts
philprime: I need to do this in this PR because the tests can not pass in CI otherwise.

philprime:
@philipphofmann only thing left to do is getting the test suite to work in CI for both ubuntu and macos

philipphofmann:
> @philipphofmann only thing left to do is getting the test suite to work in CI for both ubuntu and macos

OK thanks.

src/apple/xcode-manager.ts
philprime: For now it can just be mocked in tests, no need to make it replaceable.

.github/workflows/build.yml
```diff
@@ -95,8 +97,31 @@ jobs:
...
       - name: Install dependencies with yarn
         run: yarn install --frozen-lockfile
+
+      - # The output of `xcode-select --print-path` is something like `/Applications/Xcode-16.3.0.app/Contents/Developer`
+        # We want to extract the path to the Xcode.app, which is `/Applications/Xcode-16.3.0.app`
+        name: Find path to Xcode.app
+        id: find-xcode-app-path
+        if: startsWith(matrix.os, 'macos-')
+        run: |
+          XCODE_APP_PATH=$(xcode-select --print-path | sed 's|/Contents/Developer||')
```
philprime, line 107: @philipphofmann I am stuck here. For some reason it keeps using the wrong SDK directory:

Expected: "/Applications/Xcode_16.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk"
Received: "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk"

Do you have any ideas?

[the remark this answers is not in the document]
philprime: Sorry, my bad, I wasn't clear with my point.
The resolving of the path to the Xcode.app works fine and is set correctly here:

https://github.com/getsentry/sentry-wizard/actions/runs/14497903478/job/40670226192#step:7:5

The actual issue is that the unit test for `findSDKRootDirectoryPath()` is failing, which uses `xcrun --show-sdk-path`. Locally the xcrun command resolves to this:

```
/Applications/Xcode-16.3.0.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
```

but in CI it resolves to the command line tools at:

```
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```

I think the best attempt right now is that I am going to inject the full path to the SDK into env, instead of building the path relative to the Xcode.app

src/apple/macos-system-helper.ts
```diff
@@ -0,0 +1,64 @@
+import { execSync } from 'child_process';
+import { debug } from '../utils/debug';
+
+export class MacOSSystemHelpers {
+  static findSDKRootDirectoryPath(): string | undefined {
+    try {
+      const sdkPath = execSync('xcrun --show-sdk-path', {
+        encoding: 'utf8',
+      }).trim();
```
philprime, line 9: In CI this resolves to the Command Line Tools:

```
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```

Locally it resolves to the Xcode.app:

```
/Applications/Xcode-16.3.0.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
```

.github/workflows/build.yml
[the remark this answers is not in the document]
philipphofmann: Ah OK, hmm, but when it fails in CI, isn't there also a risk that it will fail for others locally as well? Maybe the approach in `findSDKRootDirectoryPath` is wrong.

test/apple/macos-system-helper.test.ts
```diff
@@ -0,0 +1,163 @@
...
+          projectWithSynchronizedFolders,
+        );
+
+        // -- Assert --
+        // The build settings are a massive list of key-value pairs, so we'll just check a few of them
+        // which are relevant for our use cases.
+        expect(buildSettings?.['CONFIGURATION_BUILD_DIR']).toEqual(
+          path.join(
+            appleProjectsPath,
+            'xcode-15-compatible-project/build/Release-unknown',
```
philipphofmann, line 134: `l`: Can't this break easily when it relies on xcode-15?

test/apple/macos-system-helper.test.ts
```diff
@@ -0,0 +1,163 @@
...
+            encoding: 'utf8',
+          }).trim();
+        }
+
+        // -- Act --
+        const developerDirectoryPath =
+          MacOSSystemHelpers.findDeveloperDirectoryPath();
+
+        // -- Assert --
+        expect(developerDirectoryPath).toBe(xcodeAppPath);
```
philipphofmann, line 100: `m`: If I'm not mistaken, it's not a problem that the `MacOSSystemHelpers.findDeveloperDirectoryPath` returns a different result in CI, as pointed out here: https://github.com/getsentry/sentry-wizard/pull/904#discussion_r2048733766. So instead of passing in the path via a env variable, wouldn't it be sufficient to adapt the assertion to ensure it's either the local variation or the CI variation?

philipphofmann (commented):
The env via CI for the tests looks a bit weird to me.

test/apple/macos-system-helper.test.ts
philprime: The project directory is called `xcode-15-compatible-project`. It's just a project where I hard-coded the object version number

test/apple/macos-system-helper.test.ts
philprime: Yes, we can also just check for any variation, but if the CI setup ever changes i.e. due to a new version of the github actions runner image, it can break a previously green commit. My intention here is injecting the exact path we are expecting depending on the CI, but I also don't like this

philprime: I'll actually go for this approach and remove the injection for now. If it becomes a problem, we'll have to revisit

.github/workflows/build.yml
```diff
@@ -95,6 +97,15 @@ jobs:
...
         run: yarn install --frozen-lockfile
+      - name: Print Xcode environment
+        if: startsWith(matrix.os, 'macos-')
+        run: |
+          echo "Xcode Path: $(xcode-select --print-path)"
+          echo "SDK Path: $(xcrun --show-sdk-path)"
+          echo "SDK Version: $(xcrun --show-sdk-version)"
+          echo "SDK Build Version: $(xcrun --show-sdk-build-version)"
+          echo "SDK Platform Path: $(xcrun --show-sdk-platform-path)"
+          echo "SDK Platform Version: $(xcrun --show-sdk-platform-version)"
```
Lms24, line 108: l: just checking, is this intentionally kept in the PR or was it for testing CI? No problem if you wanna keep it

Lms24 (approved):
As discussed via DM, I didn't check the apple-specific code but the CI changes look fine to me!

.github/workflows/build.yml
philprime: As these paths could change over time due to update of the CI runner image, I would like to keep them in, so in case our tests are failing, we can directly compare the expected and actual paths.

I'll add this as a comment

PR #903, opened 2025-03-24
style(apple): remove unused eslint-disable declarations

philprime:
- refactors the rather large `apple-wizard.ts` into multiple smaller files, representing individual steps
- telemetry tracing of steps is moved to each file

Split into multiple smaller PRs:

- [x] #906
- [x] #907
- [x] #908
- [x] #909
- [x] #910
- [x] #911
- [x] #912 

After all the sub-PRs are merged, only the clean up of imports and linting rules should remain

#skip-changelog

PR #906, opened 2025-03-25
refactor(apple): move check-installed-cli to separate file

philprime:
Derived from #903 for easier review.

This is one of multiple commits to refactor the apple wizard into individual steps with the goal of improving maintainability and testability further down the line.

#skip-changelog


philipphofmann (approved):
LGTM

PR #913, opened 2025-03-26
feat(sveltekit): New example page

ArthurKnaus:
Redesign the example page for Sveltekit

https://github.com/user-attachments/assets/344750e8-58cf-43dd-8392-e31666cf2daa

Note:
Once https://github.com/getsentry/sentry-javascript/issues/15780 is done, I will revisit this and show the adblocker warning only when needed.

- part of https://github.com/getsentry/sentry-wizard/issues/843

### DIFF
--- a/src/sveltekit/templates.ts
+++ b/src/sveltekit/templates.ts
@@ -90,6 +90,8 @@ Feel free to delete this file and the entire sentry route.
 
 <script>
   import * as Sentry from '@sentry/sveltekit';
+  
+  let hasSentError = false;
 
   function getSentryData() {
     Sentry.startSpan(
@@ -100,6 +102,7 @@ Feel free to delete this file and the entire sentry route.
       async () => {
         const res = await fetch('/sentry-example');
         if (!res.ok) {
+          hasSentError = true;
           throw new Error('Sentry Example Frontend Error');
         }
       }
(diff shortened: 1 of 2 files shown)


src/sveltekit/templates.ts
```diff
@@ -100,75 +102,160 @@ Feel free to delete this file and the entire sentry route.
...
+      background-color: #7553FF;
+      border: 1px solid #553DB8;
+      transform: translateY(-4px);
+    }
+
+    &:hover > span{
+      transform: translateY(-8px);
+    }
+
+    &:active > span{
```
s1gr1d, line 226:
```suggestion
    &:hover > span {
      transform: translateY(-8px);
    }

    &:active > span {
```

src/sveltekit/templates.ts
```diff
@@ -90,6 +90,8 @@ Feel free to delete this file and the entire sentry route.
 
 <script>
   import * as Sentry from '@sentry/sveltekit';
+  
+  let hasSentError = false;
```
s1gr1d, line 94: I'm not sure how important this is on an error page but we could add the syntax for Svelte Runes here if people prefer that.

```suggestion
  // Svelte Runes (requires Svelte 5)
  // let hasSentError = $state(false);
  let hasSentError = false;
```

s1gr1d (approved):
Thanks for adding this, it looks awesome 🙌

src/sveltekit/templates.ts
ArthurKnaus: nice! TIL

PR #915, opened 2025-03-26
feat(react-native): Add a session replay step for react-native

smeubank:
Add a step asking user if they want to add session replay, and then inject it if possible, if not it will print the snippet with new options to the terminal for htem to copy paste

Screenshot 2025-04-02 at 2 21 41 PM


src/react-native/react-native-wizard.ts
```diff
@@ -209,8 +214,48 @@ Or setup using ${chalk.cyan(
...
+    Sentry.setTag('enable-session-replay', enableSessionReplay);
+
+    if (enableSessionReplay) {
+      clack.log.info(
+        'Session Replay will be enabled with default settings (replaysSessionSampleRate: 0.1, replaysOnErrorSampleRate: 1.0).',
+      );
+      clack.log.info(
+        'By default, all text content, images, and webviews will be masked for privacy. You can customize this in your code later.',
+      );
+    }
```
antonis, line 244: l: This is just a suggestion. Wdyt of moving the two `log.info` in the `javascript.ts` were the configuration is injected? This way they won't be printed if `addSentryInit` fails with an error.

antonis (commented):
Thank you for your work on this @smeubank 🙇 
I understand that the PR is still draft but I had an early look at it. The code changes LGTM and worked as expected in my tests injecting the desired code.
I think that with the addition of tests verifying the new code this will be ready to go.

src/react-native/react-native-wizard.ts
smeubank: @antonis I was honestly mainly playing around with Cursor based on my understanding of how the steps are built in the wizard for different platforms. I am open to all suggestions :D

smeubank:
> I think that with the addition of tests verifying the new code this will be ready to go.
would you mind taking the PR over, and making some of those changes? I will be at a conference for the rest of the week.

src/react-native/react-native-wizard.ts
antonis: Updated the code with 5fd61dc1ae2a4fcbeccebe16a41165f3e0f61a55 moving the logs

antonis (commented):
>> would you mind taking the PR over, and making some of those changes? I will be at a conference for the rest of the week.
>
> I'll be happy to :) Enjoy the conference!

Heads up that I've updated the code and added a test case for the changes.

@krystofwoldrich I'm adding you as a reviewer for a 2nd pair of eyes on my changes :)

src/react-native/javascript.ts
```diff
@@ -46,8 +55,20 @@ export async function addSentryInit({ dsn }: { dsn: string }) {
     return;
   }
 
+  if (enableSessionReplay) {
+    clack.log.info(
+      `Session Replay will be enabled with default settings (replaysSessionSampleRate: ${sessionReplaySampleRate}, replaysOnErrorSampleRate: ${sessionReplayOnErrorSampleRate}).`,
+    );
+    clack.log.info(
```
krystofwoldrich, line 62: L: Doesn't print the blue point before the text, which will make it look more like part of the previous info log. 

````suggestion
    clack.log.message(
````

File: src/react-native/react-native-wizard.ts [hunk 3]
```diff
@@ -209,8 +214,39 @@ Or setup using ${chalk.cyan(
...
+          'Do you want to enable Session Replay to help debug issues? (See https://docs.sentry.io/platforms/react-native/session-replay/)',
+      }),
+    );
+
+    Sentry.setTag('enable-session-replay', enableSessionReplay);
+  } else if (sdkVersion) {
+    clack.log.info(
+      `Session Replay is supported from Sentry React Native SDK version ${SESSION_REPLAY_SUPPORTED_SDK_RANGE}. Your version (${sdkVersion}) doesn't support it yet.`,
+    );
+    clack.log.info(
```
krystofwoldrich, line 240: L:

````suggestion
    clack.log.message(
````

krystofwoldrich (approved):
LGTM 🚀 

Just left one nit about the two blue points (two info logs, right after each other).

PR #926, opened 2025-03-27
test: Switch from `jest` to `vitest`

Lms24:
This PR replaces all jest usage in the wizard repo with vitest. The main motivation is performance but also being more up to date, having better timer mocks and better ESM compatibility.

### Concrete changes
- replace unit jest tests with vitest
- replace e2e test jest tests with vitest
- no longer use jest/vitest globals but explicitly import vitest APIs
- replace `eslint-plugin-jest` with `eslint-plugin-vitest`. This is no longer the most up to date version (which would be `@vitest/eslint-plugin`) but it still works with Eslint 8. If we upgrade to Eslint 9 (with its new config structure) we can also switch to this new plugin. 
- also resolves the same eslint errors as #924 (let's see what gets merged first)

### Perf improvements in CI 

- Unit tests: ~45s -> ~30s
- E2E tests: 1m30s -> ~1m

So not super impressive but nice!

### Why the Drop in Coverage??

If you take a look at the codecov report, we can see a coverage drop from ~52 to ~24%. This doesn't look great, so what happened?

Two things:

1. Jest didn't calculate coverage for _all_ files but only for the ones where at least one statement was actually executed. We seem to have a lot of files where nothing is executed (i.e. completely untested files). Vitest includes these by default. We _can_ disable this by setting `coverage.all: false` in the vitest options but IMHO this delivers a false understanding of coverage. If I set `all: false`, overall line coverage increases to 45% which is still lower but there's also a second reason:

2. Jest counted multi-line statements only once. So if you had something like
  ```js
  return ( isTs ?
    'const a: number;' :
    'const a'
  );  
  ```
  jest would mark line 1 as (un)covered and not report anything for the rest. Vitest however marks lines 1-4 as (un)covered. Depending of the ratio of overall covered/uncovered lines per file, this change significantly influences the coverage score of the file.

**Conclusion**: As harsh as it sounds but the drop in coverage is because jest's coverage config was very "optimistic". The new number unfortunately shows the true state of coverage.

Also worth noting: For both, jest and vitest, we don't collect any meaningful coverage in our e2e tests. This is because we execute the actual wizard code in a child process where coverage isn't reported. So there's hardly any value in uploading these "empty" coverage report to codecov (but it doesn't harm us either)

Final remark: Actually, neither Jest nor Vitest measure coverage themselves but they use different tools for that. Jest uses babel by default and vitest v8. This is likely the reason for 2.

#skip-changelog


andreiborza (approved):
🚀 nice!

PR #928, opened 2025-03-28
Move deno package-manager to unreleased

joshdales:
The `deno` was not included in the 4.6.0 release, move it to unreleased

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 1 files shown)


joshdales:
@s1gr1d Looks like my PR just missed the release but is now in the changelog for 4.6.0. Just moving it to the unreleased

Lms24 (approved):

Lms24:
Good catch, thanks! We'll release another version early next week.

PR #929, opened 2025-03-28
ref: Hide e2e test args in `--help` command

Lms24:
We have a bunch of `preSelectedProject.*` CLI args which are only meant to bypass authentication and project selection in our e2e tests. These are not meant to be user-facing, so this PR removes them from the output of out `--help` command.

This also adds a snapshot test against the `--help` output, so that we're aware if something changes (and think twice if e.g. a newly added option should be user-facing.

#skip-changelog

### DIFF
--- a/e2e-tests/tests/help-message.test.ts
+++ b/e2e-tests/tests/help-message.test.ts
@@ -1,0 +1,56 @@
+import { execSync } from 'node:child_process';
+import { join } from 'node:path';
+
+describe('--help command', () => {
+  it('prints the up to date help command', () => {
+    const binName = process.env.SENTRY_WIZARD_E2E_TEST_BIN
+      ? ['dist-bin', `sentry-wizard-${process.platform}-${process.arch}`]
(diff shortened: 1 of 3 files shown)


e2e-tests/tests/help-message.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      ? ['dist-bin', `sentry-wizard-${process.platform}-${process.arch}`]
+      : ['dist', 'bin.js'];
+
+    const binPath = join(__dirname, '..', '..', ...binName);
+
+    const rv = execSync(`${binPath} --help`, {
+      stdio: 'pipe',
+    });
+
+    expect(rv.toString()).toMatchInlineSnapshot(`
```
Lms24, line 16:
```suggestion
    const output = execSync(`${binPath} --help`, {
      stdio: 'pipe',
    });

    expect(output.toString()).toMatchInlineSnapshot(`
```

PR #933, opened 2025-04-01
feat(profiling): convert to continuous profiling v2

armcknight:
Convert the snippets to the new continuous profiling v2 api.

For https://github.com/getsentry/sentry-cocoa/issues/4996


src/apple/templates.ts
```diff
@@ -22,11 +22,13 @@ export function getSwiftSnippet(dsn: string): string {
             options.debug = true // Enabled debug when first installing is always helpful
             // Set tracesSampleRate to 1.0 to capture 100% of transactions for performance monitoring.
             // We recommend adjusting this value in production.
-            options.tracesSampleRate = 1.0
```
philprime, line 25: `m`: is there a specific reason why setting it to `1` instead of `1.0`? If yes, please also apply to the comments above to keep it uniform.
This goes for all changes of this PR

armcknight: I set it to the most idiomatic swift/objc examples: in objc, `1.0` is a `double` and `1.0f` is a `float`, which can be shortened to `1.f`. In swift, `1` is implicitly coerced to a `float`.

I deliberately left it as 1.0 in the docs since that's the most precise human language representation of the decimal value.

philprime: IMHO while it might be technically correct, it feels like over complication without much additional value to have different styles in such a small code sample.

I'll leave it to you

philprime (approved):
LGTM

src/apple/templates.ts
armcknight: I might even just change the docs to say "set `tracesSampleRate` to capture 100%`... and change the obj version to `@1` as well

armcknight: OK, I just changed it back to 1.0 everywhere.

PR #938, opened 2025-04-07
feat(nextjs): Add `onRouterTransitionStart` to client instrumentation template

lforst:
This is necessary for routing instrumentation to work.

### DIFF
--- a/test/nextjs/templates.test.ts
+++ b/test/nextjs/templates.test.ts
@@ -42,7 +42,9 @@ describe('Next.js code templates', () => {
 
           // Setting this option to true will print useful information to the console while you're setting up Sentry.
           debug: false,
-        });"
+        });
+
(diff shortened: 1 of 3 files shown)


test/nextjs/templates.test.ts
```diff
@@ -42,7 +42,9 @@ describe('Next.js code templates', () => {
 
           // Setting this option to true will print useful information to the console while you're setting up Sentry.
           debug: false,
-        });"
+        });
+
+        export const onRouterTransitionStart = Sentry.captureRouterTransitionStart;"
```
Lms24, line 47: Is this something that users could already have defined prior to the wizard setting it? Not saying this needs to be handled in this PR but maybe somthing to keep in mind/follow up on(?)

lforst: Yes, although highly unlikely.

Lms24 (approved):

PR #941, opened 2025-04-11
telemetry: Add basic telemetry for electron and cordova wizards

Lms24:
This PR adds very basic telemetry collection to Electron and Cordova wizards. This doesn't add most of the tags we collect in other wizards but it should be good enough to at least get an idea of usage of these two wizards. Which is the main thing we're interested in.

#skip-changelog


timfish (approved):

lforst (approved):

kahest (approved):
thanks!

PR #949, opened 2025-04-15
fix(remix): linting issues in generated client init code

constantinius:
Closes https://linear.app/getsentry/issue/TET-74/[remix]-linting-errors-in-default-project-after-wizard

### DIFF
--- a/test/remix/client-entry.test.ts
+++ b/test/remix/client-entry.test.ts
@@ -28,17 +28,17 @@ describe('initializeSentryOnEntryClient', () => {
         useMatches,
       } from "@remix-run/react";
 
-      import * as Sentry from "@sentry/remix";
+      import * as Sentry, {  replayIntegration, browserTracingIntegration,} from "@sentry/remix";
 
       Sentry.init({
           dsn: "https://sentry.io/123",
           tracesSampleRate: 1,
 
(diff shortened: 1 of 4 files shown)


test/remix/client-entry.test.ts
```diff
@@ -106,13 +106,13 @@ describe('initializeSentryOnEntryClient', () => {
         useMatches,
       } from "@remix-run/react";
 
-      import * as Sentry from "@sentry/remix";
+      import * as Sentry, {  browserTracingIntegration,} from "@sentry/remix";
```
s1gr1d, line 109:
```suggestion
      import { browserTracingIntegration } from "@sentry/remix";
```

lforst:
I advise against this. This is due to an upstream issue, see https://github.com/import-js/eslint-plugin-import/issues/2969

lforst:
Maybe it's fine idk

s1gr1d (approved):
Just for some additional info around this: There is a bug in the eslint import/namespace package and that's the reason why we even have those warnings here. https://github.com/getsentry/sentry-wizard/pull/949

However, as users are confronted with this warning right after they install Sentry I think importing it directly (not via the namespace) is a good workaround.

PR #950, opened 2025-04-15
ref: Extract `git` utility functions

Lms24:
This PR 

- extracts `isInGitRepo` and `getUncommittedOrUntrackedFiles` from clack utils to a separate file
- adds tests for the two functions

UPDATE: Originally, this PR also proposed a change to the typing of the `cwd` param in some function. I removed this part (see review comments)

#skip-changelog

### DIFF
--- a/src/utils/clack/index.ts
+++ b/src/utils/clack/index.ts
@@ -183,7 +183,7 @@ You can turn this off at any time by running ${chalk.cyanBright(
  */
 export async function confirmContinueIfNoOrDirtyGitRepo(options: {
   ignoreGitChanges: boolean | undefined;
-  cwd?: string | undefined;
+  cwd: string | undefined;
 }): Promise<void> {
   return traceStep('check-git-status', async () => {
     if (
@@ -718,7 +718,7 @@ async function addCliConfigFileToGitIgnore(filename: string): Promise<void> {
  *
  * @param opts.cwd The directory of the project. If undefined, the current process working directory will be used.
  */
-export async function runPrettierIfInstalled(opts?: {
+export async function runPrettierIfInstalled(opts: {
   cwd: string | undefined;
 }): Promise<void> {
   const cwd = opts?.cwd;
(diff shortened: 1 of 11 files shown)


src/utils/clack/index.ts
```diff
@@ -183,7 +183,7 @@ You can turn this off at any time by running ${chalk.cyanBright(
  */
 export async function confirmContinueIfNoOrDirtyGitRepo(options: {
   ignoreGitChanges: boolean | undefined;
-  cwd: string | undefined;
+  cwd?: string | undefined;
```
philprime, line 186: `m`: Please see this comment why I explicitly changed it to `string | undefined` and not made it optional

https://github.com/getsentry/sentry-wizard/pull/901#issuecomment-2748353716

philprime: I do not have a strong opinion on this, if you prefer to remove it, go ahead.

Lms24: ah I see, sorry I missed this. Yeah no strong opinion either, let's leave it as-is then. I'll remove the `cwd` refactor from the PR and go with just the extraction. 

fwiw, this came up when I was rebasing and picking up work on the Angular wizard (#672) and I was surprised why I should add `cwd: undefined` to the two functions... I'm not sure I like this pattern in the long-run because to me it entails that I _should_ actually do something else than adding `undefined`. In reality, this would only matter if the wizard would support setting a special cwd (which requires quite a few refactors across the board...).

philprime: IMHO the wizard should assume the current directory as a cwd but allow customization for the target path in any case. It was required for the apple-wizard, and it was the simplest way of making everyone else notice it.

Lms24: ok, let's tackle this separately but if we want to go down this route more generally, your original typing makes more sense anyway. Will revert the changes in the next commit.

philprime (approved):
LGTM

PR #951, opened 2025-04-15
feat(nextjs): add connectivity check to example page

shellmayr:
- Uses the new diagnoseSdkConnectivity() function to check whether something is preventing data from reaching sentry https://github.com/getsentry/sentry-javascript/pull/15821
- If the connection is disturbed, call this out to the user using a banner and link to the docs explaining how to solve the issue
- Contributes to [TET-59](https://linear.app/getsentry/issue/TET-59/add-error-state-to-example-pages)

Before: 
Screenshot 2025-04-15 at 14 47 49

After:
Screenshot 2025-04-16 at 09 21 45


lforst (commented):
beaut

(random thought: some applications put the Next.js devserver floaty thingy in the bottom right corner so idk whether the bottom right corner is a good place for the alert box. If it's easy I would put it in the center of the page with the other content but no biggie if that is too hard.)

lforst (approved):

matejminar (approved):
Looks good!  
Nit: Can we slightly tweak the copy so that we take up only 3 lines? That one word in the fourth line makes it more massive than it needs to be.

PR #944, opened 2025-04-14
feat(nextjs): make errors on example page more useful

shellmayr:
- Add error classes for the example errors
- Add some context that one is raised on the frontend, and one on the backend to educate about full stack support

Before:
Screenshot 2025-04-14 at 14 17 09

After:
Screenshot 2025-04-14 at 14 15 22

PR #953, opened 2025-04-15
feat(angular): Add example component

Lms24:
This PR adds an example component (inspired by the Nuxt wizard) to the WIP Angular wizard. 

In Angular, we can't easily add a new page because the routes are declared in code (as opposed to files like in NextJs/SvelteKit/etc). Finding and injecting our component there would be quite a task, so instead, we just emit a component file and tell users how to add it to one of their pages/components.

Flow:
1. wizard asks users if they want to add the component
2. by default, it will add a new file into `./src/app` (which is the default app root folder in a conventional Angular app)
3. if this folder is not present, wizard asks users to specify a path
4. wizard writes the example component file
5. wizard prompts users via a copy/paste snippet to add the component to one of their existing components

Other stuff to note:
- The component is a standalone component which should be supported by all Angular versions the wizard supports (NG17 and newer). This means we don't have to worry about registering the component in an NGModule.
- The example component is built with Chonk UI introduced into all other wizards recently. I also pulled in the changes from https://github.com/getsentry/sentry-wizard/pull/944 to throw a custom `SentryExampleError` instead of a conventional error.
- I had to refactor `showCopyPasteInstructions` a bit to show a custom instruction message. This is responsible for all the file changes outside of the angular wizard

Screenshots:

(For reference, I threw this component into the default angular starter template, just as instructed by the wizard. So the placement is "weird" on purpose to resemble what users would experience it when they "just add it". 

Default state:

image

Success state:

image

No connection to Sentry:

image

#skip-changelog

### DIFF
--- a/src/angular/angular-wizard.ts
+++ b/src/angular/angular-wizard.ts
@@ -12,6 +12,7 @@ import {
   featureSelectionPrompt,
   getOrAskForProjectData,
   getPackageDotJson,
+  getPackageManager,
   installPackage,
   printWelcome,
   runPrettierIfInstalled,
@@ -25,6 +26,7 @@ import { updateAppConfig } from './sdk-setup';
 import { runSourcemapsWizard } from '../sourcemaps/sourcemaps-wizard';
 import { addSourcemapEntryToAngularJSON } from './codemods/sourcemaps';
 import { createExampleComponent } from './example-component';
+import { NPM } from '../utils/package-manager';
 
 const MIN_SUPPORTED_ANGULAR_VERSION = '14.0.0';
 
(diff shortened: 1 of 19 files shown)


src/angular/angular-wizard.ts
```diff
@@ -120,7 +122,7 @@ ${chalk.underline(
   Sentry.setTag('sdk-already-installed', sdkAlreadyInstalled);
 
   await installPackage({
-    packageName: '@sentry/angular@^8',
+    packageName: '@sentry/angular',
```
Lms24, line 125: this was a leftover from the v8->v9 transition period

src/utils/clack/index.ts
```diff
@@ -1457,21 +1467,23 @@ export async function askForToolConfigPath(
  *       this might require adding a custom message parameter to the function
  */
 export async function showCopyPasteInstructions(
-  filename: string,
-  codeSnippet: string,
-  hint?: string,
+  opts: ShowCopyPasteInstructionsOptions,
```
Lms24, line 1470: I changed the typing here a bit so that this function als accepts an entirely custom instruction message other than composing a predefined one from the filename and hint params. This was necessary for Angular because the wording wouldn't have fitted the predefined message.

shellmayr (approved):
Looks good to me - two small nits: 
1. There is some layout shift when Sentry's servers are not reachable - could be fixed by specifying width of the example component somehow?
2. Do selectors have to be prefixed with `app-`? Maybe this is an angular thing, but as someone not familiar with the framework, it seemed weird that I can't just paste <sentry-example-component> or similar

Lms24:
Updated the host div's layout to use fixed width and height to avoid layout shift

PR #954, opened 2025-04-16
chore(ci): add test jobs for macOS runner

philprime:
Adds the `macOS-latest` to the `runs-on` to run unit and E2E tests also on macOS.
This way we can also run tests that are expecting macOS specific environments, i.e. CLI utilities only available for macOS.

Full context in [this PR review comment thread](https://github.com/getsentry/sentry-wizard/pull/904#discussion_r2041962378).

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,12 +78,13 @@ jobs:
         run: yarn lint
 
   job_unit_test:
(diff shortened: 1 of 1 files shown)


philprime:
This might require us to update branch protections to also include the macOS ones

lforst (approved):

PR #969, opened 2025-04-18
feat(react-native): Add Feedback Widget step

antonis:
Add a step asking user if they want to add the feedback widget.

Part of https://github.com/getsentry/sentry-react-native/issues/4302

Screenshot 2025-04-22 at 1 14 37 PM


src/react-native/javascript.ts
```diff
@@ -63,11 +69,17 @@ export async function addSentryInit({
       'By default, all text content, images, and webviews will be masked for privacy. You can customize this in your code later.',
     );
   }
+  if (enableFeedbackWidget) {
+    clack.log.info(
+      `The Feedback Widget will be enabled with default settings.`,
```
antonis, line 74: We should probably move this message just after enabling the step and do the same for session replay. 
As shown in the screenshot the info texts appear mixed when both flags are turned on and this might be confusing.

Screenshot 2025-04-18 at 4 38 14 PM

src/react-native/javascript.ts
```diff
@@ -127,15 +156,32 @@ ${
...
+  enableSessionReplay = false,
+  enableFeedbackWidget = false,
+) {
+  if (!enableSessionReplay && !enableFeedbackWidget) {
+    return '';
+  }
+  return `  integrations: [${
+    enableSessionReplay ? 'Sentry.mobileReplayIntegration()' : ''
+  }${enableSessionReplay && enableFeedbackWidget ? ', ' : ''}${
+    enableFeedbackWidget ? 'Sentry.feedbackIntegration()' : ''
```
antonis, line 180: Just adding the integration for now. We may add more options with with feature drop 2 (e.g. enable capture screenshot)

src/react-native/javascript.ts
antonis: Updated with 12fac4f090bca07cac6f9db01b39af988315d37b

Screenshot 2025-04-22 at 1 14 37 PM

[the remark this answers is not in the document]
antonis: Added with a7aace287db47e7106373fd416bc9f43d1152d36

krystofwoldrich (approved):
LGTM 🚀 Ty!

PR #971, opened 2025-04-22
doc(contributing): Update package linking instructions

Lms24:
The procedure basically is:

1. register the wizard as a linked project (only needs to be done once)
2. install latest official version of the wizard (this creates the correct binary in `node_modules`)
3. link the local version (binary forwards call to local version)
4. ???
5. profit


shellmayr (approved):
These instructions worked for me - awesome! 🙏

PR #973, opened 2025-04-22
fix(sveltekit): sentry example page should be at /sentry-example-page for consistency

shellmayr:
- This aligns with the location of the page in other frameworks (consistency)
- Our onboarding tells you that the page is located at `sentry-example-page` as well:
Screenshot 2025-04-22 at 14 35 53


Lms24 (approved):
Good catch! I was loosely aware of this inconsistency with other wizards but never bothered to change it. Didn't know that it also was inconsistent with docs, so changing it makes a lot of sense. Thanks!

PR #982, opened 2025-05-05
fix(apple): Use project-relative path to gitignore file

philprime:
- Removes unused code to create `.sentryclirc`
- Fixes the path to the `.gitignore` to ignore `sentryclirc`

### DIFF
--- a/lib/Helper/SentryCli.ts
+++ b/lib/Helper/SentryCli.ts
@@ -1,15 +1,8 @@
-import * as fs from 'node:fs';
 import type { Answers } from 'inquirer';
 import * as path from 'node:path';
 
 import type { Args } from '../Constants';
-import { addToGitignore } from './Git';
-import { green, l, nl, red } from './Logging';
 import { Config } from '../Types';
-
-const SENTRYCLIRC_FILENAME = '.sentryclirc';
-const GITIGNORE_FILENAME = '.gitignore';
-const PROPERTIES_FILENAME = 'sentry.properties';
 
 export interface SentryCliProps {
   'defaults/url': string;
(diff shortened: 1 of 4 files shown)


lib/Helper/SentryCli.ts
```diff
@@ -94,70 +87,4 @@ export class SentryCli {
...
-   * Creates `.sentryclirc` and `sentry.properties` files with the CLI properties
-   * obtained from the user answers (or from logging into Sentry).
-   * The `.sentryclirc` only contains the auth token and will be added to the
-   * user's `.gitignore` file. The properties file contains the rest of the
-   * properties (org, project, etc.).
-   *
-   * @param sentryCli instance of the Sentry CLI
-   * @param cliProps the properties to write to the files
-   */
-  public async createSentryCliConfig(cliProps: SentryCliProps): Promise<void> {
```
philprime, line 108: This method is not used anywhere

test/apple/configure-sentry-cli.test.ts
```diff
@@ -0,0 +1,131 @@
+// @ts-expect-error - clack is ESM and TS complains about that. It works though
+import * as clack from '@clack/prompts';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import path from 'node:path';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { configureSentryCLI } from '../../src/apple/configure-sentry-cli';
+
+vi.mock('@clack/prompts', async () => ({
```
philprime, line 9: @Lms24 I am struggling to execute tests locally due to this error:

```
 FAIL  test/apple/configure-sentry-cli.test.ts > configureSentryCLI > .gitignore file exists > does not contain '.sentryclirc' > should append the .sentryclirc file to the .gitignore file
Error: [vitest] No "default" export is defined on the "@clack/prompts" mock. Did you forget to return it from "vi.mock"?
If you need to partially mock a module, you can use "importOriginal" helper inside:

vi.mock(import("@clack/prompts"), async (importOriginal) => {
  const actual = await importOriginal()
  return {
    ...actual,
    // your mocked methods
  }
})

 ❯ Module.createSentryCLIRC src/utils/sentrycli-utils.ts:27:7
     25|     const gitIgnore = fs.readFileSync(gitignorePath).toString();
     26|     if (!gitIgnore.includes('.sentryclirc')) {
     27|       clack.log.info(
       |       ^
     28|         `Appending .sentryclirc to .gitignore file at path: ${chalk.cyan(
     29|           gitignorePath,
 ❯ Module.configureSentryCLI src/apple/configure-sentry-cli.ts:17:15
 ❯ test/apple/configure-sentry-cli.test.ts:118:9
 ```

Any guidance how to fix this?

lib/Helper/SentryCli.ts
andreiborza: @philprime I'm not seeing myself in git blame for this, where are you seeing that? I did recently make some changes all over the codebase to enable eslint rules and fix resulting issues, might be that.

I'm not aware of it not being used by accident.

philprime: You are correct, my bad, I looked at the wrong lines because you updated `dumpConfig` which is called from `createSentryCliConfig`, which was actually written by @Lms24

Lms24: removing it sounds fine to me! IIRC we used stuff from this file in some of the new wizards for a while until we rewrote it at some point. Likely, we just missed that it's no longer used. Thanks for taking care of this!

test/apple/configure-sentry-cli.test.ts
philprime: Thank you!

andreiborza (approved):

PR #983, opened 2025-05-05
!remove(react-native): Uninstall command

krystofwoldrich:
The uninstall command is not up to date and it's technically not possible to automatically remove all Sentry code from the project code base safely.

For similar reasons none of other wizard have uninstall functionality.

This is a breaking change. Should go out with RN SDK V7.


philprime (commented):
I agree with removing this in the next major. Not sure what the procedure here is, therefore only commenting and not approving.

antonis (commented):
The changes on this PR LGTM and removing the uninstall functionality makes sense to me 👍 
Happy to approved when we are ready to merge this 🙏

Lms24 (commented):
Thanks for making this change! I Fully agree with it. Let's cut a wizard major for this just to be on the safe side. Luckily, this is very straight forward:

- please add a sentence in the changelog mentioning the breaking changes in a dedicated section
- let's add a `!` to the PR and commit message to point this out clearly (so `remove(react-native)!`)

We want to cut an SDK release today since there already a few pending changes. Let's merge this after that release so that the major itself is as small as possible.

antonis (approved):
LGTM 🚀

PR #984, opened 2025-05-05
!remove(react-native): Old SDK versions support

krystofwoldrich:
:basecamp: Based on https://github.com/getsentry/sentry-wizard/pull/983

The React Native wizard supports multiple Sentry RN SDKs versions (v5 v6) and we are currently working on a new major version v7 and related to that this PR cleans up the old versions support.

The wizard is ment to setup new projects, new project should use the latest version. We don't intentionally break compatibility with older SDK version, but as you can see in this PR, the support of older versions required a lot of extra logic. 

New supported version is RN SDK 6.12.0+

This is a breaking change.


krystofwoldrich:
Angular Job fail is unrelated to this PR, likely related to the momental Sentry unavailability.

src/react-native/react-native-wizard.ts
```diff
@@ -61,43 +55,14 @@ import {
 import xcode from 'xcode';
 
 export const RN_SDK_PACKAGE = '@sentry/react-native';
-export const RN_SDK_SUPPORTED_RANGE = '>=5.0.0';
+export const RN_SDK_SUPPORTED_RANGE = '>=6.12.0';
```
antonis, line 58: l: I think that, [If `package.json` already includes RN SDK](https://github.com/getsentry/sentry-wizard/blob/7a0443df6d87b7f2400c48bbf91481adf5de3c43/src/react-native/react-native-wizard.ts#L175) we should limit the range to `>=6.9.0` to support session replay and the feedback widget
```suggestion
export const RN_SDK_SUPPORTED_RANGE = '>=6.9.0';
```

antonis: Resolving this comment since (obviously) the existing version check covers this. I mistakenly read `6.12.0` as ` 6.1.2` 😓

antonis (approved):
LGTM 🚀 

> The wizard is ment to setup new projects, new project should use the latest version.

This makes a lot of sense ➕

Lms24 (approved):

PR #997, opened 2025-05-19
feat: Add option to NextJS for creating sentry rules LLM file

codyde:
Adding a step to the NextJS wizard which either uses a .rules folder or gives the option to create one, and places a sentryrules file in there that can be used with LLMs 

This rules file includes basic up to date examples for tracing, logging, and error tracking / exceptions. 

Based off of a discussion with @HazAT.


src/nextjs/nextjs-wizard.ts
```diff
@@ -344,24 +344,57 @@ export async function runNextjsWizardWithTelemetry(
...
+        );
+
+        clack.log.success(
+          `Created ${chalk.cyan('sentryrules.md')} in ${chalk.cyan('.rules')} directory.`,
+        );
+      } catch (error) {
+        clack.log.error(
+          `Failed to create ${chalk.cyan(
+            'sentryrules.md',
+          )} in ${chalk.cyan('.rules')} directory. Please create it manually.`,
```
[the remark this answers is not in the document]
codyde: Oh I love that idea. Sort of a "Couldn't do it, you can create it manually though!". Great suggestion.

Lms24 (commented):
I didn't fully look at this PR yet (on my list for tomorrow) but to get CI to pass, please update the NextJS E2E tests in the repo. Looks like they time out because we added a user prompt to the wizard which we don't answer in the CLI interaction in the e2e tests. 

Let me know if you need help with updating the tests.

Also it looks like the lint job is failing. Running `yarn fix` or `yarn fix:prettier` should solve that.

src/nextjs/templates.ts
```diff
@@ -711,3 +711,139 @@ export default function RootLayout({
...
 }
 `;
+
+export function getAiRulesFileContent(): string {
+  return `These examples should be used as guidance when configuring Sentry functionality within a project.
+
+# Exception Catching
+
+- Use \`Sentry.captureException(error)\` to capture an exception and log the error in Sentry.
+- Use this in try catch blocks or areas where exceptions are expected
```
Lms24, line 721: q: Should we also tell llms what NOT to do? (genuinely asking, I don't know much about rule files)

```suggestion
- Use this in try catch blocks or areas where exceptions are expected
- Do not create try/catch blocks just to send errors to Sentry. These errors will be caught and sent anyway.
```

src/nextjs/templates.ts
```diff
@@ -711,3 +711,139 @@ export default function RootLayout({
...
+      type="button" 
+      onClick={handleTestButtonClick}
+    >
+      Test Sentry
+    </button>
+  );
+}
+\`\`\`
+
+## Custom span instrumentation in API calls
```
Lms24, line 766: I'm kinda torn on this example. On the one hand, for today's purpose this is necessary if the fetch call happens after an automatically started `pageload` or `navigation` span finishes. On the other hand, we actually want to lift the restriction of only automatically starting `http.client` spans if a parent span exists. For this though, the Sentry product and ingestion pipeline needs to support span streaming which is WIP.

Either way what will happen in this example is that the manually started `http.client` span will be created and then, because now we have a parent span, our `fetch` instrumentation will start an `http.client` span on its own as well (assuming users installed `browserTracingIntegration`). 

So technically, the `op` doesn't really matter as much and neither does the name. In fact, since we're also converting the response to json (and potentially users do a lot more than this), giving this span the `http.client` op might even be wrong.

src/nextjs/templates.ts
```diff
@@ -711,3 +711,139 @@ export default function RootLayout({
...
+- The \`name\` and \`op\` properties should be meaninful for the activities in the call.
+- Attach attributes based on relevant information and metrics from the request
+
+\`\`\`javascript
+async function fetchUserData(userId) {
+  return Sentry.startSpan(
+    {
+      op: "http.client",
+      name: \`GET /api/users/\${userId}\`,
+    },
```
Lms24, line 777: IMHO manually created spans don't need to contain a well curated op. As per my comment above, I think `http.client` wouldn't be a good choice because a `http.client` span should only describe the duration of the fetch call. Also, since we can actually parameterize the route fetch request here (i.e. show the :userId parameter instead of the actual id (e.g. 1234)) we should leverage that. 

```suggestion
    {
      name: \'fetch /api/users/:userId',
    },
```

src/nextjs/templates.ts
```diff
@@ -711,3 +711,139 @@ export default function RootLayout({
...
+
+# Exception Catching
+
+- Use \`Sentry.captureException(error)\` to capture an exception and log the error in Sentry.
+- Use this in try catch blocks or areas where exceptions are expected
+
+# Tracing Examples
+
+- Spans should be created for meaningful actions within an applications like button clicks, API calls, and function calls
+- Ensure you are creating custom spans with meaningful names and operations
```
Lms24, line 726: another idea for a rule: 

```suggestion
- Ensure you are creating custom spans with meaningful names and operations
- Span names should be parameterized. For example, when get a user by an id, name the span \`fetch /users/:id\` instead of \`fetch /users/1234\`
```

Lms24 (commented):
This looks great to me! Left a comment about the e2e test, some questions and ideas for rules and a minor style correction. Thanks for working on this!

src/nextjs/nextjs-wizard.ts
```diff
@@ -1071,6 +1117,40 @@ async function askShouldSetTunnelRoute() {
...
+ * This is useful for giving the LLM context on common actions in Sentry like custom spans,
+ * logging, and error / exception handling.
+ */
+
+async function askShouldCreateAiRulesFile(): Promise<boolean> {
+  return await traceStep('ask-create-ai-rules-file', async (span) => {
+    const shouldCreateAiRulesFile = await abortIfCancelled(
+      clack.select({
+        message:
+          'Do you want to create a ./rules/sentryrules.md file with AI rule examples for Sentry?',
```
Lms24, line 1131: here we should add `chalk.cyan` around the filename

```suggestion
          'Do you want to create a ./rules/sentryrules.md file with AI rule examples for Sentry?',
```

src/nextjs/templates.ts
codyde: This is good to know! Thank you for the detail. My goal on this example was structural (i.e. educating the LLM on what a good structure looks like for guiding instrumentation). Do you have any thoughts on a better way to accomplish that? 

I def don't want to tell the LLM to use a span op thats straight up wrong.

src/nextjs/templates.ts
codyde: I made a slight update to be more assertive towards the LLM. I like your suggestion - let me know if you think my phrasing makes sense.

src/nextjs/templates.ts
codyde: I like this! Thanks for the add!

src/nextjs/templates.ts
Lms24: makes sense to me!

src/nextjs/nextjs-wizard.ts
codyde: Thanks for getting this in for me! I was going to toss it in this AM and discovered it done! 🎉

Lms24:
Hey @codyde are you still working on this? Just checking in since I'm going over older PRs to see if we can close some of them. No pressure, happy to keep it open.

Lms24:
closing as the wizard will be superseded by `sentry init`

PR #999, opened 2025-05-21
feat(sourcemaps): Add Cloudflare Wrangler Tool Flow

### LINKED ISSUE #824: Improve @sentry/wizard -i sourcemaps for Cloudflare/Wrangler
### Problem Statement

[The docs](https://docs.sentry.io/platforms/javascript/guides/cloudflare/#add-readable-stack-traces-to-errors) point you to `@sentry/wizard -i sourcemaps` when setting up Cloudflare + sourcemaps.

If you're using wrangler's default bundler, that doesn't help you much:
- It asks you for an output directory, when wrangler will build to `.wrangler/tmp/bundle-{hash}` by default.
- It also asks for a "production build command", which most boilerplates don't have (the simply have `"deploy": "wrangler deploy"`, which takes an `--env` arg)

Related: https://github.com/getsentry/sentry-javascript/issues/14841

### Solution Brainstorm

If you're using the default bundler, you really need to be using `--outdir ./dist` so sentry CLI knows where to look.

Also, rather than tacking on `npm run sentry:sourcemaps`, I think using NPM's [Pre & Post Scripts](https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts) would be a better choice, eg:

```
{
  "scripts": {
    "deploy": "wrangler deploy --outdir ./dist",
    "postdeploy": "npm run sentry:sourcemaps",
    "sentry:sourcemaps": "sentry-cli sourcemaps inject ./dist && sentry-cli sourcemaps upload ./dist"
  }
}
```

This would allow people to still run `npm deploy -- --env=production` and have sourcemaps uploaded.

Even better than that, would be to get Cloudflare to expand their [custom builds](https://developers.cloudflare.com/workers/wrangler/custom-builds/) config, so you could do something _after_ a build and not just before.

Lms24:
This PR adds a Cloudflare Wrangler tool flow to the source maps wizard. Because wrangler doesn't allow us to inject debug ids after build and before deployment, we have to resort to our release-based legacy upload strategy. 

Therefore, this setup flow needs to do the following things:
- install Sentry CLI
- modify the `wrangler deploy` npm script to
  - determine the release value from `sentry-cli releases propose-version`
  - access source maps by setting the `--outdir` arg
  - inject the `SENTRY_RELEASE` environment variable
  - add the `--upload-source-maps` arg so that we can guarantee that source maps are emitted into the outdir
- add a `sentry:sourcemaps` npm script that  
  - determines the release value from `sentry-cli releases propose-version`
  - creates a new release
  - uploads source maps for the new release
- add a `postdeploy` script that invokes the `sentry:sourcemaps` script
- create a `.sentryclirc` file and `.gitignore` it

Additionally, because of the release based stuff, I decided to show a note at the beginning that this setup is a bit limited and more complicated and we'd generally recommend using the CF Vite plugin + the `vite` sourcemaps wizard flow.

Example:

https://github.com/user-attachments/assets/aab09e40-fc44-43ec-85a4-3e919fad150a

(note in this video, the wizard is only permanently installed because that's the easiest way to run it locally)

I'll add an e2e test but probably in a different PR because I wanna try something out to simplify the e2e tests

closes #824

### DIFF
--- a/src/sourcemaps/tools/wrangler.ts
+++ b/src/sourcemaps/tools/wrangler.ts
@@ -120,21 +120,12 @@ export function getSentryCliCommand(
   options: SourceMapUploadToolConfigurationOptions & { outDir: string },
 ) {
   const sentryCliOptions = options.selfHosted ? ` --url ${options.url}` : '';
-  
+  const orgAndProjectArgs = `--org=${options.orgSlug} --project=${options.projectSlug}`;
+
   return [
     '_SENTRY_RELEASE=$(sentry-cli releases propose-version)',
-    `sentry-cli${
-      sentryCliOptions
-    } releases new $_SENTRY_RELEASE --org=${options.orgSlug} --project=${
-      options.projectSlug
-    }`,
-    `sentry-cli${
-      sentryCliOptions
-    } sourcemaps upload --org=${options.orgSlug} --project=${
-      options.projectSlug
-    } --release=$_SENTRY_RELEASE --strip-prefix '${options.outDir}/..' ${
-      options.outDir
-    }`,
+    `sentry-cli${sentryCliOptions} releases new $_SENTRY_RELEASE ${orgAndProjectArgs}`,
+    `sentry-cli${sentryCliOptions} sourcemaps upload ${orgAndProjectArgs} --release=$_SENTRY_RELEASE --strip-prefix '${options.outDir}/..' ${options.outDir}`,
   ].join(' && ');
 }
 
--- a/src/sourcemaps/tools/wrangler.ts
+++ b/src/sourcemaps/tools/wrangler.ts
@@ -119,15 +119,17 @@ async function createAndAddSentrySourcemapsScript(
 export function getSentryCliCommand(
   options: SourceMapUploadToolConfigurationOptions & { outDir: string },
 ) {
+  const sentryCliOptions = options.selfHosted ? ` --url ${options.url}` : '';
+  
   return [
     '_SENTRY_RELEASE=$(sentry-cli releases propose-version)',
     `sentry-cli${
-      options.selfHosted ? ` --url ${options.url}` : ''
+      sentryCliOptions
     } releases new $_SENTRY_RELEASE --org=${options.orgSlug} --project=${
       options.projectSlug
     }`,
     `sentry-cli${
-      options.selfHosted ? ` --url ${options.url}` : ''
+      sentryCliOptions
     } sourcemaps upload --org=${options.orgSlug} --project=${
       options.projectSlug
     } --release=$_SENTRY_RELEASE --strip-prefix '${options.outDir}/..' ${
--- a/src/sourcemaps/tools/wrangler.ts
+++ b/src/sourcemaps/tools/wrangler.ts
@@ -1,0 +1,359 @@
+// @ts-expect-error - clack is ESM and TS complains about that. It works though
+import * as clack from '@clack/prompts';
+import chalk from 'chalk';
+import {
+  abortIfCancelled,
+  addSentryCliConfig,
+  getPackageDotJson,
+  getPackageManager,
+  installPackage,
+  showCopyPasteInstructions,
+} from '../../utils/clack';
+import { hasPackageInstalled } from '../../utils/package-json';
+import { NPM } from '../../utils/package-manager';
+import type { SourceMapUploadToolConfigurationOptions } from './types';
+import path from 'path';
+import fs from 'fs';
+const SENTRY_NPM_SCRIPT_NAME = 'sentry:sourcemaps';
+
+/**
+ * only exported for testing
+ */
+export const DIST_DIR = path.join('.', 'dist');
+
+export async function configureWrangler(
+  options: SourceMapUploadToolConfigurationOptions,
+) {
+  await installPackage({
+    packageName: '@sentry/cli',
+    alreadyInstalled: hasPackageInstalled(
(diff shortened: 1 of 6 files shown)


src/sourcemaps/tools/wrangler.ts
```diff
@@ -0,0 +1,396 @@
...
+      options.projectSlug
+    }`,
+    `sentry-cli${
+      options.selfHosted ? ` --url ${options.url}` : ''
+    } sourcemaps upload --org=${options.orgSlug} --project=${
+      options.projectSlug
+    } --release=$_SENTRY_RELEASE --strip-prefix '${options.outDir}/..' ${
+      options.outDir
+    }`,
+  ].join(' && ');
```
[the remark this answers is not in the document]
Lms24: thx for the suggestion. Also extracted out org and project args as they're the same across multiple individual commands

src/sourcemaps/tools/wrangler.ts
```diff
@@ -0,0 +1,396 @@
...
+  // multiple --var arguments are allowed, so no need to check for existing --var arguments.
+  newArgs.push(
+    '--var',
+    'SENTRY_RELEASE:$(sentry-cli releases propose-version)',
+  );
+
+  // Find the index of 'deploy' command:
+  // - ensure that it's preceded by 'wrangler'
+  // - global args could be between 'wrangler' and 'deploy'
+  const deployIndex = exisitingArgs.findIndex((arg, i) => {
```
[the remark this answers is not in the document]
Lms24: I had some tests for quoted arguments but I agree, this was very brittle. I switched to `yargs` for parsing the options arguments which is convenient, since we're using `yargs` already.

src/sourcemaps/tools/wrangler.ts
```diff
@@ -0,0 +1,396 @@
...
+ * wrangler deploy command is valid.
+ *
+ * only exported for testing
+ */
+export function safeInsertArgsToWranglerDeployCommand(
+  deployCommand: string,
+  outDir: string,
+): string | undefined {
+  const exisitingArgs = deployCommand
+    .split(' ')
```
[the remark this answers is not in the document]
Lms24: It worked for `&&` but I missed some chain operators like `||`. I rewrote this logic to:

- not split by chain operators to divide the deployCommand string into sub commands
- identifying the right subcommand 
- parsing the subcommand via `yargs` to identify options
- replacing the subcommand with the args appended to the subcommand instead of the brittle inserting logic I had before.

Also added a few more tests with more chain operators and other stuff edge cases. I think this should be in a better state now.

src/sourcemaps/tools/wrangler.ts
```diff
@@ -0,0 +1,409 @@
...
+  }
+
+  return DIST_DIR;
+}
+
+/**
+ * Exported for testing
+ */
+export function getWranglerDeployCommand(deployCommand: string) {
+  const individualCommands = deployCommand.split(/&&|\|\||>>|>|<|>>|\||;/);
```
AbhiPrasad, line 398:
```suggestion
  const individualCommands = deployCommand.split(/&&|\|\||>>|>|<|\||;/);
```

src/sourcemaps/tools/wrangler.ts
```diff
@@ -0,0 +1,409 @@
...
+export function getSentryCliCommand(
+  options: SourceMapUploadToolConfigurationOptions & { outDir: string },
+) {
+  const sentryCliOptions = options.selfHosted ? ` --url ${options.url}` : '';
+  const orgAndProjectArgs = `--org=${options.orgSlug} --project=${options.projectSlug}`;
+
+  return [
+    '_SENTRY_RELEASE=$(sentry-cli releases propose-version)',
+    `sentry-cli${sentryCliOptions} releases new $_SENTRY_RELEASE ${orgAndProjectArgs}`,
+    `sentry-cli${sentryCliOptions} sourcemaps upload ${orgAndProjectArgs} --release=$_SENTRY_RELEASE --strip-prefix '${options.outDir}/..' ${options.outDir}`,
```
AbhiPrasad, line 128:
```suggestion
    `sentry-cli${sentryCliOptions} sourcemaps upload ${orgAndProjectArgs} --release=$_SENTRY_RELEASE --strip-prefix '${options.outDir}${path.delimiter}..' ${options.outDir}`,
```

src/sourcemaps/tools/wrangler.ts
```diff
@@ -0,0 +1,409 @@
...
+  outDir: string,
+): string | undefined {
+  // split deployCommand into individual bash commands (potentially separated by &&, ||, >> etc.)
+  const originalWranglerDeployCommand = getWranglerDeployCommand(deployCommand);
+
+  if (!originalWranglerDeployCommand) {
+    return undefined;
+  }
+
+  const exisitingArgs = originalWranglerDeployCommand
```
AbhiPrasad, line 309:
```suggestion
  const existingArgs = originalWranglerDeployCommand
```

AbhiPrasad (approved):

src/sourcemaps/tools/wrangler.ts
Lms24: Good catch, thx! (though I think you meant `path.sep`, right?)

Lms24: ended up extracting this out and using `path.join` which takes care of the right separator and is cleaner.

PR #1003, opened 2025-05-28
feat(nextjs): Add template for generateMetadata in App router for next@14

chargome:
ref https://github.com/getsentry/sentry-javascript/issues/16388


src/nextjs/nextjs-wizard.ts
```diff
@@ -300,6 +296,44 @@ export async function runNextjsWizardWithTelemetry(
     }
   });
 
+  await traceStep('add-generate-metadata-function', async () => {
+    const isNext14 = nextVersion?.startsWith('14.');
```
andreiborza, line 300: m: This would not match if I had `"next": "^14"` in my package.json right?

It's better to use `getNextJsVersionBucket` for this and check on `14.x`.

src/nextjs/nextjs-wizard.ts
```diff
@@ -300,6 +296,44 @@ export async function runNextjsWizardWithTelemetry(
...
+
+      await fs.promises.writeFile(
+        path.join(appDirPath, newRootLayoutFilename),
+        getRootLayoutWithGenerateMetadata(typeScriptDetected),
+        { encoding: 'utf8', flag: 'w' },
+      );
+
+      clack.log.success(
+        `Created ${chalk.cyan(
+          path.join(...appDirLocation, newRootLayoutFilename),
```
andreiborza, line 324: q: Just to double check, this does not leak the entire path right? From what I can tell, `appDirLocation` is just an array of relative sub paths right, should be okay then.

src/nextjs/nextjs-wizard.ts
```diff
@@ -300,6 +296,44 @@ export async function runNextjsWizardWithTelemetry(
     }
   });
 
+  await traceStep('add-generate-metadata-function', async () => {
+    const isNext14 = getNextJsVersionBucket(nextVersion).startsWith('14.x');
```
andreiborza, line 300:
```suggestion
    const isNext14 = getNextJsVersionBucket(nextVersion) === '14.x';
```
More direct.

src/nextjs/nextjs-wizard.ts
Lms24: yup, or use `semver` directly if the helper doesn't suffice :)

test/nextjs/utils.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    it('returns "none" for undefined version', () => {
+      expect(getNextJsVersionBucket(undefined)).toBe('none');
+    });
+
+    it('returns "<11.0.0" for versions below 11', () => {
+      expect(getNextJsVersionBucket('10.0.0')).toBe('<11.0.0');
+    });
+
+    it('returns major version for versions 11 and above', () => {
+      expect(getNextJsVersionBucket('11.0.0')).toBe('11.x');
```
andreiborza, line 26: l: Can we add some more here, e.g. `11.2.5` and `13.5.2`?

andreiborza (approved):

test/nextjs/utils.test.ts
chargome: sure

PR #1010, opened 2025-06-17
feat(nextjs): deactivate error button when adblocker is detected in example page

shellmayr:
- Current design is a bit confusing when there is the warning about not being able to connect to sentry, but when you press the button, you get a success message.
- Grey out the button if there is a connection issue and display a clearer warning about ad blockers.

Before:
Screenshot 2025-06-18 at 15 24 23

After:
Screenshot 2025-07-02 at 10 20 37

Closes TET-775

### DIFF
--- a/src/nextjs/templates.ts
+++ b/src/nextjs/templates.ts
@@ -273,46 +273,47 @@ export default function Page() {
 
         <p className="description">
           Click the button below, and view the sample error on the Sentry <a target="_blank" href="${issuesPageLink}">Issues Page</a>.
-          For more details about setting up Sentry, <a target="_blank" href="https://docs.sentry.io/platforms/javascript/guides/nextjs/">read our docs</a>.
+          For more details about setting up Sentry, <a target="_blank"
+           href="https://docs.sentry.io/platforms/javascript/guides/nextjs/">read our docs</a>.
         </p>
 
-        <button
-          type="button"
-          onClick={async () => {
-            await Sentry.startSpan({
-              name: 'Example Frontend Span',
-              op: 'test'
-            }, async () => {
-              const res = await fetch("/api/sentry-example-api");
-              if (!res.ok) {
-                setHasSentError(true);
-                throw new SentryExampleFrontendError("This error is raised on the frontend of the example page.");
-              }
-            });
-          }}
-        >
-          <span>
-            Throw Sample Error
-          </span>
-        </button>
+        <div className="button-container">
+          <button
+            type="button"
+            onClick={async () => {
+              await Sentry.startSpan({
+                name: 'Example Frontend/Backend Span',
+                op: 'test'
+              }, async () => {
+                const res = await fetch("/api/sentry-example-api");
+                if (!res.ok) {
+                  setHasSentError(true);
+                }
+              });
+              throw new SentryExampleFrontendError("This error is raised on the frontend of the example page.");
+            }}
+            disabled={!isConnected}
+          >
+            <span>
+              Throw Sample Error
+            </span>
+          </button>
+        </div>
 
         {hasSentError ? (
           <p className="success">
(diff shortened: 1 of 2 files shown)


src/nextjs/templates.ts
```diff
@@ -425,6 +451,11 @@ export default function Page() {
           color: #FFFFFF;
           text-decoration: underline;
         }
+
+        .button-container {
```
andreiborza: l: I think we can remove the extra div and just change the margin on the button. It already has a margin-top anyway?

src/nextjs/templates.ts
```diff
@@ -272,46 +272,47 @@ export default function Page() {
...
+              });
+              throw new SentryExampleFrontendError("This error is raised on the frontend of the example page.");
+            }}
+            disabled={!isConnected}
+          >
+            <span>
+              Throw Sample Error
+            </span>
+          </button>
+        </div>
```
priscilawebdev:
```suggestion
          <button
            type="button"
            onClick={async () => {
              await Sentry.startSpan({
                name: 'Example Frontend/Backend Span',
                op: 'test'
              }, async () => {
                const res = await fetch("/api/sentry-example-api");
                if (!res.ok) {
                  setHasSentError(true);
                }
              });
              throw new SentryExampleFrontendError("This error is raised on the frontend of the example page.");
            }}
            disabled={!isConnected}
          >
            <span>
              Throw Sample Error
            </span>
          </button>
```

the container has only one element, so the gap is never applied. We can remove it

src/nextjs/templates.ts
```diff
@@ -378,6 +379,21 @@ export default function Page() {
...
+            & > span {
+              background-color: #AAA;
+              border-color: #888;
+              transform: translateY(-2px);
+            }
+
+            &:hover > span {
+              transform: translateY(-2px);
+            }
+          }
```
priscilawebdev: The button looks a lot like the one we use in our app, so I’m guessing we were trying to match that style. In our version, when the button is disabled, there’s no hover animation, and we just lower the opacity to show it’s inactive. The styles could look something like this:

```suggestion
          &:disabled {
            cursor: not-allowed;
            opacity: 0.6;

            & > span {
              transform: translateY(0);
              border: none
            }
          }
```

example app:


https://github.com/user-attachments/assets/6bec9028-8785-4b91-bb8a-170ba37f2ba3



with the suggested styles:

https://github.com/user-attachments/assets/c6405b8d-2807-40c3-82a5-9e7edde76773

src/nextjs/templates.ts
```diff
@@ -406,6 +422,16 @@ export default function Page() {
           color: #181423;
         }
 
+        .error {
+          padding: 12px 16px;
+          border-radius: 8px;
+          font-size: 20px;
+          line-height: 1;
+          background-color: #FF4444;
+          border: 1px solid #CC0000;
```
priscilawebdev: what about not having a border? or making it red as well?

src/nextjs/templates.ts
shellmayr: Great point, putting this in!

src/nextjs/templates.ts
shellmayr: This is leftover code, thanks for spotting - removing it; the connectivity-error class actually looks similar, but this is an approved design (and not touched in this commit).

src/nextjs/templates.ts
shellmayr: Great catch here - removed it!

src/nextjs/templates.ts
shellmayr: Same as andrei's comment below - thanks for spotting!

priscilawebdev (approved):
didn't test it locally again, but the applied code looks good! great work 🙌

PR #1011, opened 2025-06-18
feat(nextjs): Switch to randomized tunnel route as default

chargome:
ref https://github.com/getsentry/sentry-javascript/pull/16626

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- feat(nextjs): Switch to randomized tunnel route as default ([#1011](https://github.com/getsentry/sentry-wizard/pull/1011))
+
 ## 5.1.0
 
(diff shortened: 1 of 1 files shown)


andreiborza (commented):
As discussed on slack, I'm wondering if we want to actually default the tunnel route option to `true`. Maybe setting it to `true` is okay if we mention that they should also switch to an allow list for routes that should be handled by the middleware.

chargome:
I'd first test this option within our own apps before making it the default within the wizard 👍

Lms24 (commented):
This PR doesn't actually change the default value of the wizard, right? It just changes that if users opt into tunneling, they get the randomized route. 
Just asking because we shouldn't default to tunneling due to it potentially heavily increasing traffic on Vercel and therefore cost for users. IIRC the wizard mentions this when it asks users if they want tunneling.

chargome:
The risk of breaking apps that use turbopack is too high here as we cannot automatically add matchers for random routes in turbopack.

PR #1023, opened 2025-07-07
fix(react-native): Avoid importing `isPlainObject` from `@sentry/utils`

Lms24:
This PR removes an import from `@sentry/utils` from the RN wizard in favour of a slightly simplfied version of the same `isPlainObject` check. 

The import was problematic, because `@sentry/utils` is not a direct dependency of the wizard but rather a transitive dep of `@sentry/node`. For yarn and npm, this doesn't matter but pnpm complains a lot about such implicit dependency imports. Not an immediate issue but nonetheless good to fix.

Why not just register utils as a dependecy? I wanna avoid version divergences with the `@sentry/node` package. It probably doesn't matter but it also seems cleaner to me to avoid registering yet another dependency overall.

### DIFF
--- a/src/react-native/expo.ts
+++ b/src/react-native/expo.ts
@@ -184,6 +184,10 @@ export function getSentryAppConfigJsonCodeSnippet({
   });
 }
 
+/**
+ * Checks whether the given value is a plain object (as opposed to something like a class instance or a primitive).
+ * vendored from @sentry/utils to avoid registering the dependency.
+ */
 function isPlainObject(what: unknown): boolean {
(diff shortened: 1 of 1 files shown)


src/react-native/expo.ts
```diff
@@ -184,3 +183,7 @@ export function getSentryAppConfigJsonCodeSnippet({
 }`);
   });
 }
+
+function isPlainObject(what: unknown): boolean {
```
philprime, line 191: `l`: Might be worth it to add a comment here in the code, so that we don't need to lookup this PR when reading the code.

PR #1024, opened 2025-07-14
feat: Introduce run build command and prompts

priscilawebdev:
This PR improves the build artifact detection and selection flow in the setup wizard. Key changes:

**Create React App Support:**

- ./builder is now suggested as the default build folder when using Create React App.

**Improved Build Artifact Recovery for All Packages:**

- If the user-specified build folder is not found, the wizard now offers clearer fallback options:
  - Run the build command
  - Re-enter the path
  - Proceed without selecting a folder
  
- If the user chooses to run the build command:
  - The command is executed, and the wizard re-checks for the build folder.
  - If still not found, the wizard suggests other likely folders.

- If possible folders are found:
  - The user is prompted to pick one or none
  - If the user declines all suggestions, the wizard repeats the prompt with the original fallback options: run, re-enter, or proceed.

closes https://linear.app/getsentry/issue/TET-434/onboardingreact-wizard-does-not-know-where-build-artifacts-for-create

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Introduce run build command and prompts in soucemaps wizard ([#1024](https://github.com/getsentry/sentry-wizard/pull/1024))
+- Introduce run build command and prompts in sourcemaps wizard ([#1024](https://github.com/getsentry/sentry-wizard/pull/1024))
 - Add a default build path for create-react-app in soucemaps wizard ([#1025](https://github.com/getsentry/sentry-wizard/pull/1025))
 
 ## 5.3.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## Unreleased
 
-- Introduce run build command and prompts in sourcemaps wizard ([#1024](https://github.com/getsentry/sentry-wizard/pull/1024))
-- Add a default build path for create-react-app in soucemaps wizard ([#1025](https://github.com/getsentry/sentry-wizard/pull/1025))
+- feat: Introduce run build command and prompts in sourcemaps wizard ([#1024](https://github.com/getsentry/sentry-wizard/pull/1024))
+- feat: Add a default build path for create-react-app in soucemaps wizard ([#1025](https://github.com/getsentry/sentry-wizard/pull/1025))
 
 ## 5.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Introduce run build command and prompts in soucemaps wizard ([#1024](https://github.com/getsentry/sentry-wizard/pull/1024))
+
 ## 5.3.0
 
 - fix(react-native): Avoid importing `isPlainObject` from `@sentry/utils` ([#1023](https://github.com/getsentry/sentry-wizard/pull/1023))
--- a/src/utils/clack/index.ts
+++ b/src/utils/clack/index.ts
@@ -1680,7 +1680,7 @@ export async function promptWhatToDo({
 
   if (buildCommand) {
     options.push({
-      label: 'Run the build command for me',
+      label: 'Let the wizard run the build command',
       value: 'run-build',
     });
   }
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Introduce run build command and prompts in soucemaps wizard ([#1024](https://github.com/getsentry/sentry-wizard/pull/1024))
```
priscilawebdev:
```suggestion
- Introduce run build command and prompts in sourcemaps wizard ([#1024](https://github.com/getsentry/sentry-wizard/pull/1024))
```

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- Introduce run build command and prompts in sourcemaps wizard ([#1024](https://github.com/getsentry/sentry-wizard/pull/1024))
+- Add a default build path for create-react-app in soucemaps wizard ([#1025](https://github.com/getsentry/sentry-wizard/pull/1025))
```
andreiborza:
```suggestion
- feat: Introduce run build command and prompts in sourcemaps wizard ([#1024](https://github.com/getsentry/sentry-wizard/pull/1024))
- feat: Add a default build path for create-react-app in soucemaps wizard ([#1025](https://github.com/getsentry/sentry-wizard/pull/1025))
```

src/utils/clack/index.ts
```diff
@@ -1651,3 +1654,193 @@ export async function askShouldAddPackageOverride(
...
+  return typeof packageDotJson.scripts?.build === 'string' ? 'build' : null;
+}
+
+export async function artifactsExist(relativePath: string): Promise<boolean> {
+  try {
+    await fs.promises.access(path.join(process.cwd(), relativePath));
+    return true;
+  } catch {
+    return false;
+  }
```
andreiborza: m: We can replace this with ` fs.existsSync(path.join(process.cwd(), relativePath))`

src/utils/clack/index.ts
```diff
@@ -1651,3 +1654,193 @@ export async function askShouldAddPackageOverride(
...
+}
+
+export async function promptWhatToDo({
```
andreiborza: m: Could we give this a more descriptive name, maybe `askForBuild` or something the like?

src/utils/clack/index.ts
```diff
@@ -1651,3 +1654,193 @@ export async function askShouldAddPackageOverride(
...
+    try {
+      await fs.promises.access(path.join(process.cwd(), folder));
+      existingFolders.push(folder);
+    } catch {
+      // Folder doesn't exist, continue
+    }
+  }
+
+  return existingFolders;
+}
```
andreiborza: l: We could simplify this to

```js
function getPossibleBuildFolders(): string[] {
    const commonBuildFolders = ['build', 'dist', 'out', '.next'];

    return commonBuildFolders.filter(folder =>
      fs.existsSync(path.join(process.cwd(), folder))
    );
}
```

wdyt?

src/utils/clack/index.ts
```diff
@@ -1651,3 +1654,193 @@ export async function askShouldAddPackageOverride(
...
+      cwd: process.cwd(),
+    });
+    spinner.stop('Build finished running.');
+    return true;
+  } catch (error) {
+    spinner.stop(
+      `Build failed: ${error instanceof Error ? error.message : String(error)}`,
+    );
+    clack.log.error(
+      'Failed to run build command. Please run it manually and try again.',
```
andreiborza: l: 

```suggestion
      'Failed to run your build command. Please run it manually and try again.',
```

Might make it a bit clearer that it's their build command we failed to run, not the wizard's?

src/utils/clack/index.ts
priscilawebdev: maybe `askWhatToDoNext`? because it is not only a build. It offers 3x different options - run the build command, re-enter the folder's path manually, or proceed with the initial folders path

andreiborza: Can we find something that's more expressive than `WhatToDo`, something related to build? I think `askForBuild` at least gives you the idea that it's about building, I think it also fits with the 3 options because they're all related to build or build path right?

andreiborza: Maybe `askForBuildOrBuildpath` if we want to be more verbose?

andreiborza (approved):
Thanks for adding this! Good stuff.

PR #1036, opened 2025-07-25
feat(angular): Add option to enable logs to be sent to Sentry

AbhiPrasad:
resolves https://github.com/getsentry/sentry-wizard/issues/1032


e2e-tests/tests/angular-17.test.ts
Lms24, line 1: heads-up: Angular e2e tests are skipped at the moment because they were flaky af. So this won't surface any issues on CI. Not ideal but I need to revisit #976 to enable them again.

The changes seem fine to me though, so I'm fine with merging this anyway (note for myself: I guess there are no unit tests right now for the `init` snippet generation. Would probably also be good to get better coverage there.

Lms24 (approved):

PR #1045, opened 2025-07-25
feat(nextjs): refactoring wizard to use template files

nikolovlazar:
This PR refactors the wizard (**Next.js only!**) to use template files instead of inline strings. This will allow us to generate more advanced examples for Next.js users besides the `/sentry-example-page`.

DevEx is looking to develop a built-in mini Sentry crash course besides the `/sentry-example-page` that'll teach devs how to use Sentry features, and refactoring the wizard to support template files is a prerequisite for us.

I moved all the static inline strings into template files in `scripts/NextJs/templates`, built a `src/nextjs/template-loader.ts` class that knows how to load templates and replace variables.

Tests are passing, but I did need to refactor them as well because there were discrepancies in minor stuff like new lines and rogue spaces.


Lms24 (commented):
First pass over this but I have some questions/requests:

> and refactoring the wizard to support template files is a prerequisite for us.

naive question but why? I can see the example components being better suited as templates but I don't yet see this being necessary for the other files (next config, sentry.config, instrumentation, etc). Not particularly against it but I'm a bit afraid such a large change at once breaks more than it helps. Can we split this up a bit?

What makes this a bit spicy is that we also need to ensure we correctly include the template files in the tarball before publishing. Can you check this please? Furthermore, we build and ship native binaries so that the wizard can be ran on systems where node is not installed. I'm not sure if the template files play well with that. Perhaps @BYK can answer this.
 
RE failing tests e2e: It seems like all NextJS e2e tests on CI fail. I suspect this is because in CI, we run `yarn test:e2e:bin` ([see here](https://github.com/getsentry/sentry-wizard/blob/d4d2459b546fbdaaa1f4b2df05d003401c5f3608/.github/workflows/build.yml#L196)), instead of `yarn test:e2e`. The difference is that we test against the fossilized binaries I mentioned earlier. Which further makes me think that the binaries don't include the template files 🤔 

> feat(nextjs): refactoring wizard to use template files

nit but: What makes this a `feat`? I might be missing something but afaict, this is rather a `ref`, no?

Last point: I realize this isn't yet important for this very PR but rather for the whole project: If we build more complex UI in addition to the current example page, I think we need to test this more properly. Perhaps, it's worth writing actual components where the variables within the components become props? This way, we can mount them in some Vite UI or playwright tests and have this covered properly. Also means we can type and syntax check the components instead of relying on string templates (be it inline or in file). The tests could actually live within the example test apps where we add playwright tests to them. WDYT?

I know these are quite some requests from my end. Happy to chat about them. My main concern overall is that this additional functionality means, we inject more stuff into users' apps. So we should make sure this works properly as otherwise UX would be is worse than today.

BYK:
>Furthermore, we build and ship native binaries so that the wizard can be ran on systems where node is not installed. I'm not sure if the template files play well with that. Perhaps @BYK can answer this.

Great catch. Yeah `this.templateBasePath = path.join(__dirname, '../../templates/nextjs');` would definitely not work in that and we may want to add a lint rule for these patterns.

We either need to make these files JS modules so they get bundled in, or manually add them to the Node binary and use the Node SEA API to read them at runtime. I'd prefer the former solution if possible.

Lms24:
Maybe as a middleground solution, we can keep the templates as JS modules via template strings but split them up to a similar structure that you added here in this PR? (this is me assuming one of the reasons you wanted to change this was this file being super long and annoying to look up stuff)

So instead of one `templates.ts` TS file, we have multiple (e.g. per config file type, or even as many permutations as right now) with a syntax like this?

```ts
export const nextjsConfigMjsTemplate = (withSentryConfigOptionsTemplate: string) => `
import { withSentryConfig } from "@sentry/nextjs";

/** @type {import('next').NextConfig} */
const nextConfig = {};

export default withSentryConfig(
  nextConfig,
  ${withSentryConfigOptionsTemplate}
);
`;
```

this is a bit more elegant than the current `function` syntax in terms of indentation. Plus I think the builtin template strings are nicer than `string.replace` (personal pref maybe). Importantly, it saves us all the trouble around loading and looking up files, especially considering the binaries.

In terms of testing, we could probably do something around calling this function and writing it to a temp file. But realistically, doing playwright within the e2e test apps is easier. We can still add snapshot tests for the generated code per function by the way.

nikolovlazar:
@Lms24 that makes sense. I thought I'd refactor everything (examples + config) with template files to maintain consistency in the codebase. It makes sense the issue to be the missing template files in the fossilized binaries because I didn't do anything specific to make sure the templates are bundled.

Since that's the case, I do agree now that if we use JS modules it will be simpler build-wise, and I see them (single template string literal in file) being just as easy to work with as the templates I pushed in this PR.

I'm going to revert the config changes as they were before, but I'll refactor the `template-loader.ts` to only produce the advanced example and use JS modules for the example files instead of template files.

Thanks for the review!

nikolovlazar:
I'll get that working in a different PR. Will close this for now.

Issue #1007, opened 2025-06-11
Multiple wizard runs produce N-wrapped `next.config.js`


Lms24:
### Environment

What version are you running? Etc.

5.1.0, -i nextjs

### Steps to Reproduce

1. npx @sentry/wizard@latest -i nextjs
2.npx @sentry/wizard@latest -i nextjs but this time confirm that the wizard should modify `next.config.js` again

![Image](https://github.com/user-attachments/assets/afde2761-a4e2-48d1-999e-4fb975ac08c2)

3. observe `withSentryConfig` wrapping `withSentryConfig`, wrapping `withSentryConfig, wrapp...

### Expected Result

wrapper is only applied once

### Actual Result

wrapper applied multiple times

PR #1058, opened 2025-08-06
fix(nextjs): Prevent double-wrapping of Next.js config

s1gr1d:
Prevents double-wrapping in MJS/TS files. The CJS content is not replaced, but appended, so there is not double-wrapping in the first place.

I also tested this in a local Next.js project.

Closes https://github.com/getsentry/sentry-wizard/issues/1007

### DIFF
--- a/test/nextjs/wizard-double-wrap-prevention.test.ts
+++ b/test/nextjs/wizard-double-wrap-prevention.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from 'vitest';
 // @ts-expect-error - magicast is ESM and TS complains about that. It works though
-import { builders, parseModule } from 'magicast';
+import { generateCode, parseModule } from 'magicast';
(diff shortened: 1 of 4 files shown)


test/nextjs/wizard-double-wrap-prevention.test.ts
Lms24, line 1: Thanks a lot for adding these tests! I have one request: Can we test against string input and outputs(`toMatchInlineSnapshot` is probably the best option)? This should make it much easier to grasp what we expect to happen.

s1gr1d: I added the inline snapshots :) 
The `export default` is always missing in the snapshot string, but it's because it's specifically looking at `mod.exports.default`. The tests below look at the whole file, so this the general approach of keeping the `export default` is tested as well here.

Lms24 (approved):

PR #1059, opened 2025-08-06
feat: track feature selection as span attributes

AbhiPrasad:
It would be good to get an idea of how many people select different features during installation, specifically for logs.

This PR adds that.


AbhiPrasad:
Should this have a changelog entry?

PR #1060, opened 2025-08-07
feat(nuxt): Set `sendDefaultPii: true` by default

onurtemizkan:
Resolves: #961

### DIFF
--- a/e2e-tests/test-applications/nuxt-4-test-app/package.json
+++ b/e2e-tests/test-applications/nuxt-4-test-app/package.json
@@ -14,5 +14,8 @@
     "nuxt": "3.15.0",
     "vue": "latest",
     "vue-router": "latest"
+  },
+  "resolutions": {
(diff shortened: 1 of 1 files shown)


e2e-tests/test-applications/nuxt-4-test-app/package.json
```diff
@@ -14,5 +14,8 @@
     "nuxt": "3.15.0",
     "vue": "latest",
     "vue-router": "latest"
+  },
+  "resolutions": {
```
Lms24, line 18: do we need this for the test app to run? just curious, no action required

onurtemizkan: Yes, it was prompting and breaking the E2E test flow before adding this resolution

onurtemizkan:
@Lms24, updated 👍

PR #1063, opened 2025-08-08
feat: Add Sentry MCP server during wizard setup for applicable platforms 

codyde:
Added support for MCP setup as final step of Wizard based setups. Created utility file for MCP configuration to apply across all applicable projects. 

Added approriate test coverage. 

Wizard will append MCP server configurations to relevant configuration files for MCP servers based on selection (Cursor, VS Code, Claude Code). Defaults to "yes" for adding.


src/utils/mcp-config.ts
```diff
@@ -0,0 +1,222 @@
...
+  );
+
+  if (!shouldAdd) {
+    return;
+  }
+
+  type EditorChoice = 'cursor' | 'vscode' | 'claudeCode';
+  const editor: EditorChoice = await abortIfCancelled(
+    clack.select({
+      message: 'Which editor do you want to configure?',
```
s1gr1d: IntelliJ IDEs (like IDEA, WebStorm, PhpStrom, PyCharm) are [very popular as well](https://survey.stackoverflow.co/2025/technology#1-dev-id-es) and we could add it here as well. Seems like the JSON config is the same over all editors.

- WebStorm: https://www.jetbrains.com/help/webstorm/mcp-server.html
- IntelliJ IDEA: https://www.jetbrains.com/help/idea/mcp-server.html

antonis (approved):
The React Native changes LGTM 🚀

src/utils/mcp-config.ts
```diff
@@ -0,0 +1,222 @@
...
+ */
+export async function offerProjectScopedMcpConfig(): Promise<void> {
+  const shouldAdd = await abortIfCancelled(
+    clack.select({
+      message:
+        'Optionally add a project-scoped MCP server configuration for the Sentry MCP?',
+      options: [
+        { label: 'Yes', value: true },
+        { label: 'No', value: false, hint: 'You can add it later anytime' },
+      ],
```
Lms24: m: I'd suggest adding a third option here: "What is MCP?" Or something along these lines. I don't think this is yet common knowledge that everyone running the wizard will know what we're talking about. So likely users will end up asking themselves "Do I need this for Sentry? What is this all about?". For them, that third option would be nice. I imagine this just printing a paragraph of information why it's helpful and what users get out of using the Sentry MCP server. 
The benefit of this route is that we don't have to print a long explainer in every case. Which I think is good because at that point we're basically done with setup and a long message would be kinda out of proportion UX-wise (IMHO).

src/utils/mcp-config.ts
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  type EditorChoice = 'cursor' | 'vscode' | 'claudeCode';
+  const editor: EditorChoice = await abortIfCancelled(
+    clack.select({
+      message: 'Which editor do you want to configure?',
+      options: [
+        { value: 'cursor', label: 'Cursor (project .cursor/mcp.json)' },
+        { value: 'vscode', label: 'VS Code (project .vscode/mcp.json)' },
+        { value: 'claudeCode', label: 'Claude Code (project .mcp.json)' },
```
Lms24: l: If we end up attempting editor auto-detection, we can pre-select the editor we auto-detected (and fall back to the first or whatever if we fail)

src/utils/mcp-config.ts
```diff
@@ -0,0 +1,222 @@
...
+  } catch {
+    throw new Error('Failed to update .mcp.json');
+  }
+}
+
+/**
+ * Offers to add a project-scoped MCP server configuration for the Sentry MCP.
+ * Supports Cursor, VS Code, and Claude Code.
+ */
+export async function offerProjectScopedMcpConfig(): Promise<void> {
```
Lms24: h: We need to think about users who don't use one of the supported editors and also about users who use editors that do not support MCP (like for example xcode). They might out of curiosity answer the frist prompt with "Yes" and get stuck. Some thoughts here:
- I'd suggest attempting to auto-detect the used editor and simply omit asking anything if we don't detect a supported editor. We can keep track of how often we end up asking this question by using `Sentry.setTag` like we do in a lot of places and then check the data. On that note, we can also set a tag for the editor selection to get a feeling for how often users choose which option. 
- Alternatively, we need an "I use another editor" option in the editor selection prompt. Which can become quite tricky because I don't know what to suggest these users then. 
  - I don't think Sentry should tell them "well then you better start using <editor of choice> Pal" :D 
  - So maybe a polite, "we don't support your editor right now. Please check if it supports MCP and if yes, open an issue for us to add support" angle would be better.
  - Related, if we have public documentation about our MCP we should include a link there.

src/utils/mcp-config.ts
Lms24, line 1: since we're using clack prompts in this file, can we move it to `src/utils/clack/mcp-config.ts`? Ideally, we collect all reusable prompts within the clack utils folder. Feel free to split this up with functionality not adding any kind of console input/output ("true utils" if you will 😅)

e2e-tests/tests/nextjs-15.test.ts
```diff
@@ -82,13 +82,20 @@ describe('NextJS-15', () => {
...
-        'Successfully installed the Sentry Next.js SDK!',
+        'Optionally add a project-scoped MCP server configuration for the Sentry MCP?',
+        { optional: true },
       ));
 
+    // Decline optional MCP config (default is now Yes, so press DOWN to select No)
+    await wizardInstance.sendStdinAndWaitForOutput(
+      [KEYS.DOWN, KEYS.ENTER],
+      'Successfully installed the Sentry Next.js SDK!',
+    );
```
Lms24: m: I think it makes sense to assert on the config in at least one e2e test. As far as I can tell, we just always opt out right now. Feel free to skip in all but one and happy to leave up to you which one should actually check it.

src/utils/mcp-config.ts
codyde: Im a little worried on the auto detection of editor side - mostly because of the different ways people store things on systems. Predicable locations for apps can be tough sometimes. You might have a really good suggestion on this though - so im all ears there. 

I think that the "I use another editor" approach could work, and what we could do is output the MCP Server configuration JSON that they could use. 

How does that sound?

src/utils/mcp-config.ts
codyde: So both of those are their actual MCP server implementation - not their MCP client implementation. I dug around a bunch, and most of the jetbrains IDEs dont have a "JSON config location" setting (at least not that I can find) so what im going to do is have it print out the config that the user can copy and use in the jetbrains IDEs 

That work for you?

src/utils/mcp-config.ts
codyde: Done! Great suggestion. Screenshot - 

CleanShot 2025-08-13 at 1  28 40@2x

src/utils/mcp-config.ts
codyde: Heres what it looks like - 

CleanShot 2025-08-13 at 1  29 20@2x

src/utils/mcp-config.ts
codyde: Heres what I went with for now - curious if you feel like is passable. 

CleanShot 2025-08-13 at 1  40 13@2x

src/utils/mcp-config.ts
codyde: Can you explain this one a bit more to me just so im clear? Sorry!

e2e-tests/tests/nextjs-15.test.ts
codyde: Added this to nextjs 15!

src/utils/mcp-config.ts
s1gr1d: Yeah that sound good! It's a good solution that we give people the snippet they need - thanks for adding this 🙌

src/utils/mcp-config.ts
Lms24: Yeah you're bringing up a good point! I think for now the "I use another editor" option is fine! Auto detection is always limited, so we need some fallback anyway.

src/utils/mcp-config.ts
Lms24: Let's just move this file from 

`src/utils/mcp-config.ts`
to
`src/utils/clack/mcp-config.ts` 

because we're using `@clack/prompts` in this file, I want it to be where the other reusable clack helper functions are. The rest (potentially separating clack stuff from logic) we can do later (or never lol)

PR #1068, opened 2025-08-12
chore: remove unused deps (tsx, dotenv, @sentry-internal/sdk)

outslept:
continuing #1065

PR #1069, opened 2025-08-12
chore: migrate from `chalk` to `picocolors`

outslept:
image

vs

image


outslept:
Seems picocolors posts extra ansi escape sequences. Investigating.

UPD: `chalk` disables colors in CI by default and `picocolors` doesn't. I'll add `NO_COLOR` to `vitest.config`

outslept:
Alright, the problem should be fixed. Could you please re-run the jobs? It failed due to flaky e2e tests unrelated to this change (had the same-ish behavior in #1068).

Lms24 (commented):
I re-ran e2e tests. If the same test fails (cloudflare source maps), I need to strip out ansi color/format characters in my testing library that's used by the cloudflare source maps e2e test. I'm surprised we handle this in the other e2e tests though 😅 

EDIT: looks like that's the problem. Unfortunately, we can't set `NO_COLOR=1` during build for the e2e tests because they use the same build artifacts that we end up publishing (kinda the point of an e2e test xD). So I'll fix this in `clifty` in a bit.

Update: Not so sure if this is the problem anymore. Still need to do some more debugging but don't have time right now. If you get a chance, you could try running the cloudflare wrangler source maps e2e test locally and check if you spot any obvious problems. Will revisit this as soon as I can.

outslept:
I duplicated the branch to create a PR in my own fork to check the CI logs first before providing feedback. The CI log shows the issue - clifty is expecting `Installing @sentry/cli` but receiving `[35m◒[39m Installing [1m[36m@sentry/cli[39m[22m with [1mYarn V1[22m...` with ANSI escape sequences that it cannot parse, causing the problem.


If you'd like to provide striping on clifty side, you could use Node.js built-in `util.stripVTControlCharacters(str)` (https://nodejs.org/api/util.html#utilstripvtcontrolcharactersstr) to strip ANSI sequences. It should apply to this https://github.com/Lms24/clifty/blob/0d7ae7c4aa8442cd1ce6b75edd6d41fd57e32fbc/packages/clifty/src/scenarioBuilder.ts#L159-L178

And it would look like this:

```js
child.stdout.on("data", (data) => {
  const cleanData = stripVTControlCharacters(data.toString());
  this.#stdoutBuffer += cleanData;
  this.output.dispatchEvent(
    new CustomEvent("stdout", { detail: cleanData })
  );
  if (this.#testEnv.debug) {
    console.log(this.#stdErrBuffer);
  }
});
```

```js
child.stderr.on("data", (data) => {
  const cleanData = stripVTControlCharacters(data.toString());
  this.#stdErrBuffer += cleanData;
  this.output.dispatchEvent(
    new CustomEvent("stderr", { detail: cleanData })
  );

  if (this.#testEnv.debug) {
    console.log(this.#stdErrBuffer);
  }
});
```

outslept:
This is the workflow run from my fork (content doesn't differ from this branch, I just removed `NO_COLOR` from vitest config): https://github.com/outslept/sentry-wizard/actions/runs/17007457214/job/48219026060?pr=1

Open logs and look for (cloudflare) :

```
2025-08-16T10:38:59.2196791Z Waiting for: 
2025-08-16T10:38:59.2197230Z Installing @sentry/cli
```

And all the ansi hell

```
2025-08-16T10:38:59.2327378Z [999D[J[35m◒[39m  Installing [1m[36m@sentry/cli[39m[22m with [1mYarn V1[22m[999D[J[35m◐[39m  Installing [1m[36m@sentry/cli[39m[22m with [1mYarn V1[22m[999D[...
```

The reason there are so many ANSI codes is because the CLI spinner updates every x ms with new animation frames, and each update gets accumulated in stdout buffer. While clifty waits for the timeout, the spinner continuously outputs new frames with cursor movement commands and color codes, creating this ANSI sequence spam.

UPD: Maybe this is deeper than I assume. I'm trying to localize the issue, but couldn't provide a reproduction. For some reason it seems really easy on paper, but weird in practice. Weird, but I'll keep going.

Lms24:
@outslept apologies for the late reply! I added your suggestions to `clifty` and updated to the latest version in this issue. Might be worth giving this PR a rebase to check if this fixed the issue. Either way, thanks a lot for letting me know about `stripVTControlCharacters`. TIL!

Lms24:
Hey, @outslept, I don't think we'll have time to pursue this further and I also don't want to waste your time on this. We're planning on retiring the wizard in favour of https://github.com/getsentry/cli. So I'm gonna close this PR. Still thank you for working on this. Really appreciate the modernization effort :)

PR #1076, opened 2025-08-28
feat: Add wizard for `react-router` framework mode

onurtemizkan:
Added `react-router` wizard, similar to our `remix` wizard

### DIFF
--- a/lib/Constants.ts
+++ b/lib/Constants.ts
@@ -59,7 +59,7 @@ export function getIntegrationDescription(type: string): string {
     case Integration.remix:
       return 'Remix';
     case Integration.reactRouter:
-      return 'React Router';
+      return 'React Router (framework)';
     case Integration.sveltekit:
       return 'SvelteKit';
     case Integration.sourcemaps:


e2e-tests/tests/react-router.test.ts
```diff
@@ -0,0 +1,390 @@
...
+
+  test('entry.client file contains Sentry initialization', () => {
+    checkFileContents(`${projectDir}/app/entry.client.tsx`, [
+      'import { init, replayIntegration, browserTracingIntegration } from "@sentry/react-router";',
+      `init({
+    dsn: "${TEST_ARGS.PROJECT_DSN}",
+    tracesSampleRate: 1,
+    enableLogs: true,
+
+    integrations: [browserTracingIntegration({
```
andreiborza: q: Why are we expecting the `browserTracingIntegration`? I don't see this in the docs.

src/react-router/react-router-wizard.ts
```diff
@@ -0,0 +1,259 @@
...
+      await initializeSentryOnEntryClient(
+        selectedProject.keys[0].dsn.public,
+        featureSelection.performance,
+        featureSelection.replay,
+        featureSelection.logs,
+        typeScriptDetected,
+      );
+    } catch (e) {
+      clack.log.warn(
+        `Could not initialize Sentry on client entry.\n  Please do it manually using instructions from https://docs.sentry.io/platforms/javascript/guides/react-router/`,
```
andreiborza: m: In the other wizard we print what users have to add if steps fail. We should do the same for all of these.

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+  const clientEntryFilename = `entry.client.${isTS ? 'tsx' : 'jsx'}`;
+  const clientEntryPath = path.join(process.cwd(), 'app', clientEntryFilename);
+
+  try {
+    if (!fs.existsSync(clientEntryPath)) {
+      clack.log.warn(
+        `Could not find ${chalk.cyan(
+          clientEntryFilename,
+        )}. Skipping client entry instrumentation.`,
+      );
```
andreiborza: m: I think we could ask users if they want to try rerunning reveal and if that still didn't work then print out what they have to add to the entry.client file.

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+      debug('Failed to write client entry file:', writeError);
+      clack.log.warn(
+        `Failed to automatically update ${chalk.cyan(clientEntryFilename)}.`,
+      );
+
+      await showCopyPasteInstructions({
+        filename: clientEntryFilename,
+        codeSnippet: sentryInitCode,
+        hint: 'Add this code at the top of your client entry file',
+      });
```
andreiborza: Let's move this into the outer catch and just rethrow here so we always get these instructions when having to paste.

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+  const rootFilename = `root.${isTS ? 'tsx' : 'jsx'}`;
+  const rootPath = path.join(process.cwd(), 'app', rootFilename);
+
+  try {
+    if (!fs.existsSync(rootPath)) {
+      clack.log.warn(
+        `Could not find ${chalk.cyan(
+          rootFilename,
+        )}. Skipping root route instrumentation.`,
+      );
```
andreiborza: m: Same as above, I think we could ask users if they want to try rerunning reveal and if that still didn't work then print out what they have to add to the root file.

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+    const content = fs.readFileSync(rootPath, 'utf8');
+
+    // Add Sentry import if not present
+    let updatedContent = content;
+    if (!content.includes('import * as Sentry from "@sentry/react-router"')) {
+      updatedContent = `import * as Sentry from "@sentry/react-router";\nimport { isRouteErrorResponse } from "react-router";\n\n${updatedContent}`;
+    }
+
+    // Add ErrorBoundary if not present
+    if (!content.includes('export function ErrorBoundary')) {
```
andreiborza: m: I think we should also check other types of exports, i.e. `export const ErrorBoundary`

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+      debug('Failed to write root file:', writeError);
+      clack.log.warn(
+        `Failed to automatically update ${chalk.cyan(rootFilename)}.`,
+      );
+
+      await showCopyPasteInstructions({
+        filename: rootFilename,
+        codeSnippet: ERROR_BOUNDARY_TEMPLATE,
+        hint: 'Add this ErrorBoundary to your root component',
+      });
```
andreiborza: Let's move this to the other catch so we can show the snippet in all error cases.

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+    fs.writeFileSync(instrumentationPath, content);
+    clack.log.success(`Created ${chalk.cyan('instrumentation.server.mjs')}.`);
+    return instrumentationPath;
+  } catch (error) {
+    debug('Failed to create server instrumentation file:', error);
+    clack.log.error(
+      `Failed to create ${chalk.cyan(
+        'instrumentation.server.mjs',
+      )}. Please create it manually.`,
+    );
```
andreiborza: m: Can we have a pasteable snippet here too?

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+      debug(`Error processing server file ${serverFile}:`, error);
+      clack.log.warn(
+        `Failed to read ${chalk.cyan(
+          serverFile,
+        )}. Checking next server file...`,
+      );
+      // Continue to next file instead of returning false immediately
+    }
+  }
+
```
andreiborza, line 272: m: If we get to this point, we should print a pasteable snippet.

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+          `Failed to automatically update ${chalk.cyan(serverFile)}.`,
+        );
+        // Continue to next file instead of returning false immediately
+      }
+    } catch (error) {
+      debug(`Error processing server file ${serverFile}:`, error);
+      clack.log.warn(
+        `Failed to read ${chalk.cyan(
+          serverFile,
+        )}. Checking next server file...`,
```
andreiborza: m: I think we shouldn't surface this to the user, it's not actionable.

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+        clack.log.success(
+          `Updated ${chalk.cyan(serverFile)} with instrumentation import.`,
+        );
+        return true;
+      } catch (writeError) {
+        debug('Failed to write server file:', writeError);
+        clack.log.warn(
+          `Failed to automatically update ${chalk.cyan(serverFile)}.`,
+        );
+        // Continue to next file instead of returning false immediately
```
andreiborza: m: I think we should change this logic to not continue.

We found a file so it's likely the actual file users are using and continuing on doesn't seem fruitful.

We should bale out of the loop and show a pasteable snippet instead for the error case.

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+  const serverEntryFilename = `entry.server.${isTS ? 'tsx' : 'jsx'}`;
+  const serverEntryPath = path.join(process.cwd(), 'app', serverEntryFilename);
+
+  try {
+    if (!fs.existsSync(serverEntryPath)) {
+      clack.log.warn(
+        `Could not find ${chalk.cyan(
+          serverEntryFilename,
+        )}. Skipping server entry instrumentation.`,
+      );
```
andreiborza: m: As above, we could give users the option to reveal again and if that fails show a pasteable snippet.

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+      debug('Failed to write server entry file:', writeError);
+      clack.log.warn(
+        `Failed to automatically update ${chalk.cyan(serverEntryFilename)}.`,
+      );
+
+      await showCopyPasteInstructions({
+        filename: serverEntryFilename,
+        codeSnippet: sentryServerCode,
+        hint: 'Add this error handling to your server entry file',
+      });
```
andreiborza: m: Let's move this to the outer catch and rethrow here.

src/react-router/templates.ts
```diff
@@ -0,0 +1,115 @@
...
+  dsn: string,
+  enableTracing: boolean,
+  enableReplay: boolean,
+  enableLogs: boolean,
+) => {
+  const integrations = [];
+
+  if (enableTracing) {
+    integrations.push(
+      'browserTracingIntegration({\n      useEffect,\n      useLocation,\n      useNavigate\n    })',
```
andreiborza: m: I think we're supposed to use the `reactRouterTracingIntegration` instead.

andreiborza: @onurtemizkan the integration also doesn't need the hooks to be passed in

src/react-router/react-router-wizard.ts
```diff
@@ -0,0 +1,259 @@
...
+      await configureVitePlugin({
+        orgSlug: selectedProject.organization.slug,
+        projectSlug: selectedProject.slug,
+        url: sentryUrl,
+        selfHosted,
+        authToken,
+      });
+    } catch (e) {
+      clack.log.warn(
+        `Could not configure Vite plugin for sourcemap uploads. Please configure it manually using instructions from https://docs.sentry.io/platforms/javascript/guides/react-router/sourcemaps/`,
```
chargome:
```suggestion
        `Could not configure Vite plugin for sourcemap uploads. Please configure it manually using instructions from https://docs.sentry.io/platforms/javascript/guides/react-router/#source-maps-upload`,
```

src/react-router/sdk-setup.ts
```diff
@@ -0,0 +1,335 @@
...
+        )}. Checking next server file...`,
+      );
+      // Continue to next file instead of returning false immediately
+    }
+  }
+
+  return false;
+}
+
+export async function instrumentSentryOnEntryServer(
```
chargome, line 293: h: We also need to update the default export here (`handleRequest`), otherwise there will be no distributed tracing

onurtemizkan:
@chargome, @andreiborza - Added the source map upload. -> https://github.com/getsentry/sentry-wizard/pull/1076/commits/e161db283463b3a8d942930e4cf5ea8a2e9d2b85

lib/Constants.ts
```diff
@@ -57,6 +58,8 @@ export function getIntegrationDescription(type: string): string {
       return 'Next.js';
     case Integration.remix:
       return 'Remix';
+    case Integration.reactRouter:
+      return 'React Router';
```
chargome:
```suggestion
      return 'React Router (framework)';
```

PR #1077, opened 2025-08-28
feat(sveltekit): Add support for SDK setup with `instrumentation.server.ts`

Lms24:
- **initial boilerplate**
- **test expectation**
- **ast mod progress**
- **add working and tested svelte config modification**
- **add instrumentation.server.ts code insertion**
- **various improvements**
- **conditionally inject init into server**
- **.**
- **template tests**

### DIFF
--- a/src/sveltekit/sdk-setup/setup.ts
+++ b/src/sveltekit/sdk-setup/setup.ts
@@ -1,7 +1,6 @@
 import type { ExportNamedDeclaration, Program } from '@babel/types';
 import * as fs from 'fs';
 import * as path from 'path';
-import * as url from 'url';
 import chalk from 'chalk';
 
 import * as Sentry from '@sentry/node';
@@ -11,48 +10,24 @@ import clack from '@clack/prompts';
 // @ts-expect-error - magicast is ESM and TS complains about that. It works though
 import type { ProxifiedModule } from 'magicast';
 // @ts-expect-error - magicast is ESM and TS complains about that. It works though
-import { builders, generateCode, loadFile, parseModule } from 'magicast';
-// @ts-expect-error - magicast is ESM and TS complains about that. It works though
-import { addVitePlugin } from 'magicast/helpers';
-import { getClientHooksTemplate, getServerHooksTemplate } from './templates';
-import {
-  abortIfCancelled,
-  featureSelectionPrompt,
-  isUsingTypeScript,
-} from '../utils/clack';
-import { debug } from '../utils/debug';
-import { findFile, hasSentryContent } from '../utils/ast-utils';
+import { builders, generateCode, loadFile } from 'magicast';
+import { getClientHooksTemplate, getServerHooksTemplate } from '../templates';
+import { featureSelectionPrompt, isUsingTypeScript } from '../../utils/clack';
+import { findFile, hasSentryContent } from '../../utils/ast-utils';
 
 import * as recast from 'recast';
 import x = recast.types;
 import t = x.namedTypes;
-import { traceStep } from '../telemetry';
-
-const SVELTE_CONFIG_FILE = 'svelte.config.js';
-
-export type PartialSvelteConfig = {
-  kit?: {
-    files?: {
-      hooks?: {
-        client?: string;
-        server?: string;
-      };
-      routes?: string;
-    };
-  };
-};
-
-type ProjectInfo = {
-  dsn: string;
-  org: string;
-  project: string;
-  selfHosted: boolean;
-  url: string;
-};
+import type { KitVersionBucket } from '../utils';
+import type { PartialBackwardsForwardsCompatibleSvelteConfig } from './svelte-config';
+import { ProjectInfo } from './types';
+import { modifyViteConfig } from './vite';
+import { modifyAndRecordFail } from './utils';
 
 export async function createOrMergeSvelteKitFiles(
   projectInfo: ProjectInfo,
-  svelteConfig: PartialSvelteConfig,
+  svelteConfig: PartialBackwardsForwardsCompatibleSvelteConfig,
+  kitVersionBucket: KitVersionBucket,
 ): Promise<void> {
   const selectedFeatures = await featureSelectionPrompt([
     {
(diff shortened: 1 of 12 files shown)


Lms24:
Hmm I really dislike that our e2e tests don't contribute to our coverage calculation. Maybe we can find some way to run them within the same process as the test. Might be a nice addition for `clifty`. Though not sure how this would work yet. 

vitest and child process coverage doesn't seem to go anywhere (https://github.com/vitest-dev/vitest/discussions/2911)

e2e-tests/test-applications/sveltekit-hooks-test-app/.gitignore
Lms24, line 1: I had to rename the `sveltekit` test app to `sveltekit-hooks` to avoid ambiguity with `sveltekit-tracing`.

src/sveltekit/sdk-setup/setup.ts
Lms24, line 1: Decided to split this file up since there's quite a lot going on in it. Hence the various new files in the `sdk-setup` dir

src/sveltekit/sdk-setup/setup.ts
```diff
@@ -83,52 +69,95 @@ export async function createOrMergeSvelteKitFiles(
   // full file paths with correct file ending (or undefined if not found)
   const originalClientHooksFile = findFile(clientHooksPath);
   const originalServerHooksFile = findFile(serverHooksPath);
+  const originalInstrumentationServerFile = findFile(
+    path.resolve(process.cwd(), 'src', 'instrumentation.server'),
```
andreiborza, line 73: q: Is this always in `src/`? Can users not potentially have a different structure?

src/sveltekit/sdk-setup/setup.ts
```diff
@@ -83,52 +69,95 @@ export async function createOrMergeSvelteKitFiles(
...
+  }
+
   Sentry.setTag(
-    'client-hooks-file-strategy',
-    originalClientHooksFile ? 'merge' : 'create',
+    'server-hooks-file-strategy',
+    originalServerHooksFile ? 'merge' : 'create',
   );
-  if (!originalClientHooksFile) {
-    clack.log.info('No client hooks file found, creating a new one.');
```
andreiborza, line 98: q: Did you intentionally remove these? Are these not useful to users?

src/sveltekit/sdk-setup/setup.ts
```diff
@@ -278,6 +345,79 @@ Skipping adding Sentry functionality to.`,
...
+    instrumentationServerFilePath,
+  );
+  const filename = path.basename(instrumentationServerFilePath);
+
+  if (hasSentryContent(originalInstrumentationServerMod.$ast as t.Program)) {
+    // We don't want to mess with files that already have Sentry content.
+    // Let's just bail out at this point.
+    clack.log.warn(
+      `File ${chalk.cyan(filename)} already contains Sentry code.
+Skipping adding Sentry functionality to.`,
```
andreiborza:
```suggestion
Skipping adding Sentry functionality to it.`,
```

src/sveltekit/sdk-setup/svelte-config.ts
```diff
@@ -0,0 +1,397 @@
...
+  }
+}
+
+export async function enableTracingAndInstrumentation(
+  originalSvelteConfig: PartialBackwardsForwardsCompatibleSvelteConfig,
+  enableTracing: boolean,
+) {
+  const hasTracingEnabled = originalSvelteConfig.kit?.experimental?.tracing;
+  const hasInstrumentationEnabled =
+    originalSvelteConfig.kit?.experimental?.instrumentation;
```
andreiborza, line 78: q: What happens if they have `instrumentation: { server: false }` in their config?

src/sveltekit/sveltekit-wizard.ts
```diff
@@ -87,6 +88,58 @@ export async function runSvelteKitWizardWithTelemetry(
...
+        options: [
+          {
+            label: "No, I'll upgrade SvelteKit first",
+            hint: 'Recommended',
+            value: 'exit-to-upgrade',
+          },
+          {
+            label: "I'm already on SvelteKit >=2.31.0",
+            hint: 'Sorry, my bad!',
+            value: 'install-with-kit-tracing',
```
andreiborza, line 122: That's really nice! Good option.

test/apple/macos-system-helper.test.ts
```diff
@@ -64,7 +64,7 @@ describe('MacOSSystemHelpers', () => {
     );
   });
 
-  describe('findDeveloperDirectoryPath', () => {
+  describe.skip('findDeveloperDirectoryPath', () => {
```
andreiborza: q: Intentional?

test/sveltekit/sdk-setup/svelte-config.test.ts
```diff
@@ -0,0 +1,574 @@
...
+
+          experimental: {
+            tracing: {
+              server: false,
+            },
+
+            instrumentation: {
+              server: true,
+            },
+          },
```
andreiborza, line 178: q: Can we also add a case for 

`tracing: { server: true }, instrumentation: { server: false }`?

andreiborza (commented):
Really nice!

src/sveltekit/sdk-setup/setup.ts
Lms24: Yes, this is always in `src`. FYI, Svelte is generally moving away from making file names and paths that should be within `src` configurable. The reason is that they want to enforce consistency on how a svelteKit project should look and which files to expect (https://github.com/sveltejs/kit/issues/14151). Although, let's see how this progresses since some people are complaining in https://github.com/sveltejs/kit/discussions/14240.

src/sveltekit/sdk-setup/setup.ts
Lms24: Yes, I intentionally removed them. The pattern previously was that clack would lok

```
> [info] No client hooks file found, creating a new one
> [success] created `hooks.client.ts`
```

Announcing _that_ we're going to do something isn't a pattern we usually have. Also, the wording makes it slightly sound like this is something unexpected the wizard encounters. Which it isn't. Chances are quite high that users don't yet have hooks files. To avoid confusion and keep output short, I decided to remove these logs. 

Happy to bring back if you prefer :)

src/sveltekit/sdk-setup/svelte-config.ts
Lms24: It would fail the check if everything is already enabled and we'd set it `true` for them.

test/apple/macos-system-helper.test.ts
Lms24: nope, thank you 🙏

src/sveltekit/sdk-setup/setup.ts
andreiborza: No, makes sense to remove this given that people who run the wizard probably don't have hook files yet. Nice change.

test/sveltekit/sdk-setup/svelte-config.test.ts
Lms24: This is tested here: https://github.com/getsentry/sentry-wizard/blob/2862039506d34c5cb665fdc4dc53157e533c18d8/test/sveltekit/sdk-setup/svelte-config.test.ts#L311-L355

Though only for both options, not for one of them individually. I'll add two more tests, just to be sure and cover some more code paths.

Lms24: [e923fee](https://github.com/getsentry/sentry-wizard/pull/1077/commits/e923feee5ad4a5bdf485aa4f6ca17f632bee23be)

andreiborza (approved):
Nice!

PR #1085, opened 2025-09-17
Add logs to Android projects

stefanosiano:
Logs are now added by default to Android projects

Closes https://github.com/getsentry/sentry-wizard/issues/1037

### DIFF
--- a/src/android/android-wizard.ts
+++ b/src/android/android-wizard.ts
@@ -9,6 +9,7 @@ import { traceStep, withTelemetry } from '../telemetry';
 import {
   CliSetupConfig,
   abort,
+  abortIfCancelled,
   addSentryCliConfig,
   confirmContinueIfNoOrDirtyGitRepo,
   getOrAskForProjectData,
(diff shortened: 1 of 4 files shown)


src/android/android-wizard.ts
```diff
@@ -72,6 +73,21 @@ async function runAndroidWizardWithTelemetry(
...
+    clack.confirm({
+      message:
+        'Do you want to enable Logs? (See https://docs.sentry.io/platforms/android/logs/)',
+    }),
+  );
+  Sentry.setTag('enable-logs', enableLogs);
+
+  if (enableLogs) {
+    clack.log.info(
+      'Logs will be enabled with default settings. You can send logs using the Sentry.logger() APIs.',
```
romtsn:
```suggestion
      'Logs will be enabled with default settings. You can send logs using the Sentry.logger() APIs or use one of the integrations: https://docs.sentry.io/platforms/android/logs/#integrations',
```

PR #1098, opened 2025-10-14
Updates the bundle script in React Native to match the one from the docs 

alwx:
Updates the bundle script in React Native to match the one from the docs and support the changes done here: https://github.com/getsentry/sentry-react-native/pull/5221


antonis:
@Lms24 We'll push an update for a failing check in a follow-up PR

PR #1115, opened 2025-11-17
feat(all): Add --spotlight support

betegon:
This PR adds support for `--spotlight`. A boolean flag that enabled, will skip the step where we ask the user for a sentry account and automatically add them a DSN (required step). For spotlight we don't need a DSN, se we're skipping it now.

Inspired by 
- https://github.com/getsentry/spotlight/issues/273
needed to close:
-  https://github.com/getsentry/spotlight/issues/987
- https://github.com/getsentry/spotlight/issues/1117

### DIFF
--- a/test/sveltekit/templates.test.ts
+++ b/test/sveltekit/templates.test.ts
@@ -348,8 +348,6 @@ describe('getInstrumentationServerTemplate', () => {
         // Enable logs to be sent to Sentry
         enableLogs: true,
 
-        // uncomment the line below to enable Spotlight (https://spotlightjs.com)
-        // spotlight: true,
       });"
     `);
   });
@@ -369,8 +367,6 @@ describe('getInstrumentationServerTemplate', () => {
         // Enable logs to be sent to Sentry
         enableLogs: true,
 
-        // uncomment the line below to enable Spotlight (https://spotlightjs.com)
-        // spotlight: true,
       });"
     `);
   });
@@ -390,8 +386,6 @@ describe('getInstrumentationServerTemplate', () => {
         tracesSampleRate: 1.0,
 
 
-        // uncomment the line below to enable Spotlight (https://spotlightjs.com)
-        // spotlight: true,
       });"
     `);
   });
@@ -409,8 +403,6 @@ describe('getInstrumentationServerTemplate', () => {
         dsn: 'https://sentry.io/123',
 
 
-        // uncomment the line below to enable Spotlight (https://spotlightjs.com)
-        // spotlight: true,
       });"
     `);
   });


CHANGELOG.md
Lms24, line 1: l: let's revert the unrelated formatting changes here

e2e-tests/tests/sveltekit-hooks.test.ts
```diff
@@ -245,6 +245,7 @@ describe('Sveltekit', () => {
   // Enable sending user PII (Personally Identifiable Information)
   // https://docs.sentry.io/platforms/javascript/guides/sveltekit/configuration/options/#sendDefaultPii
   sendDefaultPii: true,
+
```
Lms24, line 248: l: any idea what caused this new line break? Ideally we can avoid it

e2e-tests/tests/sveltekit-tracing.test.ts
```diff
@@ -146,6 +146,7 @@ describe('Sveltekit with instrumentation and tracing', () => {
           // Enable sending user PII (Personally Identifiable Information)
           // https://docs.sentry.io/platforms/javascript/guides/sveltekit/configuration/options/#sendDefaultPii
           sendDefaultPii: true,
+
```
Lms24, line 149: same here

src/angular/codemods/main.ts
```diff
@@ -70,10 +72,11 @@ export function getInitCallArgs(
...
     logs: boolean;
   },
+  spotlightMode = false,
 ): InitCallArgs {
-  const initCallArgs: InitCallArgs = {
-    dsn,
-  };
+  const initCallArgs: InitCallArgs = {};
+
+  initCallArgs.dsn = dsn;
```
Lms24, line 79: super-l: any reason for this change? (I might just be missing it 😅)

src/nuxt/nuxt-wizard.ts
Lms24, line 1: q: don't we need to further adjust the nuxt wizard here to set `spotlight: true` and avoid setting the DSN? Totally fine with doing it iteratively in other PRs. Was just curious.

src/utils/clack/index.ts
```diff
@@ -984,7 +984,37 @@ export async function getOrAskForProjectData(
   selfHosted: boolean;
   selectedProject: SentryProjectData;
   authToken: string;
+  spotlightMode?: boolean;
```
Lms24, line 987: l: suggestion to make this a bit more typesafe:

We could use `spotlightMode` to retype this as a discriminated union:

```ts
type getOrAskForProjectDataReturnType = Promise<{
  sentryUrl: string;
  selfHosted: boolean;
  selectedProject: SentryProjectData;
  authToken: string;
  spotlightMode: false // or undefined, shouldn't matter
} | {spotlightMode: true }> 
```

(GH won't let me make a direct suggestion, so this will need some manual adjustment)

Meaning, any caller of this function has to check `spotlightMode` before being able to consume the returned project data. My thinking is this _should_ help us catch cases in which we'd e.g. rely on the DSN where in fact it is not available or vice versa. It should also allow us to get rid of the dummy data which I don't think we need (but might be missing something).

WDYT? No super strong opinion but given that this function is consumed by _every_ wizard, I think it makes sense for it to be as type-safe as possible.

src/utils/clack/index.ts
```diff
@@ -984,7 +984,37 @@ export async function getOrAskForProjectData(
...
 }> {
+  // Spotlight mode: Skip authentication and use local development setup
+  if (options.spotlight) {
+    clack.log.info(
+      chalk.cyan(
+        'Spotlight mode enabled! Setting up for local development without Sentry authentication.',
+      ),
```
Lms24, line 994: comment/FYI: Usually we use `cyan` only for file names or small code snippets in a message (see [here](https://develop.sentry.dev/sdk/expected-features/setup-wizards/#colors)). I get though why we want this message to stand out, so I'm fine with leaving it it as-is.

src/sveltekit/templates.ts
```diff
@@ -43,6 +44,13 @@ ${
   // Enable sending user PII (Personally Identifiable Information)
   // https://docs.sentry.io/platforms/javascript/guides/sveltekit/configuration/options/#sendDefaultPii
   sendDefaultPii: true,
+${
```
Lms24, line 47: ahh I guess that's where the newline is coming from (see my comments above). So we just need to ensure we don't print a newline here if !spotlightMode.

src/sveltekit/templates.ts
```diff
@@ -82,9 +91,13 @@ ${
   // Enable sending user PII (Personally Identifiable Information)
   // https://docs.sentry.io/platforms/javascript/guides/sveltekit/configuration/options/#sendDefaultPii
   sendDefaultPii: true,
-
-  // uncomment the line below to enable Spotlight (https://spotlightjs.com)
-  // spotlight: import.meta.env.DEV,
+${
```
Lms24, line 94: same here

test/angular/angular-wizard.test.ts
```diff
@@ -131,5 +131,21 @@ describe('angular-wizard', () => {
         }
       `);
     });
+
+    it('uses dummy DSN "http://test:0000" in spotlight mode', () => {
```
Lms24, line 135: h: just to confirm: the `http://test:0000` DSN is not actually injected into user files, correct?

If so, I'd strongly suggest _not_ to add this DSN but instead set an empty string or `undefined`. If no (valid) DSN is found, the SDK will fall back to checking `spotlight` anyway, so this just more clearly communicates the intention that the SDK is set up purely for spotlight.

betegon:
Thanks for the review @Lms24 , i'm going to close this PR and submit them one by one then.

PR #1119, opened 2025-11-19
feat(nextjs): Add --spotlight support

betegon:
context #1115.

### DIFF
--- a/src/nextjs/nextjs-wizard.ts
+++ b/src/nextjs/nextjs-wizard.ts
@@ -715,11 +715,7 @@ async function createOrMergeNextJsFiles(
 
       const successfullyCreated = await createNewConfigFile(
         newInstrumentationClientHookPath,
-        getInstrumentationClientFileContents(
-          dsn,
-          selectedFeatures,
-          spotlight,
-        ),
+        getInstrumentationClientFileContents(dsn, selectedFeatures, spotlight),
       );
 
       if (!successfullyCreated) {


src/android/android-wizard.ts
```diff
@@ -70,8 +70,16 @@ async function runAndroidWizardWithTelemetry(
...
-  const { selectedProject, selfHosted, sentryUrl, authToken } =
-    await getOrAskForProjectData(options, 'android');
+  const projectData = await getOrAskForProjectData(options, 'android');
+
+  if (projectData.spotlight) {
+    clack.log.warn('Spotlight mode is not yet supported for Android.');
+    clack.log.info('Spotlight is currently only available for Next.js.');
+    await abort('Exiting wizard', 0);
+    return;
+  }
```
betegon, line 80: for all wizard not supporting spotlight yet, we return some logs and exit the wizard in case someone runs the wizard with `--spotlight`.

wdyt @Lms24 ? we're going to add support for the other ones, but until then. 

This is necessary as we now do the union type based on spotlight = true/false as you suggested

src/nextjs/nextjs-wizard.ts
```diff
@@ -679,8 +722,9 @@ async function createOrMergeNextJsFiles(
       const successfullyCreated = await createNewConfigFile(
         newInstrumentationClientHookPath,
         getInstrumentationClientFileContents(
-          selectedProject.keys[0].dsn.public,
+          effectiveDsn,
           selectedFeatures,
+          spotlight,
         ),
       );
```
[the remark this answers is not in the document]
betegon: fixed, THANKS SENTRY

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- Add --spotlight support
+  ([#1115](https://github.com/getsentry/sentry-wizard/pull/1115))
```
Lms24: probably still a leftover from #1115:

```suggestion
- feat(nextjs): Add --spotlight support
  ([#1119](https://github.com/getsentry/sentry-wizard/pull/1119))
```

e2e-tests/tests/nextjs-15.test.ts
Lms24, line 1: thanks for writing the e2e test! We still need to register it so that it also runs in CI: 

https://github.com/getsentry/sentry-wizard/blob/62c6868d3e52a14050f80febd3436e797d2ed53c/.github/workflows/build.yml#L135-L154

src/android/android-wizard.ts
Lms24: Got it! Yeah makes sense to me. If we plan on adding support for the respective platforms quickly, I'm fine with leaving this in. Otherwise, a dedicated helper might make sense but let's not overcomplicate things :)

src/nextjs/nextjs-wizard.ts
```diff
@@ -116,12 +118,46 @@ export async function runNextjsWizardWithTelemetry(
...
+    selfHosted = false;
+    sentryUrl = '';
+    authToken = '';
+    // Create a minimal project structure for type compatibility
+    selectedProject = {
+      id: '',
+      slug: '',
+      organization: { id: '', slug: '', name: '' },
+      keys: [{ dsn: { public: '' } }],
+    };
```
Lms24, line 139: l/feel free to ignore: Probably fine to extract if we indeed need the same data in all wizards supporting spotlight. Feel free to disregard for now and adjust as necessary in future PRs.

src/nextjs/nextjs-wizard.ts
```diff
@@ -428,7 +466,11 @@ async function createOrMergeNextJsFiles(
   selfHosted: boolean,
   sentryUrl: string,
   sdkConfigOptions: SDKConfigOptions,
+  spotlight = false,
+  dsn?: string,
```
Lms24: l: do we need the dsn param here? Can't we check via `spotlight` or just fall back to `''` ?
Given above we set a dummy `selectedProject` we should fall back to '' anyway, right?

src/sourcemaps/sourcemaps-wizard.ts
Lms24, line 1: m: I don't think support for --spotlight in the source maps wizard would make sense. Can we just ignore the option here? I wouldn't even logs something out or at least not exit the process. Users who run this come with the clear expectation of configuring source maps upload.

Lms24 (commented):
Thanks for splitting this up! Makes it easier to review/not miss stuff. We can likely iterate faster on the next PRs. 

My final question: If you run `npm run build` in a NextJS project after running the wizard with `--spotlight`, do you get any error messages? I'm specifically thinking about source maps upload or errors regarding missing auth tokens. Just not 100% sure about the behaviour. If not totally fine. If yes, let's recheck what we can do about them.

src/android/android-wizard.ts
betegon: yeah, I'll add support to the other ones ASAP

e2e-tests/tests/nextjs-15.test.ts
betegon: i think it's not needed as i put the test under the `NextJS-15` file so it should run

src/nextjs/nextjs-wizard.ts
betegon: i'll add this in with the following wizard additions!

e2e-tests/tests/nextjs-15.test.ts
Lms24: oh yes, of course. Sorry for the confusion!

betegon:
@Lms24 added your suggestions and answered a couple of commments.

Apart from that, there's no problem building a nextjs project after instrumenting it with the wizard. no problem with npm, pnpm or bun.

Lms24 (approved):
:shipit:

PR #1123, opened 2025-11-24
chore: Update Contributing guide with publishing instructions

Lms24:


CONTRIBUTING.md
```diff
@@ -112,3 +112,15 @@ type:
...
+### Publishing a Release
+
+The wizard is released via our usual Craft and `getsetntry/publish` flow:
+
+1. Make sure the changelog is in a good and complete state.
+   If not, open a PR against `master` to [update it](#updating-the-changelog)
+2. Determine the next version (we follow semver versioning)
+3. In GitHub, go to "Actions" and start the "Release" action
+4. Enter the version to be released.
+   If you're publishing a normal release, no other input fields are required.
```
andreiborza: Hard to tell from this what a "normal" release entails and when its not normal. Maybe we can phrase this around releasing a version for a current major?

Lms24: it's not even different for a new major. The only exception would be experimental releases which shouldn't merge back to master. Will rephrase :)

Lms24:
```suggestion
4. Enter the version to be released.
   If you're publishing a stable release, no other input fields are required.
   If you release from a branch (e.g. for alpha/beta releases), set the branch as the target branch.
```

PR #1126, opened 2025-11-28
chore(deps): Upgrade @sentry/node from v7 to v10.27.0

sergical:
Closes DEVEX-184


.cursor/rules/wizard.mdc
```diff
@@ -0,0 +1,221 @@
+---
+alwaysApply: true
```
sergical, line 2: can remove this, but good to have as reference, might need to be adjusted to be more "rule"-like vs a doc

Lms24: we can leave it in from my PoV, sounds like a good idea!

src/telemetry.ts
```diff
@@ -22,53 +11,53 @@ export async function withTelemetry<F>(
...
   }
 }
 
-function createSentryInstance(enabled: boolean, integration: string) {
-  const client = new NodeClient({
+function initSentry(
+  enabled: boolean,
+  integration: string,
+): Sentry.NodeClient | undefined {
+  const client = Sentry.init({
```
Lms24, line 60: h: I can't leave a comment at the right place but gonna flag it here instead:

- we should pass in `defaultIntegrations: false, integrations: [Sentry.httpIntegration()]`. Previously, when directly creating the client, `integrations` was the complete array of integrations that would get initialized. Here we now call `Sentry.init` where without explicitly setting `defaultIntegrations`, it would init all its default integrations. We only want the `httpIntegration` though.

src/telemetry.ts
```diff
@@ -96,40 +82,36 @@ function createSentryInstance(enabled: boolean, integration: string) {
       delete event.server_name; // Server name might contain PII
       return event;
     },
-
-    transport: makeNodeTransport,
-
-    debug: true,
```
Lms24, line 102: huh interesting that `debug: true` didn't actually log something to the console 🤔 anyway, not relevant anymore 😅

src/telemetry.ts
sergical: thank you! resolved

src/utils/clack/index.ts
```diff
@@ -119,17 +119,22 @@ export const propertiesCliSetupConfig: Required<CliSetupConfig> = {
...
   // see: https://docs.sentry.io/product/insights/overview/metrics/#failure-rate
-  sentryTransaction?.setStatus(status === 0 ? 'cancelled' : 'aborted');
-  sentryTransaction?.finish();
-  const sentrySession = sentryHub.getScope().getSession();
+  if (rootSpan) {
+    rootSpan.setStatus({
+      code: status === 0 ? 1 : 2, // 1 = ok/cancelled, 2 = error/aborted
+      message: status === 0 ? 'cancelled' : 'aborted',
+    });
+    rootSpan.end();
```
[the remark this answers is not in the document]
Lms24: I think 'ok' is fine here, given cancelled is not an error but a user decision.

PR #1127, opened 2025-11-28
feat(nextjs): Add Sentry.logger examples to example pages when logs are enabled 

sergical:


src/nextjs/templates.ts
```diff
@@ -261,17 +261,34 @@ export function getSentryExamplePageContents(options: {
...
   isTypeScript?: boolean;
+  logsEnabled?: boolean;
 }): string {
   const issuesPageLink = options.selfHosted
     ? `${options.sentryUrl}organizations/${options.orgSlug}/issues/?project=${options.projectId}`
     : `https://${options.orgSlug}.sentry.io/issues/?project=${options.projectId}`;
 
+  const loggerDeclaration = options.logsEnabled
+    ? `
+const { logger } = Sentry;
```
andreiborza: l: In our docs we generally use the logger directly via `Sentry.logger.info` calls. Maybe we should keep it like that for consistency sake, wdyt?

That being said, I did notice that the AI rules you can copy in the docs still have it via `const { logger } = Sentry`.

sergical:
> Thanks for adding this!
> 
> Are we going to add this to all the wizards? :)

I think it makes sense, i'll add a ticket to do the rest

PR #1128, opened 2025-11-30
feat(nextjs): Add Biome support and fix linting issues in generated code

sergical:


src/nextjs/nextjs-wizard.ts
```diff
@@ -427,14 +428,16 @@ export async function runNextjsWizardWithTelemetry(
...
-
   // Offer optional project-scoped MCP config for Sentry with org and project scope
   await offerProjectScopedMcpConfig(
     selectedProject.organization.slug,
     selectedProject.slug,
   );
 
+  // Run formatters as the last step to fix any formatting issues in generated/modified files
+  await runPrettierIfInstalled({ cwd: undefined });
+  await runBiomeIfInstalled({ cwd: undefined });
```
andreiborza: l/m: I think we could extract this out into a `runFormatters` helper, given that we'll probably also want to support oxfmt at some point, wdyt?

src/nextjs/nextjs-wizard.ts
```diff
@@ -910,7 +913,25 @@ async function createOrMergeNextJsFiles(
...
           );
 
-          const newCode = mod.generate().code;
+          let newCode = mod.generate().code;
+
+          // Post-process to fix formatting issues that magicast doesn't handle
+          // (needed for Biome/ESLint compatibility):
+          // 1. Add spaces inside import braces: {withSentryConfig} -> { withSentryConfig }
+          newCode = newCode.replace(
+            /import\s*{(\w+)}\s*from/g,
```
andreiborza, line 922: m: This wouldn't catch some cases we might use, e.g.

* `import Foo, {Bar} from 'baz'`
* `import {Foo, Bar} from 'baz'`

We might want to widen the regex or run through several such replaces.

src/utils/clack/index.ts
```diff
@@ -805,6 +805,91 @@ export async function runPrettierIfInstalled(opts: {
...
+          () => {
+            // Ignore errors, just continue
+            resolve();
+          },
+        );
+      });
+
+      await new Promise<void>((resolve) => {
+        childProcess.exec(
+          `npx @biomejs/biome check --write --unsafe ${changedOrUntrackedFiles}`,
```
andreiborza, line 994: q/m: When would we want to run `--unsafe` formats?

src/nextjs/nextjs-wizard.ts
sergical: good call!

src/utils/clack/index.ts
sergical: Biome marks some things `--unsafe` as it applies potentially behavior-changing fixes like removing unused imports. in this instance i think it's fine https://biomejs.dev/linter/#why-does-rule-x-have-an-unsafe-fix-it-seems-safe-to-me

src/nextjs/nextjs-wizard.ts
sergical: added several regexes for different import patterns, good catch!

src/utils/clack/index.ts
andreiborza: Hm, seems a bit loose to me but I guess people can opt out of running the formatter.

PR #1134, opened 2025-12-04
feat(nuxt): Add deletion glob to delete sourcemaps

s1gr1d:
This adds a source maps deletion glob for Nuxt as client source maps need to be explicitly enabled in Nuxt.


andreiborza:
Please don't forget updating the changelog, otherwise LGTM

Lms24 (approved):
Let's fix the comma (see Seer review) but otherwise LGTM!

PR #1138, opened 2025-12-09
chore(publish): Use fully automatic changelog generation

Lms24:
- remove the Danger workflow
- switch from deprecated `changelogPolicy` to `changelog.policy` in `.craft.yml` and bump the minimum craft version to [2.12.0](https://github.com/getsentry/craft/releases/tag/2.12.0)
- add `release.yml` to customize changelog generation. We use the default regexes but add label support to override and skip PRs.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,4 +1,4 @@
-minVersion: 2.13.0
+minVersion: 2.17.0
 
 changelog: 
   policy: auto
--- a/.github/workflows/changelog-preview.yml
+++ b/.github/workflows/changelog-preview.yml
@@ -6,5 +6,5 @@ on:
 
 jobs:
   changelog-preview:
-    uses: getsentry/craft/.github/workflows/changelog-preview.yml@0ca5e6d874c893fbaa8605e24e2835113f29f119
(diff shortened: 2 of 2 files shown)


.github/release.yml
```diff
@@ -0,0 +1,30 @@
+changelog:
+  exclude:
+    labels:
+      - skip-changelog
```
BYK, line 4: Ah, we should definitely add this to our default template.

.github/release.yml
```diff
@@ -0,0 +1,30 @@
...
+  categories:
+    - title: Breaking Changes
```
BYK, line 9: You may wanna add a new category, "Important Changes" to pull out them with labels. If you place it at the top, it will be displayed at the top and will pick the PRs/commits first (before other categories)

Lms24: added, thanks!

.github/release.yml
```diff
@@ -0,0 +1,33 @@
+changelog:
+  exclude:
+    labels:
+      - skip-changelog
+  categories:
```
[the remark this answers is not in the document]
Lms24: I'm not sure what to do with this comment. These two files have technically nothing to do with eath other.

PR #1149, opened 2025-12-16
feat(cloudflare): Enable update of the wrangler file

### LINKED ISSUE #1148: Cloudflare update wrangler configs
To ensure better support in the Cloudflare wizard we like to update the wrangler config automatically, so the user doesn't have to.

JPeer264:
(closes #1148) 

This PR will update the wrangler config automagically. The wrangler config can exist in `.toml`, `.json` and `.jsonc` ([source](https://developers.cloudflare.com/workers/wrangler/configuration/)) that is why I added ~`smol-toml` (BSD-3) and~ `jsonc-parser` (MIT).

These are the new steps:

1. Wrangler config does not exist: We create a minimal jsonc config [with the minimum required keys](https://developers.cloudflare.com/workers/wrangler/configuration/#inheritable-keys) (also jsonc is recommended for new projects)
    - On top we instruct that the `name` and the `main` line has to be changed 
2. Wrangler config does exist in either format: It adds Sentry specifics
    - For .toml we add a copy paste snippet

For now we add:

- `nodejs_als` instead of `nodejs_compat` to keep the bundle size smaller, and it is not yet required
- `compatibility_date` to the date of creation (suggested by the [Cloudflare docs](https://developers.cloudflare.com/workers/configuration/compatibility-dates/#setting-compatibility-date))
- And our required `version_metadata`

## Drawbacks

~Toml parsers are not quite popular and are not really feature complete, that is why toml files would loose all comments, once we update them (another option would be to not support `.toml` at all and give only a hint)~ (the toml parser has been removed)

### DIFF
--- a/src/cloudflare/wrangler/get-entry-point-from-wrangler-config.ts
+++ b/src/cloudflare/wrangler/get-entry-point-from-wrangler-config.ts
@@ -1,7 +1,7 @@
 import fs from 'node:fs';
 import path from 'node:path';
-import { parse } from 'jsonc-parser';
 import { findWranglerConfig } from './find-wrangler-config';
+import { parseJsonC, getObjectProperty } from '../../utils/ast-utils';
 
 /**
  * Reads the main entry point from the wrangler config file
@@ -28,9 +28,23 @@ export function getEntryPointFromWranglerConfig(): string | undefined {
     case '.json':
     case '.jsonc':
       try {
-        const config = parse(configContent) as { main?: string };
+        const { jsonObject } = parseJsonC(configContent);
 
-        return config.main;
+        if (!jsonObject) {
+          return undefined;
+        }
+
+        const mainProperty = getObjectProperty(jsonObject, 'main');
+
+        if (
+          (mainProperty?.value.type === 'StringLiteral' ||
+            mainProperty?.value.type === 'Literal') &&
+          typeof mainProperty.value.value === 'string'
+        ) {
+          return mainProperty.value.value;
+        }
+
+        return undefined;
       } catch {
         return undefined;
       }
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
     "glob": "9.3.5",
     "inquirer": "^6.2.0",
     "js-yaml": "^4.1.1",
-    "jsonc-parser": "^3.3.1",
     "magicast": "^0.2.10",
     "opn": "^5.4.0",
     "read-env": "^1.3.0",
--- a/src/cloudflare/wrangler/update-wrangler-config.ts
+++ b/src/cloudflare/wrangler/update-wrangler-config.ts
@@ -2,10 +2,18 @@
 import clack from '@clack/prompts';
 import chalk from 'chalk';
 import fs from 'node:fs';
-import * as jsonc from 'jsonc-parser';
 import path from 'node:path';
 import { findWranglerConfig } from './find-wrangler-config';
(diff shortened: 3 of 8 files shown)


src/cloudflare/wrangler/get-entry-point-from-wrangler-config.ts
```diff
@@ -0,0 +1,41 @@
...
+    case '.toml': {
+      const mainMatch = configContent.match(/^main\s*=\s*["'](.+)["']/m);
+
+      return mainMatch ? mainMatch[1] : undefined;
+    }
+
+    case '.json':
+    case '.jsonc':
+      try {
+        const config = parse(configContent) as { main?: string };
```
Lms24: l/m: We could try to reuse our existing `parseJsonC` helper and avoid adding another dependency. I'll admit, the parsing strategy is a "bit" hacky, so if there are any obvious drawbacks, feel free to continue using `jsonc-parser`.

Lms24 (approved):
Nice! 

> Toml parsers are not quite popular and are not really feature complete, that is why toml files would loose all comments, once we update them (another option would be to not support .toml at all and give only a hint)

If we discard user-created comments, I'd rather tend towards showing copy/paste-able instructions in the console. We have a `showCopyPasteInstructions` and `makeCodeSnippet` helpers that should make this _fairly_ user friendly. 

Logaf-L though, so happy to leave the decision up to you!

src/cloudflare/wrangler/get-entry-point-from-wrangler-config.ts
JPeer264: There are 3 drawbacks which I found on the quick implementation, where I'm not 100% certain if it is worth the effort to fix - since `jsonc-parser` does have almost 30m weekly downloads and is maintained by Microsoft.

---

Drawback #1. We apparently remove whitespaces

Screenshot 2025-12-18 at 11 43 24

---

Drawback #2: Comments are sometimes moved somewhere else

Screenshot 2025-12-18 at 11 43 56

---

Drawback #3: The complexity of the code gets a little bit more, since we have to deal with merging AST (e.g. existing array fields or existing objects), which was plain objects before see `mergeValue`

JPeer264:
> If we discard user-created comments, I'd rather tend towards showing copy/paste-able instructions in the console. We have a showCopyPasteInstructions and makeCodeSnippet helpers that should make this fairly user friendly.

I removed the toml parser, since it seems really a little odd when the comments get removed. In the future we can always add it as feature in case there is demand. Cloudflare itself is also leaning more to jsonc anyways.

src/cloudflare/wrangler/get-entry-point-from-wrangler-config.ts
Lms24: more than enough reasons to use `jsonc-parser`. thanks for checking!

JPeer264:
It seems that the jsonc-parser has some problems with esbuild, as stated [here](https://github.com/microsoft/node-jsonc-parser/issues/57). The workaround works great for the [Cloudflare workers-sdk](https://github.com/cloudflare/workers-sdk/blob/e06c4db85ee4585857f044d442b7c9902ce70431/packages/wrangler/scripts/bundle.ts#L49-L50), which we can't use, since [fossilize](https://github.com/BYK/fossilize/blob/1b44f81dd852618d1ef66d34db9821a0bb8a58e1/src/impl.ts#L120-L141) doesn't allow extra fields.

I now removed the external dependencies and stick with our drawbacks, they are not that harmful. However, one test is skipped, where the comments are at the wrong place.

PR #1156, opened 2025-12-23
feat(cloudflare): Support wrapping workers main file

### LINKED ISSUE #1155: Cloudflare update workers default export
Cloudflare Workers are available in the default export of the main entrypoint, which is required to be listed in the `wrangler.jsonc`.

JPeer264:
(closes #1155)
(closes [WIZARD-115](https://linear.app/getsentry/issue/WIZARD-115/cloudflare-update-workers-default-export))

This adds support for the Wizard to automatically wrap the default export of the Cloudflare Workers with `Sentry.withSentry`.

As of now no support for e.g. logging has been added (so the template stayed the same as the printed one from before). 

The traceStep "Update Wrangler config with Sentry requirements" has moved down, as it will create a default entry file automatically if there is none, and will then update the `wrangler.jsonc` with the correct main file in case that was not given just yet. This is actually an edge case but a possibility.

### DIFF
--- a/e2e-tests/tests/cloudflare-worker.test.ts
+++ b/e2e-tests/tests/cloudflare-worker.test.ts
@@ -76,4 +76,12 @@ describe('cloudflare-worker', () => {
       '"binding": "CF_VERSION_METADATA"',
     ]);
   });
+
+  it('modifies the worker file to include Sentry initialization', () => {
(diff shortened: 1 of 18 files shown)


JPeer264:
> Nice addition and good tests! Did you try the CLI locally in an actual project as well?

Surely, with couple of `npm create cloudflare@latest` projects

PR #1158, opened 2025-12-29
test(remix): Use `clifty` in Remix e2e tests

Lms24:
This PR migrates the Remix e2e test from the previous, imperative CLI interaction logic to a new declarative interaction logc by using `clifty`.

Key benefits:
- declaring interactions and expected outputs prior to executing the wizard means less room for flakes, especially in conditional flows
- less code, less intermediate variables we need to assert on to proceed
- much more readable (IMHO)

#skip-changelog


Lms24:
superseded by #1166

PR #1161, opened 2025-12-30
test(e2e):  Isolate test runs with tmp directories

Lms24:
We get a lot of flakes in our wizard e2e tests and I suspect one of the main reasons for the flakes is interference from concurrent tests. Therefore, this PR now adds a `createIsolatedTestEnv` helper which 

- creates a temporary directory
- copies the respective test app into the tmpdir
- inits git 
- provides a cleanup function to be called once tests have ended

Furthermore, this PR also adapts all existing e2e tests to use this. I had to make some slight adjustments to get all tests working:

- some string assertions needed indentation adjustments. I suspect due to prettier running slightly differently in the tmpdir than in the repo (?)
- removed some unnecessary "helper functions" that extracted tests. I really don't like this pattern and would rather have a bit of duplication. Extracted `test`/`it` calls make it super hard to understand which `describe` block has which tests. 

#skip-changelog

### DIFF
--- a/e2e-tests/tests/expo.test.ts
+++ b/e2e-tests/tests/expo.test.ts
@@ -13,10 +13,11 @@ import { afterAll, beforeAll, describe, test, expect } from 'vitest';
 describe('Expo', () => {
   const integration = Integration.reactNative;
 
-  const { projectDir, cleanup } = createIsolatedTestEnv('react-native-expo-test-app');
+  const { projectDir, cleanup } = createIsolatedTestEnv(
+    'react-native-expo-test-app',
+  );
 
   beforeAll(async () => {
-
     const wizardInstance = startWizardInstance(integration, projectDir);
     const packageManagerPrompted = await wizardInstance.waitForOutput(
       'Please select your package manager.',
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -138,6 +138,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
     strategy:
+      fail-fast: false
```
Lms24, line 141: I never liked the fail-fast strategy here because it made it really hard to see which e2e tests failed after a change. Therefore, I decided to disable it

e2e-tests/tests/cloudflare-wrangler-sourcemaps.test.ts
```diff
@@ -15,16 +13,16 @@ import { KEYS, withEnv } from 'clifty';
...
-  );
-
+  let projectDir: string;
+  let cleanup: () => void;
   let wizardExitCode: number;
 
   beforeAll(async () => {
-    initGit(projectDir);
-    revertLocalChanges(projectDir);
+    const testEnv = createIsolatedTestEnv(
```
JPeer264: l: Following works as well and has fewer LOC 😏 

```js
({ projectDir, cleanup } = createIsolatedTestEnv(...))
```

On a second thought since it is inside a `beforeAll` you can even put it right outside and only have this structure:

```js
describe('', () => {
  const { projectDir, cleanup } = createIsolatedTestEnv(...);

  beforeAll(() => {});

  afterAll(cleanup);
});
```

Lms24: Great point, thanks! [f284176](https://github.com/getsentry/sentry-wizard/pull/1161/commits/f284176bdb3ee68304249d32a793dde9c01249ad)

PR #1162, opened 2026-01-02
test(e2e): Automatically build e2e test app matrix in CI

Lms24:
I noticed that we never added the `nextjs-16` test to our e2e test matrix in CI. Since this is super easy to miss, let's avoid this pitfall and automatically create the matrix.

(Also had to fix a bunch of problems in the e2e test. Looks like this test never worked at all 🤔)


JPeer264 (approved):
Great addition. LGTM

PR #1163, opened 2026-01-02
test(macos): Increase test timeouts

Lms24:
The `readXcodeBuildSettings > should return the build settings` flakes every now and then with a timeout.

Full disclosure: I'm not sure if this is a reasonable fix!  However, it looks like, `readXcodeBuildSettings` expectedly takes a while to complete. So I doubled the timeout to check if this gets rid of the flakes. Let me know if there's a better way to fix this!

EDIT: Example flake: https://github.com/getsentry/sentry-wizard/actions/runs/20601611749/job/59168314440#step:6:192

### DIFF
--- a/test/apple/macos-system-helper.test.ts
+++ b/test/apple/macos-system-helper.test.ts
@@ -125,8 +125,8 @@ describe('MacOSSystemHelpers', () => {
           ),
         );
       },
-      // Increased timeout to 10 seconds due to timeout errors for Node 18 and Node 20
-      10000,
+      // Increased timeout to 20 seconds due to timeout errors for Node 18 and Node 20
+      20_000,
     );
 
     test.runIf(process.platform !== 'darwin')(


philprime:
Please add a link to a run where it flaked so I can evaluate the issue you want to fix

Lms24:
@philprime good point, sorry! Here's a link: https://github.com/getsentry/sentry-wizard/actions/runs/20601611749/job/59168314440#step:6:192

philprime:
So my gut tells me that this is once again caused by GitHub's Action Runner images being of poor quality when working with Xcode and this delay is caused by the Xcode build environment.

As an example [this run](https://github.com/techprimate/TPPDF/actions/runs/20711806640/job/59453843950) of one of my projects shows that iOS 26.2 which is preinstalled by GitHub was not found. [Rerunning the job without changes](https://github.com/techprimate/TPPDF/actions/runs/20711806640/job/59456185774) passed without issues.

I believe we have the same issue here, so a timeout increase from 10 to 20 seconds won't help.

In [sentry-cocoa](https://github.com/getsentry/sentry-cocoa) we created a script [ci-ensure-runtime-loaded.sh](https://github.com/getsentry/sentry-cocoa/blob/main/scripts/ci-ensure-runtime-loaded.sh) which seems to have improved the situation for us. We could try to add the same script to this repository as a pre-step for macOS runners and observe the situation.

Lms24:
Thanks for taking a look @philprime! I gave your suggestion a try in https://github.com/getsentry/sentry-wizard/pull/1177.

Will tag you once I got this running.

PR #1165, opened 2026-01-02
test(e2e): Migrate Angular e2e tests to clifty

Lms24:
Angular e2e tests were previously skipped because they failed in CI. This PR:

- re-enables them
- changes the wizard flow to NOT add the source maps upload command to the build because `sentry-cli` fails with a non-zero exit code when no valid auth token is present.
- migrates the CLI interaction to `clifty` for better readability, fewer loc, no intermediate variables and hopefully less flakes.


JPeer264 (approved):
Love it!

PR #1169, opened 2026-01-05
chore(test): Update `@vitest/coverage-v8` to bump `glob` to `10.5.0`

Lms24:
closes https://github.com/getsentry/sentry-wizard/security/dependabot/90


Lms24:
yes, because `@vitest/coverage-v8` is the dependency that required the vulnerable `glob` version. I'll clarify this in the PR title/description.

PR #1181, opened 2026-01-05
fix(react-router): Avoid force-running npx react-router reveal

Lms24:
This PR fixes a bug in the React Router wizard where, when detecting a Typescript-based project, we'd force-run `npx react-router reveal` although to the best of my judgment, we didn't intend to do this.

This PR actually started out as a `clifty` migration for the RR e2e test, so it also includes this migration.


andreiborza (approved):
Thanks, and yeah we should always call `reveal` if the entry file doesn't exist. Nice catch!

PR #1183, opened 2026-01-08
feat(nextjs): Add --non-interactive flag for headless CLI operation

codyde:
## Summary

Adds support for fully headless/automated Sentry setup without browser authentication. This enables CI/CD pipelines and AI agents to scaffold Sentry configuration with environment variable placeholders that can be populated later.

## New CLI Flags

| Flag | Description |
|------|-------------|
| `--skip-auth` | Skip Sentry authentication, use env var placeholders |
| `--tracing` | Enable performance/tracing monitoring |
| `--replay` | Enable Session Replay |
| `--logs` | Enable Sentry Logs |
| `--tunnel-route` | Enable tunnel route for ad-blocker circumvention |
| `--mcp-cursor` | Add MCP config for Cursor |
| `--mcp-vscode` | Add MCP config for VS Code |
| `--mcp-claude` | Add MCP config for Claude Code |
| `--mcp-opencode` | Add MCP config for OpenCode |
| `--mcp-jetbrains` | Show MCP config for JetBrains IDEs |

## Example Usage

```bash
npx @sentry/wizard -i nextjs \
  --skip-auth \
  --tracing \
  --replay \
  --logs \
  --mcp-opencode \
  --ignore-git-changes
```

## Behavior in `--skip-auth` Mode

- Config files use `process.env.SENTRY_DSN` / `process.env.NEXT_PUBLIC_SENTRY_DSN`
- `next.config.js` uses `process.env.SENTRY_ORG` and `process.env.SENTRY_PROJECT`
- Creates `.env.example` documenting all required environment variables
- MCP configs use base URL (`mcp.sentry.dev/mcp`) without project scope
- Skips: example page creation, CI setup prompts, turbopack warning
- Feature flags default to `false` (must be explicitly enabled)

## Generated `.env.example`

```bash
# Sentry Configuration
SENTRY_DSN=
NEXT_PUBLIC_SENTRY_DSN=
SENTRY_ORG=
SENTRY_PROJECT=
SENTRY_AUTH_TOKEN=
```

## Use Cases

1. **AI Agent Setup**: Agents can run the wizard headlessly and populate env vars programmatically
2. **CI/CD Scaffolding**: Set up Sentry structure in pipelines without interactive prompts
3. **Template Projects**: Create project templates with Sentry pre-configured for env vars

## Testing

- [x] `yarn build` passes
- [x] `yarn lint` passes
- [x] `yarn test` passes
- [x] Manual testing with `--skip-auth` flag in test Next.js project

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,42 +1,5 @@
 # Changelog
 
-## Unreleased
-
-### Features
-
-- feat(nextjs): Add `--skip-auth` flag for headless CLI operation ([#TBD](https://github.com/getsentry/sentry-wizard/pull/TBD))
-
-  The Next.js wizard now supports a `--skip-auth` mode that scaffolds all Sentry files with environment
-  variable placeholders instead of connecting to Sentry for authentication. This enables fully headless
-  CLI operation where configuration values can be populated later (e.g., by an AI agent or CI/CD).
-
-  New CLI flags:
-
-  - `--skip-auth`: Skip Sentry authentication and use env var placeholders
-  - `--tracing`: Enable performance/tracing (with `--skip-auth`)
-  - `--replay`: Enable Session Replay (with `--skip-auth`)
-  - `--logs`: Enable Sentry Logs (with `--skip-auth`)
-  - `--tunnel-route`: Enable tunnel route (with `--skip-auth`)
-  - `--mcp-cursor`: Add MCP config for Cursor (with `--skip-auth`)
-  - `--mcp-vscode`: Add MCP config for VS Code (with `--skip-auth`)
-  - `--mcp-claude`: Add MCP config for Claude Code (with `--skip-auth`)
-  - `--mcp-opencode`: Add MCP config for OpenCode (with `--skip-auth`)
-  - `--mcp-jetbrains`: Show MCP config for JetBrains IDEs (with `--skip-auth`)
-
-  Example usage:
-
-  ```bash
-  npx @sentry/wizard -i nextjs --skip-auth --tracing --replay --logs --mcp-opencode --ignore-git-changes
-  ```
-
-  When using `--skip-auth`, the wizard:
-
-  - Creates config files with `process.env.SENTRY_DSN` / `process.env.NEXT_PUBLIC_SENTRY_DSN`
-  - Uses `process.env.SENTRY_ORG` and `process.env.SENTRY_PROJECT` in next.config
-  - Creates `.env.example` documenting required environment variables
-  - Skips example page creation, CI setup, and turbopack warning
-  - Uses base MCP URL (`mcp.sentry.dev/mcp`) without project scope
-
 ## 6.10.0
 
 - chore(deps): Upgrade `@sentry/node` from v7 to v10.29.0 ([#1126](https://github.com/getsentry/sentry-wizard/pull/1126))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,42 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+- feat(nextjs): Add `--skip-auth` flag for headless CLI operation ([#TBD](https://github.com/getsentry/sentry-wizard/pull/TBD))
+
+  The Next.js wizard now supports a `--skip-auth` mode that scaffolds all Sentry files with environment
+  variable placeholders instead of connecting to Sentry for authentication. This enables fully headless
+  CLI operation where configuration values can be populated later (e.g., by an AI agent or CI/CD).
+
+  New CLI flags:
+
+  - `--skip-auth`: Skip Sentry authentication and use env var placeholders
+  - `--tracing`: Enable performance/tracing (with `--skip-auth`)
+  - `--replay`: Enable Session Replay (with `--skip-auth`)
+  - `--logs`: Enable Sentry Logs (with `--skip-auth`)
+  - `--tunnel-route`: Enable tunnel route (with `--skip-auth`)
+  - `--mcp-cursor`: Add MCP config for Cursor (with `--skip-auth`)
+  - `--mcp-vscode`: Add MCP config for VS Code (with `--skip-auth`)
+  - `--mcp-claude`: Add MCP config for Claude Code (with `--skip-auth`)
+  - `--mcp-opencode`: Add MCP config for OpenCode (with `--skip-auth`)
+  - `--mcp-jetbrains`: Show MCP config for JetBrains IDEs (with `--skip-auth`)
+
+  Example usage:
+
+  ```bash
+  npx @sentry/wizard -i nextjs --skip-auth --tracing --replay --logs --mcp-opencode --ignore-git-changes
+  ```
+
+  When using `--skip-auth`, the wizard:
+
+  - Creates config files with `process.env.SENTRY_DSN` / `process.env.NEXT_PUBLIC_SENTRY_DSN`
+  - Uses `process.env.SENTRY_ORG` and `process.env.SENTRY_PROJECT` in next.config
+  - Creates `.env.example` documenting required environment variables
+  - Skips example page creation, CI setup, and turbopack warning
+  - Uses base MCP URL (`mcp.sentry.dev/mcp`) without project scope
+
 ## 6.10.0
 
 - chore(deps): Upgrade `@sentry/node` from v7 to v10.29.0 ([#1126](https://github.com/getsentry/sentry-wizard/pull/1126))
--- a/bin.ts
+++ b/bin.ts
@@ -163,35 +163,37 @@ const argv = yargs(hideBin(process.argv), process.cwd())
     'non-interactive': {
       default: false,
       describe:
-        'Run in non-interactive mode. Skips all prompts and uses environment variable placeholders for auth.',
+        '[NextJS only] Run in non-interactive mode. Skips all prompts and uses environment variable placeholders for auth.',
       type: 'boolean',
     },
     tracing: {
       describe:
-        'Enable performance/tracing monitoring. When set, skips the tracing prompt.',
+        '[NextJS only] Enable performance/tracing monitoring. When set, skips the tracing prompt.',
       type: 'boolean',
     },
     replay: {
-      describe: 'Enable Session Replay. When set, skips the replay prompt.',
+      describe:
+        '[NextJS only] Enable Session Replay. When set, skips the replay prompt.',
       type: 'boolean',
     },
     logs: {
-      describe: 'Enable Sentry Logs. When set, skips the logs prompt.',
+      describe:
+        '[NextJS only] Enable Sentry Logs. When set, skips the logs prompt.',
       type: 'boolean',
     },
     'tunnel-route': {
       describe:
-        'Enable tunnel route for ad-blocker circumvention. When set, skips the tunnel route prompt.',
+        '[NextJS only] Enable tunnel route for ad-blocker circumvention. When set, skips the tunnel route prompt.',
       type: 'boolean',
     },
     'example-page': {
       describe:
-        'Create an example page to test Sentry. When set, skips the example page prompt.',
+        '[NextJS only] Create an example page to test Sentry. When set, skips the example page prompt.',
       type: 'boolean',
     },
     mcp: {
       describe:
-        'Add MCP (Model Context Protocol) config for specified IDE(s). Options: cursor, vscode, claude, opencode, jetbrains',
+        '[NextJS only] Add MCP (Model Context Protocol) config for specified IDE(s). Options: cursor, vscode, claude, opencode, jetbrains',
       type: 'array',
       choices: ['cursor', 'vscode', 'claude', 'opencode', 'jetbrains'],
     },
--- a/bin.ts
+++ b/bin.ts
@@ -160,10 +160,10 @@ const argv = yargs(hideBin(process.argv), process.cwd())
         'Enable Spotlight for local development. This does not require a Sentry account or project.',
       type: 'boolean',
     },
-    'skip-auth': {
+    'non-interactive': {
       default: false,
       describe:
-        'Skip Sentry authentication and use environment variable placeholders. Enables fully headless CLI operation.',
+        'Run in non-interactive mode. Skips all prompts and uses environment variable placeholders for auth.',
       type: 'boolean',
     },
     tracing: {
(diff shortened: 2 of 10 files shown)


src/nextjs/nextjs-wizard.ts
```diff
@@ -138,13 +138,55 @@ export async function runNextjsWizardWithTelemetry(
...
+    await installPackage({
+      packageName: '@sentry/nextjs@^10',
+      packageNameDisplayLabel: '@sentry/nextjs',
+      alreadyInstalled: !!packageJson?.dependencies?.['@sentry/nextjs'],
+      forceInstall,
+    });
+
+  // Determine tunnel route setting
+  let tunnelRoute: boolean;
+  if (skipAuth) {
```
Lms24: m: is there a reason we only apply the `--tunnelRoute` argument on if `--skip-auth` is set? I'd tend towards treating these flags independently of each other

src/nextjs/nextjs-wizard.ts
```diff
@@ -377,78 +421,181 @@ export async function runNextjsWizardWithTelemetry(
...
-        selectedProject,
-        sentryUrl,
-        typeScriptDetected,
-        logsEnabled,
-      ),
-    );
+  // In skip-auth mode, skip example page creation
+  let shouldCreateExamplePage = false;
+  if (!skipAuth) {
+    shouldCreateExamplePage = await askShouldCreateExamplePage();
```
Lms24: m: I'd rather have this be its own flag. Nothing about `--skip-auth` suggests that it skips example page creation. 

Instead, I suggest we do something like this:

```js
const shouldCreateExamplePage = options.examplePage ?? await askShouldCreateExamplePage();
```

src/nextjs/nextjs-wizard.ts
```diff
@@ -377,78 +421,181 @@ export async function runNextjsWizardWithTelemetry(
...
-            label: 'I understand.',
-            hint: 'press enter',
-            value: true,
-          },
-        ],
-        initialValue: true,
-      }),
-    );
+  // Skip turbopack warning in skip-auth mode
+  if (!skipAuth) {
```
Lms24: m: I don't think we should skip this but rather just log the warning instead if the `clack.select` prompt that halts the wizard execution. My thinking is that this might even be valuable information for agents.

src/nextjs/nextjs-wizard.ts
```diff
@@ -377,78 +421,181 @@ export async function runNextjsWizardWithTelemetry(
...
-  const mightBeUsingVercel = fs.existsSync(
-    path.join(process.cwd(), 'vercel.json'),
-  );
-
-  if (mightBeUsingVercel && !options.comingFrom) {
-    clack.log.info(
-      "▲ It seems like you're using Vercel. We recommend using the Sentry Vercel \
-      integration to set up an auth token for Vercel deployments: https://vercel.com/integrations/sentry",
+  // Skip CI setup in skip-auth mode
+  if (!skipAuth) {
```
Lms24: m: same here. I think a log message that users/agents need to configure CI after the wizard finished _instead_ of the prompt, is better than entirely skipping this step, no?

src/nextjs/templates.ts
```diff
@@ -1,12 +1,43 @@
...
+# These environment variables are required for Sentry to work properly.
+# Copy this file to .env.local and fill in the values from your Sentry project.
+
+# Your Sentry DSN (from Project Settings > Client Keys)
+# Used on the server-side
+SENTRY_DSN=
+
+# Your Sentry DSN for client-side (same value as SENTRY_DSN)
+# The NEXT_PUBLIC_ prefix exposes this to the browser
+NEXT_PUBLIC_SENTRY_DSN=
```
Lms24, line 15: m: isn't it enough to just always set `NEXT_PUBLIC_SENTRY_DSN`? I think having one DSN env variable would be much less friction than two. Especially because it would tempt users to use different DSNs, when we don't recommend this in docs/other setup guides.

CHANGELOG.md
Lms24, line 1: We don't update `CHANGELOG.md` anymore. Instead, you can customize the PR description and put the custom description into a 

```md
### Changelog Entry

your description
```

section

Lms24 (commented):
Thanks for opening this PR! I think adding a non-blocking execution flow makes sense. Both for agents as well as humans :) However, I would ask that we take one of these two routes but not mix things up:

1. We add a `--agentMode` CLI flag which fully automates every step to make it non-blocking. It can work in conjuction with other high-level CLI flags like `--tracing` for high-level options but it can be used to skip over multiple steps or change wizard behaviour however otherwise necessary
2. We add specific flags for each step. For example `--auth=false` (or `--skip-auth`) controls authentication and the related code injection steps requiring dsns, project/org data etc (like right now). But in addition, `--example-page=false` skips the example page, `--tunnelRoute` decides whether to set tunnelRoute, `--setup-ci` for (not) skipping CI setup, etc. 

I would personally prefer 2 because this allows to make the wizard non-blocking for non-agentic use, too. For example, we could use it in docs to customize the wizard code snippet more based on the selected features on the setup page. Same in the in-product onboarding.

I just want to avoid that a flag like `--skip-auth` does more than its name implies. 
Likewise, I'd slightly prefer "enable" naming over "skip". So users set `--example-page=false` instead of `--skip-example-page`.  This more resembles the default behaviour of the wizard in the interactive mode. Though happy to hear counter arguments here :)

WDYT?

Lms24:
Also, please add an e2e test that covers the agentic setup flow so that we have a good picture of what kind of file output we expect from this flow.

bin.ts
```diff
@@ -160,6 +160,58 @@ const argv = yargs(hideBin(process.argv), process.cwd())
...
+    'mcp-opencode': {
+      default: false,
+      describe: 'Add MCP config for OpenCode (used with --skip-auth)',
+      type: 'boolean',
+    },
+    'mcp-jetbrains': {
+      default: false,
+      describe: 'Show MCP config for JetBrains IDEs (used with --skip-auth)',
+      type: 'boolean',
+    },
```
Lms24: Instead of adding 5 (or more in the future?) individual options, what about one combined option?

```sh
--mcp claude,cursor,jetbrains
# or
--mcp cursor
# or
--mcp cursor --mcp jetbrains
```

I think `yargs` handles all of these formats automatically, so we only need to adjust how we read these options (as an array of strings, rather than boolean flags, I presume)

bin.ts
```diff
@@ -160,6 +160,58 @@ const argv = yargs(hideBin(process.argv), process.cwd())
         'Enable Spotlight for local development. This does not require a Sentry account or project.',
       type: 'boolean',
     },
+    'skip-auth': {
+      default: false,
+      describe:
+        'Skip Sentry authentication and use environment variable placeholders. Enables fully headless CLI operation.',
```
Lms24: For the time being, let's prefix all these options with something like

```suggestion
    'skip-auth': {
      default: false,
      describe:
        '[NextJS only] Skip Sentry authentication and use environment variable placeholders. Enables fully headless CLI operation.',
```

Users can call `npx @sentry/wizard --help` and  might be confused about why this flag does nothing in e.g. the nuxt or sveltekit wizard.

Which brings me to my last point (for now):

Long-term, we should definitely bring all of these options to all our wizards. I fully understand that trialing this in the most used one makes sense but I _really_ want to avoid the other flows falling behind. We'Ve had that before and it led to sub-par wizard experiences for non-NextJS users. Though let's tackle this in individual PRs please 😅

Lms24:
Oh yea, [speaking of](https://github.com/getsentry/sentry-wizard/pull/1183#discussion_r2671942095) `--help`, the e2e test needs adjustments for whatever options we settle on.

src/nextjs/nextjs-wizard.ts
```diff
@@ -377,7 +415,11 @@ export async function runNextjsWizardWithTelemetry(
     }
   });
 
-  const shouldCreateExamplePage = await askShouldCreateExamplePage();
+  // Example page - use CLI flag if provided, otherwise prompt (skip in skip-auth mode if not explicitly enabled)
+  const shouldCreateExamplePage =
+    options.examplePage ??
+    (skipAuth ? false : await askShouldCreateExamplePage());
```
Lms24: m: Why should we continue to fall back to `skipAuth` here? I'd prefer it if this only depended on `--example-page`

src/nextjs/nextjs-wizard.ts
```diff
@@ -390,28 +432,72 @@ export async function runNextjsWizardWithTelemetry(
...
-          {
-            label: 'I understand.',
-            hint: 'press enter',
-            value: true,
-          },
-        ],
-        initialValue: true,
-      }),
-    );
+    if (skipAuth) {
```
Lms24: I get why this needs to be skipped in "agent mode" but, to me, `--skip-auth` makes no semantic sense here. I'm wondering if we can silence these warning steps with something like `--accept-warnings`.

I guess the alternative here would be to re-think the finely grained approach and just add a general `--non-interactive` mode. This would skip auth, accept warnings, and make the best decisions along the way without requiring user input.

src/nextjs/nextjs-wizard.ts
```diff
@@ -421,34 +507,94 @@ export async function runNextjsWizardWithTelemetry(
       "▲ It seems like you're using Vercel. We recommend using the Sentry Vercel \
       integration to set up an auth token for Vercel deployments: https://vercel.com/integrations/sentry",
     );
+  } else if (skipAuth) {
```
Lms24: actually another candidate for something like `--non-interactive`

src/nextjs/nextjs-wizard.ts
```diff
@@ -421,34 +507,94 @@ export async function runNextjsWizardWithTelemetry(
...
+${chalk.yellow('Next steps:')}
+1. Copy ${chalk.cyan('.env.example')} to ${chalk.cyan('.env.local')}
+2. Fill in your Sentry DSN, org, project, and auth token from ${chalk.cyan(
+      'https://sentry.io',
+    )}
+3. Restart your dev environment (e.g. ${chalk.cyan(
+      `${packageManagerForOutro.runScriptCommand} dev`,
+    )})
+
+${chalk.dim('Environment variables needed:')}
```
Lms24: nit: I'd rather make sure this is stronger highlighted than dimmed but feel free to disregard

```suggestion
${chalk.cyan('Environment variables needed:')}
```

src/nextjs/nextjs-wizard.ts
```diff
@@ -461,31 +607,94 @@ async function createOrMergeNextJsFiles(
...
+        ? wizardOptions.tracing ?? false
+        : promptedFeatures.performance,
+      replay: replayFlagProvided
+        ? wizardOptions.replay ?? false
+        : promptedFeatures.replay,
+      logs: logsFlagProvided
+        ? wizardOptions.logs ?? false
+        : promptedFeatures.logs,
+    };
+  }
```
Lms24: l: I think we can simplify this entire logic here:
1. build the `featuresToPrompt` array based on the `xProvided` checks
2. if `featuresToPrompt.length > 0`, run the selection prompt
3. print selected features

Should save us a couple of lines here, as it makes the special case (= all features provided) obsolete.  :)

Lms24 (commented):
Thanks for making the changes! After thinking about this for a while, I think something like a `--non-interactive` or `--headless` flag, would make sense and could even replace `--skip-auth`. My thinking here:

- `--skip-auth` does not imply that prompts/warnings/steps requiring input are skipped
- `--headless`/`--non-interactive` (feel free to suggest other names) suggest skipping prompts and as a consequence authentication (at least to a point I guess xD)

WDYT?

e2e-tests/tests/nextjs-non-interactive.test.ts
```diff
@@ -0,0 +1,364 @@
...
+
+describe('NextJS Non-Interactive Mode', () => {
+  const integration = Integration.nextjs;
+  let wizardExitCode: number;
+
+  const { projectDir, cleanup } = createIsolatedTestEnv('nextjs-15-test-app');
+
+  beforeAll(async () => {
+    // Create package-lock.json so npm is auto-detected (avoids package manager prompt)
+    createPackageLockForNpm(projectDir);
```
Lms24, line 53: we should be able to bypass package manager selection with the `--inon-interactive` flag. Otherwise the non-interactive mode might hang here. Probably fine to fall back to npm here in case the detection fails.

Lms24 (approved):
I think this is fine and `--non-interactive` sounds like a good choice. Thanks for making the hanges! 

All good from my end! Before we merge this, I'd be curious on the new `sentry` CLI's perspective, given it's also designed for agentic use (cc @betegon). Ideally, we can scope the individual arguments to the respective wizard flows, similarly to what I sketched out in #1200, once the new CLI calls the wizard. Anyway, not a blocker to merge this into the current wizard from my PoV.

src/nextjs/nextjs-wizard.ts
```diff
@@ -461,31 +632,78 @@ async function createOrMergeNextJsFiles(
   sentryUrl: string,
   sdkConfigOptions: SDKConfigOptions,
   spotlight = false,
+  wizardOptions: WizardOptions,
+  useEnvVars = false,
```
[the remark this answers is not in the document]
codyde: Fixed: Added `nonInteractive` guards around all unguarded `clack.confirm()` calls in `createOrMergeNextJsFiles`:
- Sentry server/edge config overwrite prompt (line ~733): logs info message and defaults to overwrite in non-interactive mode
- CJS next.config SDK detection prompt (line ~1114): logs info message and defaults to inject in non-interactive mode
- ESM/MJS/TS next.config SDK detection prompt (line ~1168): same treatment
- ESM config catch-block confirm (line ~1297): skipped in non-interactive mode

src/nextjs/nextjs-wizard.ts
```diff
@@ -77,7 +86,7 @@ export function runNextjsWizard(options: WizardOptions) {
 export async function runNextjsWizardWithTelemetry(
   options: WizardOptions,
 ): Promise<void> {
-  const { promoCode, telemetryEnabled, forceInstall } = options;
+  const { promoCode, telemetryEnabled, forceInstall, nonInteractive } = options;
```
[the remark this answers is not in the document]
codyde: Fixed: `--non-interactive` now automatically bypasses the git repo check by passing `ignoreGitChanges: nonInteractive || options.ignoreGitChanges` to `confirmContinueIfNoOrDirtyGitRepo`. Users no longer need to explicitly pass `--ignore-git-changes` alongside `--non-interactive`.

Lms24:
closing as the wizard will be superseded by `sentry init`

PR #1185, opened 2026-01-09
chore(publish): Fix `release.yml` commit patterns

Lms24:
The regexp patterns in `release.yml` were incorrect and craft couldn't parse them. This made it choose its default changelog template over our customized version. This PR fixes the patterns, by mostly following the [defaults](https://getsentry.github.io/craft/configuration/) from craft.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -10,24 +10,29 @@ changelog:
       labels:
         - "Changelog: Breaking Change"
       commit_patterns:
-        - "^\w+(\(\w+\))?!:"
-    - title: New Features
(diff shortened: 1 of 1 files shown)


BYK:
@Lms24 the only change seems to be the title "Internal Changes", right? I can pull that into the default template so you don't have to override.

Lms24:
The primary intent for the custom config was 

* internal changes
* removing the AI-esque emojis from the main sections 😅 

No strong feelings about this being default or not. Whatever you think makes more sense for craft 🙂

PR #1189, opened 2026-01-14
chore: add unlabeled trigger to changelog-preview

BYK:
This PR adds the `unlabeled` trigger to the changelog preview workflow to ensure the preview is updated when a label is removed.


Lms24 (commented):
Let's fix the conflict, LGTM otherwise

PR #1233, opened 2026-03-02
chore(agents): Use `AGENTS.md` with symlinks

chargome:


AGENTS.md
andreiborza: q: Should we generalize the model here? Wondering if it'll misattribute otherwise.

PR #1236, opened 2026-03-03
ref(agents): Convert commands to dotagents skills

chargome:
Move agent commands from `.claude/commands/` and `.cursor/commands/` into `.agents/skills/` as SKILL.md files with frontmatter. This consolidates on the dotagents skill format so all agent tools discover commands from the canonical `.agents/skills/` directory.


.agents/skills/add-changelog/SKILL.md
```diff
@@ -1,4 +1,7 @@
-# Add Changelog Entry
+---
+name: add-changelog
```
andreiborza: Do we even need this skill anymore? I thought we handle auto-handle the changelog now with craft. cc @Lms24.

I think we can drop this skill entirely.

PR #1247, opened 2026-03-31
fix(apple): Dynamic SDK version, enableLogs fix, dSYM guidance, and doctor command

filipealva:
### fix(apple): Dynamic SDK version, enableLogs fix, dSYM guidance, and doctor command

The Apple wizard had several issues affecting iOS users:

- SPM dependency was hardcoded to minimumVersion 8.0.0 with
  upToNextMajorVersion, meaning it would never resolve to 9.x.
  Now fetches the latest version from the Sentry release registry
  (sentry.cocoa) and falls back to 8.0.0 if the fetch fails.

- The enableLogs code snippet used options.experimental.enableLogs
  which no longer exists in sentry-cocoa v9. Removed the .experimental
  namespace so the generated code compiles correctly.

- The dSYM upload build phase script had no way to skip uploads and
  lacked CI/CD auth guidance. Added SENTRY_SKIP_DSYM_UPLOAD env var
  support and a note about setting SENTRY_AUTH_TOKEN in CI.

- Added an info message about the expected Sentry.framework dSYM
  warning in App Store uploads (normal for pre-compiled SPM frameworks).

- There was no way to diagnose or fix an existing integration. Added
  a --fix flag that runs five diagnostic checks (sentry-cli, auth
  token, SDK version, build phase, init code) and offers to auto-fix
  what it can.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## Unreleased
+
+### Features
+
+#### Apple
+
+- feat(apple): Dynamically fetch latest Sentry Cocoa SDK version for SPM instead of hardcoding
+- feat(apple): Add `--fix` flag for diagnosing and repairing Apple integrations
+- fix(apple): Remove `.experimental` namespace from `enableLogs` option in code snippets
+- ref(apple): Add CI/CD guidance and `SENTRY_SKIP_DSYM_UPLOAD` option for dSYM upload build phase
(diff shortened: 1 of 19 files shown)


philipphofmann:
I'm sorry, @filipealva, for the late reply, and thanks for your contribution. This PR got lost in our notifications. As it's outdated now and it was anyway a bit too big, please double-check whether your required changes still make sense, and please split up the changes into smaller PRs.

PR #995, opened 2025-05-14
test(react-native): Add E2E Tests verifying RN iOS project builds correctly

antonis:

PR #1263, opened 2026-04-28
test(e2e): Build RN and Expo iOS/Android release projects

### LINKED ISSUE #948: Build RN and Expo E2E iOS and Android projects
Run release builds of the native projects to ensure the wizard has not broken the `build.gradle` or Xcode project when modifying them.

alwx:
Closes getsentry/sentry-wizard#948

The idea here is to add end-to-end coverage that ensures the React Native and Expo wizards won't break the native iOS or Android projects when patching `build.gradle` or the Xcode project.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,7 +112,9 @@ jobs:
         env:
           OS_VERSION: '18.5'
           PLATFORM: 'iOS'
-        run: ./scripts/ci-ensure-runtime-loaded.sh --os-version "$OS_VERSION" --platform "$PLATFORM"
+        run:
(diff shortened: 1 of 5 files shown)


antonis (commented):
Thank you for looking into this @alwx 🙇 
Linking previous attempt on this in https://github.com/getsentry/sentry-wizard/pull/993 and https://github.com/getsentry/sentry-wizard/pull/995. The main issue I was facing was the long build times which can be acceptable if we restrict the jobs to run only when needed.

alwx:
@antonis you're probably right — even though having these checks could be nice, if I think more pragmatically then it doesn't seem like we're experiencing any issues with that anyway and the costs outweight the benifits. Will close it with the same comment in the issue itself.

PR #1278, opened 2026-05-28
chore(sentry-cli): Upgrade to 2.58.6

szokeasaurusrex:
Upgrade sentry-cli to [2.58.6](https://github.com/getsentry/sentry-cli/releases/tag/2.58.6), which includes security fixes.

### DIFF
--- a/src/sourcemaps/tools/sentry-cli.ts
+++ b/src/sourcemaps/tools/sentry-cli.ts
@@ -35,7 +35,7 @@ export async function configureSentryCLI(
   const packageDotJson = await getPackageDotJson();
 
   await installPackage({
-    packageName: '@sentry/cli@^2',
+    packageName: '@sentry/cli@2.58.6',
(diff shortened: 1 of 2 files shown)


src/sourcemaps/tools/sentry-cli.ts
```diff
@@ -35,7 +35,7 @@ export async function configureSentryCLI(
   const packageDotJson = await getPackageDotJson();
 
   await installPackage({
-    packageName: '@sentry/cli@^2',
+    packageName: '@sentry/cli@2.58.6',
```
szokeasaurusrex, line 38: Will this also allow newer versions, e.g. if we ever have a `2.58.7`?

andreiborza: No this will only install that particular version, the way we had it is already always installing the latest v2. This PR  can be closed.

andreiborza:
Not necessary, the wizard will always pull in the latest v2 version.

PR #1286, opened 2026-06-12
fix(all): Update `dataCollection` opt-out hint to include `httpBodies`

s1gr1d:
Update all snippets to exclude `httpBodies` and also revert the changes back to `sendDefaultPii` in the React Native, Expo and Electron snippets.


chargome:
bugbot run

chargome (approved):

PR #1299, opened 2026-06-30
feat(apple)!: Remove CocoaPods as an installation option

philprime:
## Summary

Remove CocoaPods as an installation option from the Sentry wizard for the Apple platform (COCOA-1241). CocoaPods trunk is going read-only on Dec 2, 2026, and we stopped publishing to CocoaPods trunk after v9.19.1.

- Delete `configurePackageManager` (inlined SPM tag into apple wizard)
- Remove `shouldUseSPM` parameter from `configureXcodeProject` (always SPM)
- Move `cocoapod.ts` from `src/apple/` to `src/react-native/` (still used there)

**BREAKING CHANGE**: CocoaPods is no longer offered as an installation option in the Apple wizard. The wizard always uses Swift Package Manager.

Closes COCOA-1241

## Related PRs

- sentry-docs: https://github.com/getsentry/sentry-docs/pull/18600
- sentry: https://github.com/getsentry/sentry/pull/118723


philipphofmann (approved):
LGTM, thank you

PR #1303, opened 2026-07-16
ref(react-router): Install SDK package with version @^10

### LINKED ISSUE #1275: Pin installed JavaScript SDK packages to current major in sentry-wizard
Before releasing JS SDK v11, pin all packages that sentry-wizard installs to the current major (`^10`) so we don't run into compatibility issues once v11 is released.

msonnb:
in preparation of the v11 major release, let's ensure we limit the installed SDK package version to latest v10. We'll bump to 11 post-release.

Only react-router and profiling-node are missing, the others have it already from when we released v10.

closes getsentry/sentry-wizard#1275

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- ref(react-router): Install SDK package with version `@^10` ([#1303](https://github.com/getsentry/sentry-wizard/pull/1303))
+
 ### Breaking Changes
 
 - feat(apple)!: Remove CocoaPods as an installation option ([#1299](https://github.com/getsentry/sentry-wizard/pull/1299))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- ref(react-router): Install SDK package with version `@^10` ([#1303](https://github.com/getsentry/sentry-wizard/pull/1303))
```
nicohrubec, line 5: afaik putting this here is optional so you can also omit it next time and then we add an auto-entry during release, just an fyi can leave as is 😅

nicohrubec (approved):
thanks! so in all other wizards this is pinned already?

msonnb:
> thanks! so in all other wizards this is pinned already?

yes! happened in https://github.com/getsentry/sentry-wizard/pull/1048 already

andreiborza (approved):
Lets also create an issue to remind ourselves to unskip this.

PR #1311, opened 2026-07-23
fix(react-native): Add `use_modular_headers!` to iOS Podfile for Swift RNSentry pod

antonis:
## Summary

Re-enables the React Native e2e test on macOS (skipped in #1303) by fixing the underlying `pod install` failure.

**Root cause:** As of `@sentry/react-native` 8.19.0, the `RNSentry` pod ships Swift code (the `RNSentryInternal` bridge over `SentrySDK.internal`, part of the migration to the prebuilt `Sentry.xcframework` — see sentry-react-native [#6413](https://github.com/getsentry/sentry-react-native/pull/6413) / [#6495](https://github.com/getsentry/sentry-react-native/pull/6495)). CocoaPods refuses to integrate a Swift pod that depends on a non-modular Objective-C pod (e.g. `React-hermes` on RN versions that don't modularize it by default) unless the Podfile opts into module maps. So `pod install` fails with:

```
[!] The following Swift pods cannot yet be integrated as static libraries:
The Swift pod `RNSentry` depends upon `React-hermes`, which does not define
modules. [...] you may set `use_modular_headers!` globally in your Podfile
```

The error was masked in CI because `pod install --silent` only surfaces curl progress, not the CocoaPods error (which is why #1305's `--verbose` repro was on the right track). This affected real RN users too — the SDK CHANGELOG instructs them to add `use_modular_headers!` by hand — so the wizard now does it for them.

## Changes

- `src/react-native/cocoapod.ts`: new idempotent `addModularHeaders()` that inserts `use_modular_headers!` above the first `target` block in `ios/Podfile` (no-op if already present, ignores commented-out targets, logged transparently).
- `src/react-native/react-native-wizard.ts`: patch the Podfile before `pod install` whenever the project uses CocoaPods, so the wizard's install and any later manual `pod install` both work.
- `e2e-tests/tests/react-native.test.ts`: remove `describe.skipIf(darwin)` and assert the Podfile gets `use_modular_headers!`. Fixture Podfile stays vanilla so the test exercises the wizard adding it.
- `test/react-native/cocoapod.test.ts`: 7 new unit tests for `addModularHeaders`.
- `CHANGELOG.md`: entry under Unreleased.

## Testing

- Reproduced the exact `pod install` failure with `@sentry/react-native` 8.19.0 on the RN 0.78 test app (exit 1); confirmed `use_modular_headers!` fixes it (exit 0).
- **Full RN e2e test passes on macOS** (11/11) — the wizard logs adding the line, then `Pods installed.` succeeds.
- `test/react-native/` unit tests: 140/140. `yarn build` and `yarn lint` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- fix(react-native): Add `use_modular_headers!` to the iOS `Podfile` so `pod install` succeeds with the Swift-based RNSentry pod
 - feat(react-native): Prompt for Logs first in the feature selection flow ([#1308](https://github.com/getsentry/sentry-wizard/pull/1308))
 - fix(react-native): Use User Feedback Widget terminology in the setup flow
 - ref(react-router): Install SDK package with version `@^10` ([#1303](https://github.com/getsentry/sentry-wizard/pull/1303))
--- a/e2e-tests/tests/react-native.test.ts
+++ b/e2e-tests/tests/react-native.test.ts
@@ -11,8 +11,7 @@ import { afterAll, beforeAll, describe, test, expect } from 'vitest';
 //@ts-expect-error - clifty is ESM only
 import { KEYS, withEnv } from 'clifty';
 
-// TODO: remove skipIf when macOS CI is fixed
-describe.skipIf(process.platform === 'darwin')('ReactNative', () => {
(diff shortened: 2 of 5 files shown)


msonnb (approved):
nice, thank you!

PR #1312, opened 2026-07-24
fix(sveltekit): support Cloudflare server setup

### LINKED ISSUE #803: Adjust SvelteKit wizard to configure the SDK correctly for Cloudflare Pages
Right now, the wizard doesn't take into account the special setup needed for Cloudflare. 

We should eventually add a check for cloudflare config and adjust the setup in `hooks.server.ts` accordingly. No high prio on this though. For now we just tell people to follow the docs.

sidsri14:
## Summary
- detect `@sveltejs/adapter-cloudflare` in the loaded SvelteKit config
- initialize Cloudflare server hooks with `initCloudflareSentryHandle` before `sentryHandle`
- skip Node-only `instrumentation.server` setup for Cloudflare and preserve existing hooks
- preserve source line endings and cover new and merged hook configurations

Fixes #803

## Testing
- `corepack yarn fix`
- `corepack yarn tsc`
- `corepack yarn lint`
- `corepack yarn vitest run test/sveltekit/templates.test.ts test/sveltekit/sdk-setup/svelte-config.test.ts test/sveltekit/sdk-setup/setup.test.ts` (48 passed)
- `corepack yarn test` (909 passed; 54 unrelated Windows-only path and line-ending assertions failed, with the changed SvelteKit tests passing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- fix(sveltekit): Configure Cloudflare adapters with the worker-compatible server initialization
 - fix(react-native): Add `use_modular_headers!` to the iOS `Podfile` so `pod install` succeeds with the Swift-based RNSentry pod
 - feat(react-native): Prompt for Logs first in the feature selection flow ([#1308](https://github.com/getsentry/sentry-wizard/pull/1308))
 - fix(react-native): Use User Feedback Widget terminology in the setup flow
--- a/src/sveltekit/sdk-setup/setup.ts
+++ b/src/sveltekit/sdk-setup/setup.ts
@@ -28,6 +28,7 @@ import x = recast.types;
 import t = x.namedTypes;
 import {
   enableTracingAndInstrumentation,
+  isCloudflareAdapter,
   type PartialBackwardsForwardsCompatibleSvelteConfig,
 } from './svelte-config';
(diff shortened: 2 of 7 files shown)


Lms24 (commented):
Hey @sidsri14 thanks for contributing!

This looks good to me mostly, but we need to make a few changes to get this merged:

1. Can you add an e2e test app and test for a cloudflare setup?
2. I'm not yet sure detecting the cloudflare adapter alone is enough to start cloudflare configuration. Users could also [use](https://svelte.dev/docs/kit/adapter-auto) `adapter-auto` on cloudflare, correct? If so, I think this check would not be enough. What do you think about asking users if they deploy their app to cloudflare? I think it would be fair to ask this on every setup. We can use the cloudflare adapter detection logic to preselect the answer but still leave it up to users to make the final decision, whether to set up the SDK for cloudflare or the regular platforms.

Lms24:
We discussed this quickly in the team and decided to not merge this PR (or other features for that matter) at this time. The Sentry wizard will be replaced by our new, AI-powered `sentry init` flow, so it makes little sense to continue work on the wizard. You can see the instructions for the new AI-based installation here: https://docs.sentry.io/platforms/javascript/guides/sveltekit/#install

Apologies for not closing the original issue sooner. We should have done that once we deprioritized work on the wizard. Thanks for contributing though, it's really appreciated.

PR #1348, opened 2026-09-11
feat(sveltekit): Import sentrySvelteKit from @sentry/sveltekit/vite on SDK v11

chargome:
SDK v11 moves `sentrySvelteKit` to the `@sentry/sveltekit/vite` subpath and drops it from the root export. No v10 release ships the subpath, so the wizard picks the import path from the SDK major that is actually installed after the install step: `/vite` on v11+, root export on v10. Unknown versions fall back to the pinned install range (`SENTRY_SVELTEKIT_SDK_RANGE`), so the `^11` bump flips the default with one constant.

Also writes `sentryUrl` instead of `url` for self-hosted setups. The top-level `url` option never existed on the v10 plugin, so the self-hosted URL was silently dropped before.

Verified with the sveltekit-tracing e2e suite on `10.74.0` and sveltekit-hooks on `11.0.0-beta.2`. The e2e assertions still expect the root import and must switch to `/vite` when the range is bumped.

closes https://linear.app/getsentry/issue/WIZARD-153/sveltekit-vite-subpath-and-sentryurl


chargome:
bugbot run

chargome:
bugbot run

msonnb (approved):

Issue #1336, opened 2026-09-10
React Router: `/vite` subpath, 7.15 minimum, session profiling


chargome:
* `sentryReactRouter` and `sentryOnBuildEnd` move to `@sentry/react-router/vite`. Update the vite and react-router-config codemods and the copy-paste templates.
* v11 peers require `react-router >=7.15`. The pre-7.15 branches behind `supportsOnError` and `supportsInstrumentationAPI` become dead. Remove them and warn on older projects.
* The profiling template still writes `profilesSampleRate`, removed in v11. Use `profileSessionSampleRate` plus `profileLifecycle: 'trace'`.
* Update the react-router e2e assertions.

PR #1352, opened 2026-09-14
feat(react-router): Import Vite plugin helpers from @sentry/react-router/vite on SDK v11

chargome:
SDK v11 moves `sentryReactRouter` and `sentryOnBuildEnd` to the `@sentry/react-router/vite` subpath and drops them from the root export. No v10 release ships the subpath, so the wizard picks the import path from the SDK major that is actually installed after the install step: `/vite` on v11+, root export on v10. Unknown versions fall back to the pinned install range (`SENTRY_REACT_ROUTER_SDK_RANGE`), so the `^11` bump flips the default with one constant. The vite and react-router.config codemods, the default config template, and the manual copy-paste snippets all take the path.

The profiling template moves from the removed `profilesSampleRate` to session profiling (`profileSessionSampleRate` plus `profileLifecycle: "trace"`), which v10 already supports.

React Router older than 7.15 now gets a warning that SDK v11 requires 7.15+. The pre-7.15 branches (`supportsOnError`, `supportsInstrumentationAPI`) stay while the wizard installs v10; removing them belongs with the bump.

Verified with the react-router e2e suite on `10.74.0` and react-router-instrumentation-api on `11.0.0-beta.2`.

clsoes https://github.com/getsentry/sentry-wizard/issues/1336

closes https://linear.app/getsentry/issue/WIZARD-155/react-router-vite-subpath-715-minimum-session-profiling


chargome:
bugbot run

chargome:
bugbot run

chargome:
bugbot run

src/react-router/templates.ts
```diff
@@ -44,7 +45,7 @@ Sentry.init({
     enableTracing ? '// Capture 100% of the transactions' : ''
```
msonnb, line 45:
```suggestion
    enableTracing ? '// Capture 100% of the traces' : ''
```

msonnb (approved):

PR #1354, opened 2026-09-14
feat(js): Warn when the wizard runs on a Node.js version SDK v11 does not support

chargome:
SDK v11 drops Node.js 18 and requires `>=20.19.0 <22.0.0 || >=22.12.0 <23.0.0 || >=23.2.0`. The JS wizards (Next.js, SvelteKit, Nuxt, React Router, Remix, Angular, Cloudflare) now warn right after the welcome banner when the running Node.js is outside that range. The wizard continues, since the project may be built and run on a different Node.js than the one the wizard runs on.

This is the warning half of WIZARD-151. The wizard's own `engines` field is unchanged: raising it is a breaking change for the wizard package and ships with the next wizard major, together with moving the wizard's own telemetry to `@sentry/node` v11, which is what actually requires the newer Node.js at runtime.

closes https://linear.app/getsentry/issue/WIZARD-151/gate-js-wizards-on-node-2019-and-bump-wizard-engines


chargome:
bugbot run

chargome:
bugbot run

msonnb (approved):

